PR #1, opened 2022-02-16
Initial CLI for generating test data

pcapriotti:
This adds a new `cli` package, containing the beginning of a CLI executable intended to be used for testing the MLS implementation on the server side. At the moment, it can only generate new key packages for a given client identity.

As part of this PR, I've also changed how `ClientId` values are represented as strings, so that they follow the convention (used by both the backend and the clients) that the device ID be represented in hex.


.gitignore
```diff
@@ -4,9 +4,6 @@
 **/debug/
 **/target/
 
-# Libs don't ship cargo.lock
-Cargo.lock
```
OtaK, line 8: The workspace shouldn't ship a Cargo.lock file

What could be done is the following: 
* Exclude the cli from the workspace
* Version-control the Cargo.lock produced at `cli/Cargo.lock` - Crates that are not part of a workspace get their own Cargo.lock

OtaK (changes requested):
The other concerns I've noted will change drastically once I push the fixes for OpenMLS 0.4.0 compatibility, so I didn't add a review for them.

OtaK (approved):

PR #32, opened 2022-05-11
feat: Implement KeyPackage pruning [CL-16]

OtaK:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Keypackages that were uploaded and are now backend-owned would stay in the keystore, leading to duplication and maybe multi-use of Keypackages

### Solutions

Added an API to allow consumers to prune uploaded Keypackages from the Keystore

#### Test Coverage (Optional)

- [x] I have added automated test to this contribution

#### How to Test

`cargo test client_generates_correct_number_of_kpbs`

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [x] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [x] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References


OtaK:
> @OtaK If the entire keypackage is deleted, where do the corresponding ciphersuite, init_key, and (inside the LeafNode) encryption_key get correlated with the KeyPackageRef? When the Lifetime expires, how do you invalidate them. You also need a reference to the appropriate Credential and signature_key.

I don't understand, pruning is supposed to be when the KeyPackage is sent to the backend and is then backend-owned, correct? All of this is sent to the backend (MLS DS) and once it is sent and saved there, we should just delete the local keypackages so that we (or OpenMLS) have no way of accidentally interact with them. 
Maybe it might help for the rationale but: there's no API within OpenMLS to filter on keypackages, if it exists locally, it might get used, and that's not okay with the single-use nature of keypackages.

> where do the corresponding ciphersuite, init_key, and (inside the LeafNode) encryption_key

Isn't this in the context of a MLS Group? Keypackage pruning is only client-local, basically it's keystore housekeeping.

> When the Lifetime expires, how do you invalidate them.

Good call, I'm gonna add transparent lifetime-based pruning at least to make sure we get rid of expired material. I'm thinking of adding that when consumers try to get keypackages

beltram (approved):

OtaK:
@rohan-wire 
> I think you only want to delete the keypackages once you receive a welcome for them
I think this can cause a race condition between the backend and the client where a client may use a Keypackage to create a local MLS Group and someone might use the same keypackage to create a welcome message at the same time remotely. 

I really think it's safer to just straight up delete the local copy of the Keypackage once it has been sent to the backend (sole ownership of the DS makes it impossible to have those kinds of race conditions)

OtaK:
Superseded by #52

PR #39, opened 2022-05-30
feat: Adding support for standard WASM32 [CL-3]

OtaK:
Based on the past wasm32-unknown-emscripten work


.github/workflows/rust.yml
```diff
@@ -7,23 +7,41 @@ jobs:
...
     steps:
     - uses: actions/checkout@v3
+
     - name: Install minimal stable with clippy and rustfmt
       uses: actions-rs/toolchain@v1
       with:
         profile: minimal
         toolchain: stable
+
+    - name: Install wasm-pack
```
beltram, line 17: There's [an action](https://github.com/marketplace/actions/wasm-pack-action) for that (which claims to be much faster)

beltram (changes requested):

.github/workflows/rust.yml
OtaK: > There's [an action](https://github.com/marketplace/actions/wasm-pack-action) for that (which claims to be much faster)

Much faster than `cargo install wasm-pack`, yes, but I'm using the prescribed method by wasm-pack [here](https://rustwasm.github.io/docs/wasm-bindgen/wasm-bindgen-test/continuous-integration.html#github-actions)

I saw this action but I'm very concerned (looking that the source code [here](https://github.com/jetli/wasm-pack-action/blob/master/src/main.ts)) that it might break if one day the `wasm-pack` folks change anything in their package structure.

By using their shell script it does things the way they want to and we guarantee no sudden breakage

PR #57, opened 2022-07-08
feat: join by external commit support - CL-47

augustocdias:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Will add 3 new functions to enable callers to make external join requests. When calling the join function the library will store a temporary state of the returned group. When the caller calls merge, it will merge the commit, persist the state of the group and delete the temporary group (a new table was created for that purpose).

#### Test Coverage (Optional)

- [x] I have added automated test to this contribution

##### References


crypto/src/external_commit.rs
```diff
@@ -0,0 +1,243 @@
...
+            &MlsConversationConfiguration::default().as_openmls_default_configuration(),
+            &[],
+            credentials,
+        )
+        .await
+        .map_err(MlsError::from)
+        .map_err(CryptoError::from)?;
+
+        let mut buf = vec![];
+        group.save(&mut buf)?;
```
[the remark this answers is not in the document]
augustocdias: but this object is discarded in the end... the state is changed but after it serializes it. The method then just drops this object.

crypto/src/external_commit.rs
```diff
@@ -0,0 +1,243 @@
...
+
+                    // receive the ack from the external join with outdated epoch
+                    // should fail because of the wrong epoch
+                    alice_group
+                        .decrypt_message(&message.to_bytes().unwrap(), &alice_backend)
+                        .await
+                        .unwrap_err();
+                    assert_eq!(alice_group.members().unwrap().len(), 2);
+
+                    // lets try again
```
beltram, line 201: isn't the rest of the test redundant with `test_join_by_external_commit` ?
You test method just wants to check what happens if you receive outdated external commit

crypto/src/external_commit.rs
```diff
@@ -0,0 +1,243 @@
...
+
+#[cfg(test)]
+mod tests {
+    use mls_crypto_provider::MlsCryptoProvider;
+
+    use crate::{credential::CredentialSupplier, member::ConversationMember};
+
+    wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
+
+    pub mod external_join {
```
beltram, line 89: Other tests in mind:
* What if `merge_pending_group_from_external_commit` while there's no pending group ?
* What happens if PublicGroupState is not the one of the group I am joining
* If `merge_pending_group_from_external_commit` fails, can I retry it (useful from a client POV)

Also in your tests, if you can, verify that at the end pending_groups are pruned from keystore

keystore/src/mls.rs
```diff
@@ -255,6 +261,41 @@ impl CryptoKeystoreMls for crate::connection::Connection {
             .map(|group: PersistedMlsGroup| (group.id.clone(), group.state.clone()))
             .collect())
     }
+
+    async fn mls_groups_get(&self, group_id: &[u8]) -> CryptoKeystoreResult<Vec<u8>> {
```
beltram, line 265: unused, you can safely remove

crypto/src/external_commit.rs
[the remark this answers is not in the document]
beltram: Yes I know. I  never said it does not work (in fact it does), I said it's misleading.
This method is aimed at serializing a group for being persisted in a database. We want to serialize for crossing ffi boundaries: we should own this. We might decide that instead of json we want the group to be serialized in xml for whatever reason.
Also, tomorrow, this save method might decide (for its own reasons which might be perfectly valid) to mutate the group before serializing it.
The long-term clean solution is to handle ourselves the (de)serialization, it shouldn't be tied to openmls

PR #60, opened 2022-07-13
docs: Improving docs of Core-Crypto - [CL-50]

augustocdias:
----

- The **PR Title**

----

# What's new in this PR?
This PR intends to add and improve the documentation of the crate. This is being done using rust docs

----
##### References


crypto/src/client.rs
```diff
@@ -98,6 +100,18 @@ fn identity_key(credentials: &CredentialBundle) -> Result<Vec<u8>, MlsError> {
...
 
 impl Client {
+    /// Initialize a new client.
+    ///
+    /// # Arguments
+    /// * `id` - id of the client
+    /// * `certificate_bundle` - an optional x509 certificate
+    /// * `backend` - the KeyStore to read identities from
+    ///
+    /// # Errors
```
beltram, line 110: I'm not sure documenting errors like this brings value. We will end up most of the time with 3 kind of errors: keystore, openmls & corecrypto.  
Instead, I think documenting the errors conditions would bring more value e.g.
```rust
/// # Errors
/// Fails if:
/// - [condition-1]
/// - [condition-2] etc..
```

crypto/src/client.rs
```diff
@@ -237,6 +276,13 @@ impl Client {
 
     /// Requests `count` keying material to be present and returns
     /// a reference to it for the consumer to copy/clone.
+    ///
+    /// # Arguments
+    /// * `count` - number of keys to generate
```
beltram, line 281:
```suggestion
    /// * `count` - number of [`KeyPackage`] to generate
```

crypto/src/conversation.rs
```diff
@@ -31,8 +31,10 @@ use crate::{
     CryptoError, CryptoResult, MlsCiphersuite, MlsError,
 };
 
+/// A type definition for the group/conversation id which is a byte array
```
beltram, line 34: Same remark as for ClientId

crypto/src/conversation.rs
```diff
@@ -31,8 +31,10 @@ use crate::{
...
 pub type ConversationId = Vec<u8>;
 
+/// The configuration parameters for a group/conversation
```
beltram, line 37: Document members

crypto/src/conversation.rs
```diff
@@ -59,6 +62,7 @@ impl MlsConversationConfiguration {
     }
 }
 
+/// This type will store the state of a group
```
beltram, line 65: Be a bit more descriptive about what a conversation is, its link with a MLS group etc..

crypto/src/conversation.rs
```diff
@@ -68,20 +72,25 @@ pub struct MlsConversation {
     configuration: MlsConversationConfiguration,
 }
 
+/// It is a wrapper for the welcome and the message returned when creating a group/conversation
```
beltram, line 75:
```suggestion
/// Returned when initializing a conversation. Different from conversation created from a [`Welcome`] message or an external commit.
```
Also document members

crypto/src/conversation.rs
```diff
@@ -175,10 +202,15 @@ impl MlsConversation {
...
 
+    /// Group/conversation id
     pub fn id(&self) -> &ConversationId {
         &self.id
     }
 
+    /// Returns all members credentials from the group/conversation
+    ///
+    /// # Errors
+    /// At the moment no errors can happen
```
beltram, line 213: Then change the return type. It should not be `Result`

crypto/src/conversation.rs
```diff
@@ -304,6 +361,17 @@ impl MlsConversation {
         Ok((message, welcome))
     }
 
+    /// Encrypts a raw payload then serializes it to the TLS wire format
```
beltram, line 364:
```suggestion
    /// Encrypts an Application Message then serializes it to the TLS wire format
```

crypto/src/external_commit.rs
```diff
@@ -31,6 +31,17 @@ impl MlsCentral {
     /// intended for example when a new client wants to join the user's existing groups.
     /// On success this function will return the group id and a message to be fanned out to other
     /// clients.
+    ///
+    /// # Arguments
+    /// * `group_state` - a verifiable public group state. it can be obtained by deserializing a TLS
+    /// serialized `PublicGoupState` object
+    ///
+    /// # Return type
+    /// It will return a tupple with the group/conversation id and the message containing the
```
beltram, line 40:
```suggestion
    /// It will return a tuple with the group/conversation id and the message containing the
```

crypto/src/lib.rs
```diff
@@ -14,6 +14,12 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see http://www.gnu.org/licenses/.
 
+//! Core Crypto is a wrapper on top of OpenMLS aimed to provide an ergonic API for usage in web
```
beltram, line 17:
```suggestion
//! Core Crypto is a wrapper on top of OpenMLS aimed to provide an ergonomic API for usage in web
```

crypto/src/lib.rs
```diff
@@ -112,6 +121,15 @@ mod config {
     }
 
     impl MlsCentralConfiguration {
+        /// Creates a new instance of the configuration.
+        ///
+        /// # Arguments
+        /// * `store_path` - location where the SQLite database will be stored
```
beltram, line 127:
```suggestion
        /// * `store_path` - location where the SQLite/IndexedDB database will be stored
```

crypto/src/lib.rs
```diff
@@ -112,6 +121,15 @@ mod config {
     }
 
     impl MlsCentralConfiguration {
+        /// Creates a new instance of the configuration.
```
beltram, line 124: An example would be great here ;)

crypto/src/lib.rs
```diff
@@ -223,33 +256,65 @@ impl MlsCentral {
...
+    /// # Errors
+    /// This function currently can't cause any errors
     pub fn callbacks(&mut self, callbacks: Box<dyn CoreCryptoCallbacks>) -> CryptoResult<()> {
         self.callbacks = Some(callbacks);
         Ok(())
     }
 
     /// Returns the client's public key as a buffer
+    /// # Errors
+    /// This function currently can't cause any errors
```
beltram, line 272: Then you should modify the return type

beltram (changes requested):

crypto/src/conversation.rs
```diff
@@ -68,20 +72,25 @@ pub struct MlsConversation {
...
+/// removal of other clients. It is returned when calling [core_crypto::MlsCentral::leave_conversation][leave_conversation]
 #[derive(Debug)]
 pub struct MlsConversationLeaveMessage {
     pub self_removal_proposal: MlsMessageOut,
     pub other_clients_removal_commit: Option<MlsMessageOut>,
 }
 
 impl MlsConversationCreationMessage {
-    /// Order is (welcome, message)
+    /// Serializes both wrapped objects into TLS and return them as a tupple of byte arrays.
```
OtaK, line 91:
```suggestion
    /// Serializes both wrapped objects into TLS and return them as a tuple of byte arrays.
```

crypto/src/conversation.rs
```diff
@@ -93,6 +102,16 @@ impl MlsConversationCreationMessage {
 }
 
 impl MlsConversation {
+    /// Creates a new group/conversation
+    ///
+    /// # Arguments
+    /// * `id` - group/conversation identifier
+    /// * `author_client` - the client responsible for creating the group
+    /// * `config` - group configuration
+    /// * `backend` - the KeyStore to persiste the group
```
OtaK, line 111: As @beltram suggested earlier, the `backend` / MLS Provider isn't just the keystore. There are sometimes other operations involved (such as calling the PRNG or crypto operations).
```suggestion
    /// * `backend` - MLS Provider that will be used to persist the group
```

crypto/src/conversation.rs
```diff
@@ -288,6 +335,16 @@ impl MlsConversation {
         Ok(message)
     }
 
+    /// Commits all pending proposals of the group
+    ///
+    /// # Arguments
+    /// * `backend` - the KeyStore to persist group changes
+    ///
+    /// # Return type
+    /// A tuple containing the a message and a possible welcome (when add proposals are in)
```
OtaK, line 344:
```suggestion
    /// A tuple containing the commit message and a possible welcome (in the case `Add` proposals were pending within the internal MLS Group)
```

crypto/src/lib.rs
```diff
@@ -70,13 +76,15 @@ use openmls_traits::OpenMlsCryptoProvider;
 use std::collections::HashMap;
 use tls_codec::{Deserialize, Serialize};
 
+/// This trait is used to provide authorization mechanisms for the MlsCentral struct, so that
```
OtaK, line 79: Not only. This is just the callback interface. Right now it only contains authorization, but it'll be expanded as needed (i.e. propagating the information that the client was removed from a MLS group through a commit).

crypto/src/lib.rs
OtaK: > Then you should modify the return type

We can do some housekeeping later on on those. Those methods return `Result`s because in the past all of those properties were behind `RwLock`s and `Mutex`es, so the operation *could* potentially fail in the case of a poisoned lock.

crypto/src/member.rs
```diff
@@ -34,6 +36,10 @@ pub struct ConversationMember {
 }
 
 impl ConversationMember {
+    /// Creates a new member from a TLS serialized byte array
```
OtaK, line 39:
```suggestion
    /// Creates a new member from a TLS-serialized keypackage
```

OtaK (changes requested):

crypto/src/client.rs
```diff
@@ -30,6 +30,7 @@ use openmls_traits::{
 
 pub(crate) const INITIAL_KEYING_MATERIAL_COUNT: usize = 100;
 
+/// Wrapper on the byte array used to identify a client id
```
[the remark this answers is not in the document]
OtaK: > Prefer functional description instead of technical one here i.e. explain what a client id is from Wire POV (unique client identifier, != userId) and also its format userID:clientID@domain and an example `b455a431-9db6-4404-86e7-6a3ebe73fcaf:3ae58155@mls.example.com`

Not sure we want this. This is only relevant in the context of Wire Cloud & CoreLogic interacting with each other. CoreCrypto intentionally performs no checks - it used to! (see here: https://github.com/wireapp/core-crypto/commit/7ba4635264c51c5900af656c420e59b66deae060#diff-35ce8a4538af9211db85cce68688f1c44d2aeeeaa1d0c3e834483ba7dec722e4L28)

But this was explicitely asked to be removed for CoreCrypto to treat any identifier as completely opaque, so we shouldn't document something that we never ever check

crypto/src/member.rs
```diff
@@ -23,8 +23,10 @@ use crate::{
 use openmls::prelude::KeyPackage;
 use tls_codec::Deserialize;
 
+/// Type definition for the Client/Member id. It is an array of bytes
```
beltram, line 26:
```suggestion
/// Type definition for the identifier of a client in a conversation (aka Member)
```

beltram (changes requested):
LGTM

PR #70, opened 2022-07-28
feat: ensure consistent state [CL-37]

beltram:
----

- The **PR Title**

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

* Local state machine (even though no transition ever happens) to validate we can perform the requested operation on a group e.g. make sure there is no pending proposal before encrypting a message.
* Start reorganized the code in a functional way i.e. everything related to commits goes in a dedicated mod
* Discarded all automatic commit merges every time a commit was created
  * Now a commit has to be `commit_accepted` in order to be merged
  * Public API remains the same. This change is just internal for the moment
* Implemented a `leave_conversation` operation which creates a proposal for removing self not requiring the other commit (the one removing other clients) to be merged

----
##### References


crypto-ffi/src/generic.rs
```diff
@@ -373,11 +373,13 @@ impl CoreCrypto<'_> {
...
                     .map_err(|_| CryptoError::LockPoisonError)?
-                    .encrypt_message(conversation_id, message),
+                    .encrypt_message(&conversation_id, message),
             ),
         )
     }
 
+    #[deprecated(note = "Will be covered by decrypt_message return type")]
+    #[allow(deprecated)]
     pub fn conversation_exists(&self, conversation_id: ConversationId) -> bool {
```
OtaK, line 383: Why deprecate this method? It's useful for very simple & cheap checks (in cases for instance where the backend would mis-order the arrival of websocket events and welcome messages, to inform CoreLogic's decision making)

crypto-ffi/src/generic.rs
```diff
@@ -539,4 +541,27 @@ impl CoreCrypto<'_> {
 
         Ok(())
     }
+
+    // Preserves "current" behaviour with auto-merged commits
+    // TODO: remove when backend counterpart implemented
+    #[allow(clippy::await_holding_lock)]
+    async fn maybe_accept_commit<T>(
```
OtaK, line 548: No async methods in this file; even if they are internal, they might deadlock because you're synchronously locking a `Mutex` then awaiting on it, probably locking it for `'static`

OtaK (changes requested):

crypto-ffi/src/generic.rs
beltram: Ok I thought about moving this to decrypt_message and return a boolean to inform CoreLogic when the client has been "self removed" from the group.
But ok to keep it.

PR #80, opened 2022-08-05
expose external senders at conversation creation for external remove proposal support [CL-53]

beltram:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Consumers can now pass an Ed25519 public key when creating a conversation. This "remove key" is the backend's key which will be used to validate external remove proposals

----
##### References

### DIFF
--- a/crypto-ffi/bindings/js/CoreCrypto.ts
+++ b/crypto-ffi/bindings/js/CoreCrypto.ts
@@ -120,7 +120,7 @@ export interface MemberAddedMessages {
 export interface CommitBundle {
     /**
      * TLS-serialized MLS Commit that needs to be fanned out to other (existing) members of the conversation
-     * 
+     *
      * @readonly
      */
     message: Uint8Array;
@@ -191,16 +191,16 @@ export class CoreCrypto {
 
     /**
      * This is your entrypoint to initialize {@link CoreCrypto}!
-     * 
+     *
      * @param params - {@link CoreCryptoParams}
-     * 
+     *
      * @example
      * ## Simple init
      * ```ts
      * const cc = await CoreCrypto.init({ databaseName: "test", key: "test", clientId: "test" });
      * // Do the rest with `cc`
      * ```
-     * 
+     *
      * ## Custom Entropy seed init & wasm file location
      * ```ts
      * // FYI, this is the IETF test vector #1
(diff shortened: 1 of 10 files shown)


crypto-ffi/bindings/js/CoreCrypto.ts
```diff
@@ -27,7 +27,7 @@ export interface ConversationConfiguration {
     admins?: Uint8Array[];
     ciphersuite?: Ciphersuite;
     keyRotationSpan?: number;
-    externalSenders: Uint8Array[];
+    externalSenders: Array<Uint8Array[]>;
```
OtaK, line 30: An array of array of buffers (= byte arrays)?
That seems confusing from an API standpoint, might also be incorrect?

Can we express it better?

crypto/Cargo.toml
```diff
@@ -40,6 +40,7 @@ rstest_reuse = "0.4"
 rcgen = { version = "0.9", features = ["x509-parser"] }
 async-std = { version = "1.12", features = ["attributes"] }
 futures-util = { version = "0.3", features = ["std", "alloc"] }
+ed25519-dalek = "1.0"
```
OtaK, line 43: Maybe we can find a way to call into the MLS Crypto provider for this? It already depends on `ed25519-dalek`

crypto/src/lib.rs
```diff
@@ -22,6 +22,8 @@
 #![deny(missing_docs)]
 #![allow(clippy::single_component_path_imports)]
 
+extern crate core;
```
OtaK, line 25: This shouldn't be needed

OtaK (dismissed):

crypto-ffi/bindings/js/CoreCrypto.ts
beltram: Yes for the moment it's an `Array<PublicKey>`, later on it will be an `Array<X509Certificate>`.  
In both cases it will be a `Vec<Vec<u8>>` received from DS that is just passed along to CoreCrypto which will do the parsing. Since this value will never be introspected by CoreLogic I think a byte array suits

OtaK: What I mean is that currently, `Array<Uint8Array[]>` translates to `Vec<Vec<Vec<u8>>>`

Because `Uint8Array` is a `Vec<u8>`, `Uint8Array[]` is a `Vec<Vec<u8>>` and `Array<Uint8Array[]>` is `Vec<Vec<Vec<u8>>>`

PR #86, opened 2022-08-09
fix: Catch up WASM bindings with the internal FFI api

OtaK:


crypto-ffi/bindings/js/CoreCrypto.ts
```diff
@@ -277,6 +312,17 @@ export class CoreCrypto {
         return await this.#cc.conversation_exists(conversationId);
     }
 
+    async exportGroupState(conversationId: ConversationId): Promise<
```
beltram, line 315: typo

beltram (approved):

crypto-ffi/bindings/js/CoreCrypto.ts
OtaK: Half-written code : D

PR #90, opened 2022-08-11
Verify our operations are durable [CL-63]

beltram:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Sometimes, we might forget to persist a MLS group after an operation. This macro will erase the current in memory group and restore it from keystore. If the methods is tested enough and forgets to persist the group (while it should) then the unit tests will catch that

----
##### References

### DIFF
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -21,7 +21,7 @@ openmls_traits = "0.1"
 # FIXME: This is being pulled only because of flaky error types from openmls
 tls_codec = "0.2"
 serde_json = "1.0"
-derive_more = { version = "0.99", features = ["deref", "from", "display"] }
+derive_more = { version = "0.99", features = ["deref"] }
 
 [dependencies.core-crypto-keystore]
 version = "0.3"
@@ -30,10 +30,6 @@ path = "../keystore"
 [dependencies.mls-crypto-provider]
 version = "0.3"
 path = "../mls-provider"
-
-[dependencies.core-crypto-attributes]
-version = "0.3"
-path = "../crypto-attributes"
 
 [dev-dependencies]
 uuid = { version = "1.0", features = ["v4", "v5"] }
@@ -45,6 +41,10 @@ rstest_reuse = "0.4"
 rcgen = { version = "0.9", features = ["x509-parser"] }
 async-std = { version = "1.12", features = ["attributes"] }
 futures-util = { version = "0.3", features = ["std", "alloc"] }
+
+[dev-dependencies.core-crypto-attributes]
+version = "0.3"
+path = "../crypto-attributes"
 
 [package.metadata.wasm-pack.profile.release]
 wasm-opt = ["-Os", "--enable-mutable-globals", "--enable-threads", "--detect-features"]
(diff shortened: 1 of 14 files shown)


crypto/src/conversation/handshake.rs
```diff
@@ -52,41 +52,53 @@ impl MlsProposalBundle {
 /// Creating proposals
 impl MlsConversation {
     /// see [openmls::group::MlsGroup::propose_add_member]
+    #[crate::durable]
```
OtaK, line 55: Maybe we switch all the `#[crate::durable]` to `#[cfg_attr(test, crate::durable)]` ?
I don't think we don't want to inflate codesize with test utils

crypto/src/conversation/handshake.rs
```diff
@@ -759,26 +908,74 @@ pub mod tests {
...
+                        // not be referenced in commit
+                        alice_central
+                            .decrypt_message(&id, commit.to_bytes().unwrap())
                             .await
-                            .is_ok());
+                            .unwrap();
+                        assert!(!alice_central[&id].group.members().contains(&&alice_key));
+
+                        // ensuring both can encrypt messages
+                        // assert!(alice_central.talk_to(&id, &mut bob_central).await.is_ok());
```
OtaK, line 978: This probably should be uncommented

crypto/src/conversation/mod.rs
```diff
@@ -43,6 +43,7 @@ use crate::{
 
 mod commit_delay;
 pub mod decrypt;
+mod durability;
```
OtaK, line 46: Maybe move the `#[cfg(test)]` here instead of `#![cfg(test)]` in `durability.rs`?

crypto/src/test_utils.rs
```diff
@@ -156,7 +156,6 @@ impl MlsCentral {
         self.commit_accepted(id).await?;
         assert_eq!(self[id].members().len(), size_before + 1);
         assert_eq!(other[id].members().len(), size_before + 1);
-        self.talk_to(id, other).await?;
```
OtaK, line 159: I don't think this is intended?

crypto/Cargo.toml
```diff
@@ -21,7 +21,8 @@ openmls_traits = "0.1"
 # FIXME: This is being pulled only because of flaky error types from openmls
 tls_codec = "0.2"
 serde_json = "1.0"
-derive_more = { version = "0.99", features = ["deref"] }
+derive_more = { version = "0.99", features = ["deref", "from", "display"] }
+paste = "1.0"
```
OtaK, line 25: Isn't `paste` a dependency of the proc_macro? We probably don't need it in the `crypto` crate

crypto/Cargo.toml
```diff
@@ -31,6 +32,10 @@ path = "../keystore"
 version = "0.3"
 path = "../mls-provider"
 
+[dependencies.core-crypto-attributes]
```
OtaK, line 35: Since it's only used in test configuration, maybe move it to `dev-dependencies`?

OtaK (changes requested):

PR #99, opened 2022-08-30
returnTLS serialized CommitBundle [CL-79]

beltram:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

All operations returning commits now return a byte array with the commit bundle serialized instead of a struct.
`PublicGroupState` has also been bundled to prepare it for future evolutions.

----
##### References

### DIFF
--- a/crypto/src/conversation/public_group_state.rs
+++ b/crypto/src/conversation/public_group_state.rs
@@ -1,0 +1,137 @@
+use std::io::{Read, Write};
+
+use openmls::prelude::{PublicGroupState, Verifiable, VerifiablePublicGroupState};
+use tls_codec::{Error, TlsDeserialize, TlsSerialize, TlsSize};
+
+use crate::{CryptoResult, MlsError};
+
+/// # PublicGroupStateEncryption
+///
+/// In order to guarantee confidentiality of the [PublicGroupState] on the wire a domain can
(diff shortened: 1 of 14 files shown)


crypto/src/conversation/public_group_state.rs
```diff
@@ -0,0 +1,138 @@
...
+    Full = 1,
+    /// Contains [PublicGroupState] changes since previous epoch (Not yet implemented)
+    /// (see [draft](https://github.com/rohan-wire/ietf-drafts/blob/main/mahy-mls-ratchet-tree-delta/draft-mahy-mls-ratchet-tree-delta.md))
+    Delta = 2,
+    ByRef = 3,
+}
+
+/// Represents the byte array in [PublicGroupStateBundle]
+#[derive(Debug)]
+#[non_exhaustive]
```
augustocdias, line 56: Do we really need this? This is likely to be used only internally in the crate and all our consumers don't use rust directly so it will make no difference for them.

augustocdias (approved):

typfel:
But we can't release this yet since the BE is not ready to consume this.

crypto/src/conversation/public_group_state.rs
beltram: good catch

PR #112, opened 2022-09-13
chore: E2E testing Rust <=> WASM [CL-88]

OtaK:


OtaK:
Note to self: use https://github.com/ad4mx/spinoff for progress report

interop/src/clients/mod.rs
```diff
@@ -0,0 +1,39 @@
+use color_eyre::eyre::Result;
+
+pub mod native;
+pub mod web;
+
+#[derive(Debug)]
+#[non_exhaustive]
```
beltram, line 7: You probably don't need this except if `EmulatedClientType` is used outside this crate
```suggestion
```

interop/src/clients/web.rs
```diff
@@ -0,0 +1,151 @@
...
+        let client_config = serde_json::json!({
+            "databaseName": format!("db-{client_id_str}"),
+            "key": "test",
+            "clientId": client_id_str
+        });
+        let browser = crate::build::web::webdriver::setup_browser(driver_addr).await?;
+
+        let _ = browser
+            .execute_async(
+                r#"
```
beltram, line 24: It would be neat to have those in separate typescript files nearby and import them with `include_str!`. WDYT ?

beltram (approved):

interop/src/clients/web.rs
OtaK: That's exactly what I used to do at the beginning, but I ditched the approach - it's a real PITA when you're refactoring and moving things around.
I think this could come back later if the JS becomes more complex. Right now it's just a couple of lines per "file", I don't think it's necessary

PR #168, opened 2022-11-03
fix!: Incorrect handling of enums across WASM FFI [CL-104]

OtaK:


crypto-ffi/bindings/js/test/CoreCrypto.test.js
```diff
@@ -27,7 +90,7 @@ test("init", async () => {
     return CoreCrypto.version();
   });
 
-  expect(version).toBe("0.5.1");
+  expect(version).toMatch("0.6.0");
```
beltram, line 93: Maybe we could relax this test, it could be cumbersome to edit this line every time we bump the version

beltram (approved):

crypto-ffi/bindings/js/test/CoreCrypto.test.js
OtaK: Those tests will become irrelevant once we have the new runner so we can keep it as is right now; As stuff will be written in rust I'll be able to parse Cargo.toml files and extract versions to never have to edit stuff again :D

PR #176, opened 2022-11-10
bench: more ciphersuites and compare MLS to Proteus

beltram:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Add ciphersuites 2, 3 & 7  
Compare MLS to Proteus

----
##### References

### DIFF
--- a/crypto/benches/mls_proteus.rs
+++ b/crypto/benches/mls_proteus.rs
@@ -1,0 +1,268 @@
+use crate::utils::*;
+use core_crypto::mls::MlsCiphersuite;
+use core_crypto::prelude::CertificateBundle;
+use criterion::{
+    async_executor::FuturesExecutor, black_box, criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion,
+};
+use futures_lite::future::block_on;
+use proteus::keys;
+use proteus::keys::{PreKey, PreKeyBundle};
+use rand::distributions::{Alphanumeric, DistString};
+
+#[path = "utils/mod.rs"]
+mod utils;
+
+fn mls_cases() -> Vec<(MlsCiphersuite, Option<CertificateBundle>, bool, &'static str)> {
+    // Ciphersuite 3 is the closest to proteus one
+    const CIPHERSUITE: MlsTestCase = MlsTestCase::Basic_Ciphersuite3;
+    let (_, ciphersuite, credential) = CIPHERSUITE.get();
+    let in_memory = (ciphersuite, credential.clone(), true, "MLS/mem");
+    let in_db = (ciphersuite, credential, false, "MLS/db");
+    if cfg!(feature = "bench-in-db") {
+        vec![in_memory, in_db]
+    } else {
+        vec![in_memory]
+    }
+}
+
+fn proteus_cases() -> Vec<(bool, &'static str)> {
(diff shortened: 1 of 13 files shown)


crypto/benches/mls_proteus.rs
```diff
@@ -0,0 +1,268 @@
...
+                                central.session_save(&keystore, session_id).await.unwrap();
+                            });
+                        }
+                        let text = Alphanumeric.sample_string(&mut rand::thread_rng(), MSG_MAX);
+                        (central, keystore, session_material, text)
+                    },
+                    |(mut central, keystore, session_material, text)| async move {
+                        for (session_id, _) in session_material {
+                            black_box(central.encrypt(&session_id, text.as_bytes()).unwrap());
+                            black_box(central.session_save(&keystore, &session_id).await.unwrap());
```
OtaK, line 79: Not sure if the session_save call is needed here as you're just encrypting payloads. Might lead to false I/O overhead.

crypto/benches/mls_proteus.rs
```diff
@@ -0,0 +1,268 @@
...
+                        for (session_id, _) in session_material {
+                            // generate a new prekey
+                            let pkb = black_box(
+                                PreKeyBundle::new(
+                                    keys::IdentityKeyPair::new().public_key,
+                                    &PreKey::new(keys::PreKeyId::new(2)),
+                                )
+                                .serialise()
+                                .unwrap(),
+                            );
```
OtaK, line 245: Could this be moved in the setup code? This also means we're benching identity & prekey generation

crypto/benches/mls_proteus.rs
```diff
@@ -0,0 +1,268 @@
...
+                            // replace existing session
+                            black_box(central.session_from_prekey(&session_id, &pkb).await.unwrap());
```
OtaK, line 247: Maybe more atomic would be the use case we're looking for?
```suggestion
                            black_box(central.session_delete(&keystore, &session_id).await.unwrap());
                            black_box(central.session_from_prekey(&session_id, &pkb).await.unwrap());
```

OtaK (changes requested):

crypto/benches/mls_proteus.rs
beltram: I think saving the session is required because `encrypt` requires a mutable session (to ratchet [here](https://github.com/wireapp/proteus/blob/develop/src/internal/session.rs#L864) I think).  
I wanted to test both in "safe" mode (we also persist the group in MLS after encrypting) to have a fair benchmark

crypto/benches/mls_proteus.rs
beltram: Yes but to be fair we should also bench new keys generation since MLS also does that by deriving new keys. I did not find an API for deriving new prekeys in proteus so I guessed we would have to regenerate them from scratch

OtaK: You can just reuse the same identity keypair for all of the prekeys - this assumes they all come from the same client

But I think you could still generate the keypairs ahead of time and use them as a parameter; even if it makes you generate dozens of them ahead of time, this should be fine (each of them is 96 bytes in memory)

OtaK (approved):

PR #184, opened 2022-11-25
feat: specialized decryption error

beltram:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

* Have a specialized error when trying to decrypt from wrong epoch hence trying to violate forward secrecy
* Bring back `out_of_order_tolerance` and `maximum_forward_distance` in our custom configuration, just for testing purposes now

----
##### References

### DIFF
--- a/crypto-ffi/src/CoreCrypto.udl
+++ b/crypto-ffi/src/CoreCrypto.udl
@@ -73,6 +73,7 @@ enum CryptoError {
     "UnauthorizedExternalCommit",
     "InvalidHashReference",
     "GenerationOutOfBound",
+    "DecryptionError",
     "HexDecodeError",
     "ProteusError",
     "CryptoboxMigrationError",
--- a/crypto/src/error.rs
+++ b/crypto/src/error.rs
@@ -101,6 +101,9 @@ pub enum CryptoError {
     /// implementation has a loophole and it'd better be evicted from the group.
     #[error("Decrypted an application message twice")]
     GenerationOutOfBound,
+    /// Tried to decrypt a message in the wrong epoch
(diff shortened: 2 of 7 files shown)


crypto/src/error.rs
```diff
@@ -101,6 +101,9 @@ pub enum CryptoError {
     /// implementation has a loophole and it'd better be evicted from the group.
     #[error("Decrypted an application message twice")]
     GenerationOutOfBound,
+    /// Tried to decrypt a message in the wrong epoch
+    #[error("Decrypted an application message from the wrong epoch")]
+    DecryptionError,
```
OtaK, line 106: The new variant needs to be added to the FFI UDL file as well

OtaK (changes requested):

crypto/src/error.rs
beltram: Good catch I forgot that one

OtaK (approved):

crypto-ffi/src/CoreCrypto.udl
```diff
@@ -73,6 +73,7 @@ enum CryptoError {
     "UnauthorizedExternalCommit",
     "InvalidHashReference",
     "GenerationOutOfBound",
+    "DecryptionError",
```
typfel, line 76: Wouldn't it make sense to make the name more specific given that it only happen when local epoch is out sync?

typfel: Ah too slow, already merged

typfel:
Is this error only happening for application messages?

PR #194, opened 2022-12-13
fix: Broken Proteus implementation [CL-110]

OtaK:


crypto/src/proteus.rs
```diff
@@ -987,18 +1033,23 @@ mod tests {
...
         assert_eq!(&decrypted, &message[..]);
 
-        // FIXME: Known bug, see CL-110
-        // let encrypted = bob.encrypt(&session_id, &message[..]);
-        // let decrypted = proteus_central
-        //     .decrypt(&mut keystore, &session_id, &encrypted)
-        //     .await
-        //     .unwrap();
-        // assert_eq!(&decrypted, &message[..]);
+        // CL-110 assertion
```
beltram, line 1044:
```suggestion
        // CL-110 assertion
        // Happens when a migrated client ratchets just after migration. It does not happen when ratchet is not required
        // Having alice(A), bob(B) and migration(M), you can reproduce this behaviour with `[A->B][B->A] M [A->B][B->A]`
        // However you won't reproduce it like this because migrated alice does not ratchet `[A->B][B->A] M [B->A][A->B]`
```

beltram (approved):

PR #204, opened 2022-12-19
chore: xtask improvements

### LINKED ISSUE #102: Tool: Improvements for xTask
- Maybe pull all refs from the remote before generating the changelog so it will be generated correctly (if not at least print a warning message so the user doesn't forget)
- Bump should update also the dependencies for the updates crates inside the project (core-crypto depends on internal crates pointing to old versions for ex)
- Bump should also bump the `package.json`, the `build.gradle.kts`

OtaK:
Closes #102


augustocdias (approved):
LGTM

PR #209, opened 2023-01-03
test(wip): WASM Test runner: WebDriver BiDi edition

OtaK:


OtaK:
#162 contains the BiDi code. It's going to run disabled for now until I make time to fix the BiDi implementation timeouts

PR #212, opened 2023-01-11
feat: Proteus errors

OtaK:
Most of the work is in https://github.com/wireapp/proteus/tree/otak/2.0-error-codes

In particular in https://github.com/wireapp/proteus/blob/otak/2.0-error-codes/crates/proteus-traits/src/lib.rs#L27-L70

which lists all the error codes. 

I also added in the proteus code inline comments to highlight when we trigger error codes in certain situations such as here: https://github.com/wireapp/proteus/blob/otak/2.0-error-codes/src/internal/session.rs#L266

### DIFF
--- a/.github/workflows/bindings.yml
+++ b/.github/workflows/bindings.yml
@@ -48,6 +48,7 @@ jobs:
     runs-on: macos-latest
     env:
       CARGO_TERM_COLOR: always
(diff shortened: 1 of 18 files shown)


augustocdias (approved):
Lgtm

PR #230, opened 2023-01-25
hotfix: 0.6.0-rc6 [CL-129]

OtaK:
Because rc5 is botched

## Post-mortem

The issue was basically that the implemented system was not working - at all.

* The use of the `?` operator in Rust short-circuits code execution, so the current code was unable to save the Proteus error codes aside - as we used a `match` statement to retrieve and query the error for its code.
	* A fix for this was to instead "localize" the short-circuiting by wrapping our proteus code in IIFEs (Immediately Invoked Function Expressions) or IICs (Immediately Invoked Closure) for Rust-specific lingo. This allows us to retrieve the Error as-is and inspect it.
* The extended JS `Error` instance in Rust was a good idea, unfortunately it's 100% unsupported by the current tooling (`wasm_bindgen` namely). 
	* So I had the idea to instead serialize a "rich error" to JSON, pass it across the WASM FFI, and build the rich error within our TypeScript bindings by parsing this JSON payload. This will create more adherence if we ever modify the internal interface, but at least this is functioning.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,12 +8,12 @@ on: [push]
 
 env:
   RUST_BACKTRACE: 1
+  CARGO_TERM_COLOR: always
+  CARGO_NET_GIT_FETCH_WITH_CLI: true
 
 jobs:
   fmt:
     runs-on: ubuntu-latest
-    env:
-      CARGO_TERM_COLOR: always
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
@@ -24,8 +24,6 @@ jobs:
 
   check:
     runs-on: ubuntu-latest
-    env:
-      CARGO_TERM_COLOR: always
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
@@ -35,8 +33,6 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
-    env:
-      CARGO_TERM_COLOR: always
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
@@ -48,8 +44,6 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
-    env:
-      CARGO_TERM_COLOR: always
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
@@ -62,8 +56,6 @@ jobs:
 
   proteus-test:
     runs-on: ubuntu-latest
-    env:
-      CARGO_TERM_COLOR: always
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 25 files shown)


OtaK:
## Post-mortem

The issue was basically that the implemented system was not working - at all.

* The use of the `?` operator in Rust short-circuits code execution, so the current code was unable to save the Proteus error codes aside - as we used a `match` statement to retrieve and query the error for its code.
	* A fix for this was to instead "localize" the short-circuiting by wrapping our proteus code in IIFEs (Immediately Invoked Function Expressions) or IICs (Immediately Invoked Closure) for Rust-specific lingo. This allows us to retrieve the Error as-is and inspect it.
* The extended JS `Error` instance in Rust was a good idea, unfortunately it's 100% unsupported by the current tooling (`wasm_bindgen` namely). 
	* So I had the idea to instead serialize a "rich error" to JSON, pass it across the WASM FFI, and build the rich error within our TypeScript bindings by parsing this JSON payload. This will create more adherence if we ever modify the internal interface, but at least this is functioning.

crypto-ffi/bindings/js/CoreCrypto.ts
```diff
@@ -20,6 +20,81 @@ import wasm from "../../../crypto-ffi/Cargo.toml";
...
+        const mappedErrorPromise = p.catch((e: Error|CoreCryptoError) => {
+            if (e instanceof CoreCryptoError) {
+                throw e;
+            } else {
+                throw this.fromStdError(e);
+            }
+        });
+
+        return await mappedErrorPromise;
+    }
```
augustocdias, line 95: Not important since this seems to be hidden from the callers, but I'd remove this from this class. When its being called I have an impression that we're always returning an error but in fact we're returning a "Result" that is mapping the possible error to this.

OtaK: I isolated this into a function because of the sheer amount of times this code is called (51 to be accurate). And it didn't make sense to me to put it anywhere else than as a static method on the error class

JS `Promise` is basically a Rust `async` `Result<T, E>` anyway; you either resolve the Promise successfully - with a value - or you throw within the promise - with usually an instance of `Error` - So it made sense to me to make it work very much like Rust's `.map_err()`

augustocdias: I agree. My comment is in regards of calling it: you call `CoreCryptoError. asyncMapErr` which leaves an impression that an error will always be returned but in fact it's still the Promise/Result. My comment was more to extract it from the `CoreCryptoError` and put it in something called `CoreCrypto` or `CoreCryptoResult`. As I said, not important as our callers don't see this, so feel free to ignore

.github/workflows/rust.yml
```diff
@@ -129,25 +117,25 @@ jobs:
...
+      # - name: Install wasm-pack
+      #   run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
       - uses: Swatinem/rust-cache@v2
       - name: Run tests (wasm)
         run: |
-          wasm-pack test --chrome --headless keystore --features proteus-keystore -- proteus
-          wasm-pack test --chrome --headless crypto --features proteus,cryptobox-migrate -- proteus
+          cargo test --release --features proteus-keystore -p core-crypto-keystore -- proteus
+          cargo test --release crypto --features proteus,cryptobox-migrate -- proteus
+      # - name: Run tests (wasm)
```
beltram, line 132: undo

beltram (approved):

PR #261, opened 2023-02-20
feat!: Improve memory efficiency

OtaK:
Our previous `HashMap`-based cache could grow indefinitely in the case of massive accounts with many, many groups/conversations, each containing a ton of clients. This replaces this memory store by a LRU cache having the following properties: 

* Limited by number of entries AND occupied memory
	* Defaults for memory: All the available system memory on other platforms / 100MB on WASM
	* Defaults for number of entries: 
		* 100 MLS groups
		* 200 Proteus sessions
* Flow for retrieving a value
	1. Check the LRU store if the value exists, if yes, it's promoted as MRU (Most Recently Used) and returned
	2. If not found, it might have been evicted, so we search the keystore
	3. If found in the keystore, the value is placed as MRU and returned
		* Special case: we evict the store as much as needed to fit the new MRU value in this case. This is designed to infaillible.
	5. If not found, we return a `None` value
	
This approach potentially allows to have an unlimited number of groups/sessions as long as a single item does not exceed the maximum memory limit.

Note on the API breakage: This comes from the fact that some APIs became async due to the internal async lock used to perform proper inner mutability on the LRU map.

### DIFF
--- a/crypto/src/mls/conversation/decrypt.rs
+++ b/crypto/src/mls/conversation/decrypt.rs
@@ -189,17 +189,24 @@ impl MlsCentral {
         conversation_id: &ConversationId,
         message: impl AsRef<[u8]>,
     ) -> CryptoResult<MlsConversationDecryptMessage> {
-        let decrypt_message = Self::get_conversation_mut(&mut self.mls_groups, conversation_id)?
+        let decrypt_message = self
+            .get_conversation(conversation_id)
+            .await?
+            .write()
+            .await
             .decrypt_message(
                 message.as_ref(),
                 &self.mls_backend,
                 self.callbacks.as_ref().map(|boxed| boxed.as_ref()),
             )
             .await?;
+
         if !decrypt_message.is_active {
-            self.mls_groups
-                .remove(conversation_id)
-                .ok_or(CryptoError::ImplementationError)?;
+            // ? Do we wipe conversations or do we just prune the group from the cache - subsequent accesses will make it look like the group doesn't exist
+            self.wipe_conversation(conversation_id).await?;
+            // self.mls_groups
+            //     .remove(conversation_id)
+            //     .ok_or(CryptoError::ImplementationError)?;
         }
         Ok(decrypt_message)
     }
@@ -314,7 +321,7 @@ pub mod tests {
                             .add_members_to_conversation(&id, &mut [charlie.clone()])
                             .await
                             .unwrap();
-                        assert!(alice_central.pending_commit(&id).is_some());
+                        assert!(alice_central.pending_commit(&id).await.is_some());
 
                         let add_debbie_commit = bob_central
                             .add_members_to_conversation(&id, &mut [debbie.clone()])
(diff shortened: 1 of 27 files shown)


crypto/src/mls/conversation/decrypt.rs
```diff
@@ -189,17 +189,22 @@ impl MlsCentral {
...
                 message.as_ref(),
                 &self.mls_backend,
                 self.callbacks.as_ref().map(|boxed| boxed.as_ref()),
             )
             .await?;
         if !decrypt_message.is_active {
-            self.mls_groups
-                .remove(conversation_id)
-                .ok_or(CryptoError::ImplementationError)?;
+            self.wipe_conversation(conversation_id).await?;
```
beltram, line 204: Yes it's probably better this way as it does not leave a ghost conversation in the keystore. Maybe amend the documentation to let users know conversation could get wiped implicitly.

crypto/src/mls/conversation/decrypt.rs
```diff
@@ -326,10 +331,26 @@ pub mod tests {
                             .await
                             .unwrap();
                         // Now Debbie should be in members and not Charlie
-                        assert!(alice_central[&id].members().get(&debbie.id).is_some());
-                        assert!(alice_central[&id].members().get(&charlie.id).is_none());
+                        assert!(alice_central
```
beltram, line 334: It could probably be worth adding a test helper to dodge this boilerplate

beltram (changes requested):

crypto/src/mls/conversation/decrypt.rs
OtaK: Yeah, I'll need to document this change for sure - I was having the case where the GroupStore was working *too well* and pruning it in cache would cause it to function as usual (because the fallback mechanism of not having it in cache would fetch it in the keystore)

There's an alternative possible, is to set the group to active = false & not wipe it implicitely, I believe we have the guards everywhere to not perform actions on inactive groups.

crypto/src/mls/conversation/decrypt.rs
OtaK: I thought about it too - our previous helper in the form of implementing `Index`/`IndexMut` couldn't do it as this is `async`.

beltram: Yes obviously not relying on `Index(Mut)` ; more something like `MlsGroup::find_member(id: &[u8])`

crypto/src/mls/conversation/decrypt.rs
beltram: I'd be in favor of the alternative option: just mark the group inactive and not delete anything either on the keystore or in cache. openmls is supposed to provide us the guards that an inactive group cannot be operated

crypto/src/mls/conversation/decrypt.rs
OtaK: I added the following API to `MlsCentral`'s test_utils and called it in all tests: 

It's panicky but whatever

```rust
pub async fn get_conversation_unchecked(
    &mut self,
    conv_id: &ConversationId,
) -> async_lock::RwLockWriteGuard<'_, MlsConversation>;
```

beltram (approved):

PR #262, opened 2023-02-22
fix: Ignore incoming messages sent by self client [CL-140\

OtaK:


crypto/src/mls/conversation/decrypt.rs
```diff
@@ -116,6 +117,20 @@ impl MlsConversation {
                 }
             }
             ProcessedMessage::StagedCommitMessage(staged_commit) => {
+                // Don't process self commits - safeguard against buggy MLS DS
```
typfel, line 120: Can we also ignore application messages from self?

OtaK: We could but unsure how reliant apps are on broadcasting messages in the self-conversation. 

I'm perfectly seeing clients using the self conversation as a message broker (i.e. I change my status, gets broadcasted to all clients, when I get the self-message, local state changes) or to check delivery status (message is greyed out when sent to backend, becomes solid white when recieved back, kinda same as the 2-tick system on WhatsApp)

typfel: Delivery status is not relying on receiving your own message back.

We currently have these statuses:
Sent -> We client got 201 OK after POST:ing the message (server received the message)
Delivered -> We get a "delivered" protobuf ACK from the receiving client (recipient client received the message)
Read -> We get a "read" protobuf ACK from the receiving client (recipient client read the message)

What happens now is that we get a "duplicated message twice error" error for our own application messages.

OtaK: I'll see if it doesn't break anything - I tried ignoring all messages and it was breaking tests

OtaK: Okay so not possible. Application Message decryption is done with `process_unverified_message` only. Other side effects are outside of CoreCrypto. 
The thing is - we cannot know it's an application message until we decrypt it!

typfel: Don't we know the sender already before we call `process_unverified_message` at 

https://github.com/wireapp/core-crypto/blob/5a7545543635ba0d7d4293136a5d61ac3382b65a/crypto/src/mls/conversation/decrypt.rs#L67

OtaK: > Don't we know the sender already before we call `process_unverified_message` at
> 
> https://github.com/wireapp/core-crypto/blob/5a7545543635ba0d7d4293136a5d61ac3382b65a/crypto/src/mls/conversation/decrypt.rs#L67

We do, but as we can't tell if it's an `ApplicationMessage` this means we'll have to ignore Proposals as well; Ignoring proposals coming from self does break how MLS work (I tried at first to ignore *all* messages and it our tests caught it: https://github.com/wireapp/core-crypto/actions/runs/4243725403/jobs/7376851374)

typfel: If it's not possible for all messages it doesn't solve the problem though. An alternative would be that we get a distinct error on attempts to decrypt duplicate messages which we can ignore.

OtaK: Then I think we should close this. We literally have no way to know which messages we've seen in the past. We only see the message generation and throw a `GenerationOutOfBound` error when decrypting the same message generation twice.

This cannot be fixed unless the backend fixes it

OtaK:
Closing this for the foreseeable future.
The current conclusion is that this bug *has* to be solved on the backend (well it's a backend bug after all) because it might have undesirable effects on the clients.

PR #270, opened 2023-03-07
fix: Exclude self from self-remove-commit delay [CL-144]

OtaK:
This is a commit delay optimization to avoid situations where both statements are true:
* The self client is the group creator. (leaf index = 0)
* The self client is being removed from the group in a proposal. This usually results in unusally high commit delays for other members from the sole virtue of a member with a lower leaf index being removed giving them a high position, thus a high commit delay (15s+)

### DIFF
--- a/crypto/src/mls/conversation/commit_delay.rs
+++ b/crypto/src/mls/conversation/commit_delay.rs
@@ -1,20 +1,74 @@
+use openmls_traits::OpenMlsCryptoProvider;
+
 use super::MlsConversation;
+use crate::{mls::MlsCryptoProvider, MlsError};
+
+/// These constants intend to ramp up the delay and flatten the curve for later positions
+pub(self) const DELAY_RAMP_UP_MULTIPLIER: f32 = 120.0;
+pub(self) const DELAY_RAMP_UP_SUB: u64 = 106;
+pub(self) const DELAY_POS_LINEAR_INCR: u64 = 15;
+pub(self) const DELAY_POS_LINEAR_RANGE: std::ops::RangeInclusive<u64> = 1..=3;
 
 impl MlsConversation {
-    /// These constants intend to ramp up the delay and flatten the curve for later positions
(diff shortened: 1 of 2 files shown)


OtaK:
I'd like a triple review on this as the code is quite touchy and does some assumptions on the return order of ratchet tree keypackages WRT their leaf node indexes. Which in turn creates assumptions on the internals of OpenMLS

crypto/src/mls/conversation/commit_delay.rs
```diff
@@ -10,11 +13,61 @@ impl MlsConversation {
     /// * `self_index` - ratchet tree index of self client
     /// * `epoch` - current group epoch
     /// * `nb_members` - number of clients in the group
-    pub fn compute_next_commit_delay(&self) -> Option<u64> {
+    pub fn compute_next_commit_delay(&self, backend: &MlsCryptoProvider) -> Option<u64> {
```
beltram, line 16: You should add some unit test to verify this

crypto/src/mls/conversation/commit_delay.rs
```diff
@@ -10,11 +13,61 @@ impl MlsConversation {
...
-            let own_index = self.group.own_leaf_index() as u64;
+            let mut own_index = self.group.own_leaf_index() as u64;
+            let members = self.group.members();
+
+            // Look for members that were removed at the left of our tree in order to shift our own leaf index (post-commit tree visualization)
+            let left_tree_diff = members
+                .iter()
+                .enumerate()
+                .filter(|(leaf_index, _)| (*leaf_index as u64) < own_index)
+                .try_fold(0u32, |mut acc, (_, keypackage)| {
```
beltram, line 55:
```suggestion
                .take(own_index as usize)
                .try_fold(0u32, |mut acc, keypackage| {
```

beltram (changes requested):

beltram (approved):

augustocdias (approved):

PR #297, opened 2023-04-06
feat: decrypt messages returns identity

beltram:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Decrypting messages now returns identity in the CommitBundle
Also fixes an issue with URIs in x509 Subject Alternative Name converted to lowercase in acme server.

----
##### References

### DIFF
--- a/crypto-ffi/src/generic.rs
+++ b/crypto-ffi/src/generic.rs
@@ -150,6 +150,7 @@ pub struct DecryptedMessage {
     pub commit_delay: Option<u64>,
     pub sender_client_id: Option<ClientId>,
     pub has_epoch_changed: bool,
+    pub identity: Option<WireIdentity>,
 }
 
 impl TryFrom<MlsConversationDecryptMessage> for DecryptedMessage {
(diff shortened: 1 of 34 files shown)


crypto-ffi/src/generic.rs
```diff
@@ -351,26 +373,26 @@ impl CoreCrypto<'_> {
...
-                client_id.clone(),
-                ciphersuites,
-                None,
+        future::block_on(
+            self.executor.lock().map_err(|_| CryptoError::LockPoisonError)?.run(
+                self.central
+                    .lock()
+                    .map_err(|_| CryptoError::LockPoisonError)?
+                    // weird but avoids importing 'either' crate here.
+                    .mls_init(Err(client_id.clone()).into(), ciphersuites),
```
OtaK, line 382: You could import `either` safely here; It's already in `core-crypto`'s dependency graph so it would just resolve to the same compilation unit.

crypto/src/mls/credential.rs
```diff
@@ -23,84 +26,49 @@ pub struct CertificateBundle {
...
     /// Generates a supplier that is later turned into a [openmls::prelude::CredentialBundle]
-    pub fn rand_certificate_bundle() -> CredentialSupplier {
-        |cs: MlsCiphersuite| {
-            // generate the leaf certificate
-            let (leaf_kp, leaf_sk) = Self::key_pair(cs);
-            let leaf_params = Self::certificate_params(leaf_kp, false);
-            Some(Self::new_certificate_bundle_from_leaf(cs, leaf_params, leaf_sk))
-        }
+    pub fn rand_certificate_bundle_old() -> CredentialSupplier {
+        todo!()
```
OtaK, line 48: Careful with `todo!()`; It panics when ran so maybe we want to return a dummy `CredentialSupplier` containing descriptive labels that it's not implemented and contains a dummy value?

OtaK (approved):

crypto/src/mls/credential.rs
beltram: Oops, it's an omission. Thanks for catching.

crypto-ffi/src/generic.rs
beltram: Yes, I wanted to avoid duplicating dependencies declaration but you're right, the workaround is too ugly

PR #307, opened 2023-04-24
feat!: CoreCrypto draft-20 edition [WPB-1192]

OtaK:


crypto/src/mls/mod.rs
```diff
@@ -536,19 +531,23 @@ impl MlsCentral {
...
     /// # Errors
     /// If the conversation can't be found, an error will be returned. Other errors are originating
     /// from OpenMls and serialization
-    pub async fn export_public_group_state(&mut self, conversation_id: &ConversationId) -> CryptoResult<Vec<u8>> {
+    pub async fn export_group_info(&mut self, conversation_id: &ConversationId) -> CryptoResult<Vec<u8>> {
         let conversation = self.get_conversation(conversation_id).await?;
+        let conversation_lock = conversation.read().await;
+        let group = &conversation_lock.group;
+        let cs = group.ciphersuite();
+        let ct = group.own_leaf().unwrap().credential().credential_type();
```
OtaK, line 544: This unwrap is safe (because you can only use this method if you're a member of the group) but we will need to clean it up at some point.

crypto/src/lib.rs
```diff
@@ -22,6 +22,7 @@
 #![doc = include_str!("../../README.md")]
 #![deny(missing_docs)]
 #![allow(clippy::single_component_path_imports)]
+#![allow(ambiguous_glob_reexports)]
```
augustocdias, line 25: why are we allowing that?

beltram (approved):

crypto/src/lib.rs
beltram: To spare some time and mostly because our prelude is never consumed externally so it's not a big deal

PR #319, opened 2023-05-16
fix: Reload proteus sessions when `restore_from_disk` is called - CL-168

augustocdias:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

When `restore_from_disk` is called, it will restore proteus sessions too if the feature flag is enabled.

----
##### References

### DIFF
--- a/crypto/src/proteus.rs
+++ b/crypto/src/proteus.rs
@@ -93,6 +93,18 @@ impl CoreCrypto {
         Ok(())
     }
 
+    /// Reloads the sessions from the key store
+    ///
+    /// Warning: The Proteus client **MUST** be initialized with [CoreCrypto::proteus_init] first or it will do nothing
+    pub async fn proteus_reload_sessions(&mut self) -> CryptoResult<()> {
+        if let Some(proteus) = self.proteus.as_mut() {
+            let keystore = self.mls.mls_backend.borrow_keystore();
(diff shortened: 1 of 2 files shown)


crypto/src/proteus.rs
```diff
@@ -93,6 +93,15 @@ impl CoreCrypto {
         Ok(())
     }
 
+    /// Reloads the sessions from the key store
+    ///
+    /// Warning: The Proteus client **MUST** be initialized with [CoreCrypto::proteus_init] first or an error will be returned
+    pub async fn proteus_reload_sessions(&mut self) -> CryptoResult<()> {
```
typfel, line 99: Does this mean that it will throw an error if I run an mls only instance of CC and call `restoreFromDisk`?  What about just making it a no-op if proteus is not initialised?

typfel: Yes, we do on Android.

typfel: And we only have one type of binary which has MLS and proteus enabled.

augustocdias: Ok. If it isn't initialised it will do nothing instead of failing

OtaK (commented):
If you wish, there's another way to go around that. You could just clear the cached proteus sessions and they will be loaded on-demand by the GroupStore

PR #341, opened 2023-06-20
feat!: PostQuantum Ciphersuite [WPB-2168]

### LINKED ISSUE #240: Post-Quantum crypto support
Seems to me we need to implement hybrid kyberXX & ed25519 KEM

https://github.com/Argyle-Software/kyber

OtaK:
Closes #240


crypto/src/mls/client/key_package.rs
```diff
@@ -158,19 +160,22 @@ impl Client {
             // let nb_kpb = to_generate * self.identities.count();
 
             let nb_kpb = to_generate;
+            let mut kps = Vec::with_capacity(nb_kpb);
+
+            for _ in 0..nb_kpb {
```
beltram, line 165: The problem with this is that it generates a KeyPackage only for 1 CredentialType. However, a client can simultaneously have 1 Basic CredentialBundle and 1 X509 one (for example, migrating some of its conversations to e2e identity but keeping some with Basic because not all Group members support x509 credentials).

But actually I had planned to delegate this to CoreLogic so I think you should fix by adding a `credential_type` argument to `get_or_create_client_keypackages`

beltram (changes requested):

beltram (approved):

PR #346, opened 2023-06-29
feat: Include certificate roots and certificate policy in GroupContext - WPB-1188

augustocdias:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Include certificate roots and certificate policy in GroupContext and expose to API

----
##### References

### DIFF
--- a/crypto-ffi/bindings/js/CoreCrypto.ts
+++ b/crypto-ffi/bindings/js/CoreCrypto.ts
@@ -161,6 +161,25 @@ export interface ConversationConfiguration {
      * Implementation specific configuration
      */
     custom?: CustomConfiguration;
+    /**
+     * Trust anchors to be added in the group's context extensions
+     */
+    perDomainTrustAnchors?: PerDomainTrustAnchor[];
+}
+
+/**
+ * A wrapper containing the configuration for trust anchors to be added in the group's context
+ * extensions
+ */
+export interface PerDomainTrustAnchor {
+    /**
+     * Domain name of the owning backend this anchor refers to. One of the certificate in the chain has to have this domain in its SANs
+     */
+    domain_name: string,
+    /**
+     * PEM encoded (partial) certificate chain. This contains the certificate chain for the CA certificate issuing the E2E Identity certificates
+     */
+    intermediate_certificate_chain: string,
 }
 
 /**
@@ -884,11 +903,13 @@ export class CoreCrypto {
         configuration: ConversationConfiguration = {}
     ) {
         try {
-            const { ciphersuite, externalSenders, custom = {} } = configuration || {};
+            const { ciphersuite, externalSenders, custom = {}, perDomainTrustAnchors = [] } = configuration || {};
             const config = new CoreCrypto.#module.ConversationConfiguration(
                 ciphersuite,
                 externalSenders,
                 custom?.keyRotationSpan,
+                custom?.wirePolicy,
+                perDomainTrustAnchors as any[],
             );
             const ret = await CoreCryptoError.asyncMapErr(this.#cc.create_conversation(conversationId, creatorCredentialType, config));
             return ret;
@@ -951,6 +972,47 @@ export class CoreCrypto {
             conversationId,
             message
         ));
+    }
+
+    /**
+     * Updates the trust anchors for a conversation. This should be called when a federated event happens (new team added/removed).
+     * Clients should add and/or remove trust anchors from the new backend to the conversation. The method will check
+     * for duplicated domains and the validity of the certificate chain.
+     *
+     * **CAUTION**: {@link CoreCrypto.commitAccepted} **HAS TO** be called afterwards **ONLY IF** the Delivery Service responds
+     * '200 OK' to the {@link CommitBundle} upload. It will "merge" the commit locally i.e. increment the local group
+     * epoch, use new encryption secrets etc...
+     *
+     * @param conversationId - The ID of the conversation
+     * @param removeDomainNames - Domains to remove from the trust anchors
+     * @param addTrustAnchors - New trust anchors to add to the conversation
+     *
+     * @returns A {@link CommitBundle}
+     */
+    async update_trust_anchors_from_conversation(conversationId: ConversationId, removeDomainNames: string[], addTrustAnchors: PerDomainTrustAnchor[]): Promise<CommitBundle> {
+        try {
+            const ffiRet: CoreCryptoFfiTypes.CommitBundle = await CoreCryptoError.asyncMapErr(this.#cc.update_trust_anchors_from_conversation(
+                conversationId,
+            removeDomainNames,
+              addTrustAnchors
+            ));
+
+            const gi = ffiRet.group_info;
+
+            const ret: CommitBundle = {
+                welcome: ffiRet.welcome,
+                commit: ffiRet.commit,
+                groupInfo: {
+                    encryptionType: gi.encryption_type,
+                    ratchetTreeType: gi.ratchet_tree_type,
+                    payload: gi.payload
+                },
+            };
+
+            return ret;
+        } catch(e) {
+            throw CoreCryptoError.fromStdError(e as Error);
+        }
     }
 
     /**
--- a/crypto-ffi/src/CoreCrypto.udl
+++ b/crypto-ffi/src/CoreCrypto.udl
@@ -67,6 +67,12 @@ dictionary ConversationConfiguration {
     Ciphersuite ciphersuite;
     sequence<bytes> external_senders;
     CustomConfiguration custom;
+    sequence<PerDomainTrustAnchor> per_domain_trust_anchors;
+};
+
+dictionary PerDomainTrustAnchor {
+    string domain_name;
+    string intermediate_certificate_chain;
 };
 
 dictionary CustomConfiguration {
@@ -179,6 +185,14 @@ enum CryptoError {
     "ClearingPendingCommitError",
     "SelfCommitIgnored",
     "UnmergedPendingGroup",
+    "X509CertDerError",
+    "PemError",
+    "DomainNameNotFound",
+    "DomainNamesDontMatch",
+    "DuplicateDomainName",
+    "InvalidCertificateChain",
+    "EmptyTrustAnchorUdpate",
+    "DuplicateCertificateChain",
 };
 
 enum MlsWirePolicy {
--- a/crypto-ffi/src/wasm.rs
+++ b/crypto-ffi/src/wasm.rs
@@ -739,6 +739,25 @@ pub struct ConversationConfiguration {
     ciphersuite: Option<Ciphersuite>,
     external_senders: Vec<Vec<u8>>,
     custom: CustomConfiguration,
+    per_domain_trust_anchors: Vec<PerDomainTrustAnchor>,
+}
+
+#[wasm_bindgen]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct PerDomainTrustAnchor {
+    domain_name: String,
+    intermediate_certificate_chain: String,
+}
+
+#[wasm_bindgen]
+impl PerDomainTrustAnchor {
+    #[wasm_bindgen(constructor)]
+    pub fn new(domain_name: String, intermediate_certificate_chain: String) -> Self {
+        Self {
+            domain_name,
+            intermediate_certificate_chain,
+        }
+    }
 }
 
 #[wasm_bindgen]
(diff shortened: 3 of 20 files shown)


crypto-ffi/bindings/js/CoreCrypto.ts
```diff
@@ -847,11 +866,13 @@ export class CoreCrypto {
...
     ) {
         try {
-            const {ciphersuite, externalSenders, custom = {}} = configuration || {};
+            const {ciphersuite, externalSenders, custom = {}, perDomainTrustAnchors} = configuration || {};
             const config = new CoreCrypto.#module.ConversationConfiguration(
                 ciphersuite,
                 externalSenders,
                 custom?.keyRotationSpan,
+                custom?.wirePolicy,
+                perDomainTrustAnchors as any[],
```
augustocdias, line 875: @OtaK  is this the correct way of passing the parameter?

OtaK: Yes but keep in mind that you have set perDomainTrustAnchors as optional on the TS side and it is handled as non-optional on the rust side so you need to check for NULL/undefined on the rust side to make sure you don't get a crash when the user provides nothing. Alternatively, you acn make `perDomainTrustAnchors` have a default value on the TS side. Like so: `perDomainTrustAnchors = [];` on line 869

Cargo.toml
OtaK, line 1: Please do not delete these tags as we use them to pin dependencies when marking releases. Keeping them commented is intended

crypto-ffi/src/wasm.rs
```diff
@@ -749,14 +768,28 @@ impl ConversationConfiguration {
...
             .unwrap_or_default();
-        Self {
+        Ok(Self {
             ciphersuite,
             external_senders,
             custom: CustomConfiguration::new(key_rotation_span, wire_policy),
+            per_domain_trust_anchors: per_domain_trust_anchors
+                .into_iter()
+                .map(|js_anchor| Ok(serde_wasm_bindgen::from_value(js_anchor)?))
+                .collect::<WasmCryptoResult<Vec<_>>>()?,
```
OtaK, line 783: As mentioned in the previous comment (on the question you asked) you need to handle the JS `null` here as the value is optional on the WASM side. Even if the wrapper makes sure the value is provided, we'd be safer if it was handled in any case (and as this is a cold path we can pay the price for safety).

```suggestion
            per_domain_trust_anchors: if per_domain_trust_anchors.is_null() {
                vec![]
            else {
                per_domain_trust_anchors
                    .into_iter()
                    .map(|js_anchor| Ok(serde_wasm_bindgen::from_value(js_anchor)?))
                    .collect::<WasmCryptoResult<Vec<_>>>()?
            },
```

crypto-ffi/src/wasm.rs
```diff
@@ -1577,6 +1612,43 @@ impl CoreCrypto {
...
+        &self,
+        conversation_id: Box<[u8]>,
+        remove_domain_names: Box<[js_sys::JsString]>,
+        add_trust_anchors: js_sys::Array,
+    ) -> Promise {
+        let this = self.inner.clone();
+        future_to_promise(
+            async move {
+                let add_trust_anchors = add_trust_anchors
+                    .into_iter()
```
OtaK, line 1628: Same here, `null` needs to be handled.

crypto/src/mls/credential/trust_anchor.rs
```diff
@@ -0,0 +1,1647 @@
...
+            .into_iter()
+            .map(|cert| cert.to_der())
+            .collect::<Result<_, x509_cert::der::Error>>()?;
+        Ok(encoded_chain)
+    }
+}
+
+/// Checks the root cert against the trust store. In wasm maybe use webpki-roots (https://github.com/rustls/webpki-roots) crate
+fn check_root_in_trust_store(_root: &Certificate) -> CryptoResult<()> {
+    // TODO: verify root certificate is valid in the device's trust store
```
OtaK, line 132: Is this verification supposed to be implemented or is this out of scope of this PR?

OtaK (changes requested):

crypto/src/mls/credential/trust_anchor.rs
augustocdias: It is out of the scope of this PR.

crypto/src/mls/credential/trust_anchor.rs
```diff
@@ -0,0 +1,1647 @@
...
+            domain_name: String::from_utf8(value.domain_name().to_vec())?,
+            intermediate_certificate_chain: pem::encode_many(&pems),
+        })
+    }
+}
+
+impl PerDomainTrustAnchor {
+    /// Converts to the OpenMls's counterpart of this struct
+    /// It performs validation first
+    pub fn into_mls(
```
beltram, line 50: rename into `try_as_checked_openmls_trust_anchor`.

crypto/src/mls/credential/trust_anchor.rs
```diff
@@ -0,0 +1,1647 @@
...
+            .map(|cert| pem::Pem::new("CERTIFICATE", cert.clone()))
+            .collect();
+        Ok(Self {
+            domain_name: String::from_utf8(value.domain_name().to_vec())?,
+            intermediate_certificate_chain: pem::encode_many(&pems),
+        })
+    }
+}
+
+impl PerDomainTrustAnchor {
```
beltram, line 47: Why not merging the 2 impl blocks ? (with the other one on L65)

crypto/src/mls/credential/trust_anchor.rs
```diff
@@ -0,0 +1,1647 @@
...
+                    anchors
+                        .iter()
+                        .any(|anchor| String::from_utf8_lossy(anchor.domain_name()) == self.domain_name)
+                })
+            {
+                return Err(CryptoError::DuplicateDomainName);
+            }
+        }
+
+        // self signed or incomplete chains are not allowed
```
beltram, line 93: This comment is misleading

```suggestion
        // empty chains are not allowed
```

crypto/src/mls/credential/trust_anchor.rs
```diff
@@ -0,0 +1,1647 @@
...
+            .unwrap()
+            .map_err(MlsError::from)?;
+        // ensure that the root is also valid
+        root_cert.is_valid().map_err(MlsError::from)?;
+
+        check_root_in_trust_store(root_cert)?;
+
+        let encoded_chain = certificate_chain
+            .into_iter()
+            .map(|cert| cert.to_der())
```
beltram, line 124: You could optimize this because encoding a certificate to DER is likely to be more expensive than converting PEM into DER (only encoding differs).

Have this at the beginning of your method:

```rust
let encoded_chain = pem::parse_many(&self.intermediate_certificate_chain)?
            .iter()
            .map(|p| p.contents())
            .collect::<Vec<_>>();
```

And then reuse it to build your PkiPath

crypto/src/mls/credential/trust_anchor.rs
```diff
@@ -0,0 +1,1647 @@
...
+                            .is_some_and(|old_anchors| {
+                                !old_anchors
+                                    .iter()
+                                    .any(|old_anchor| old_anchor.certificate_chain() == anchor.certificate_chain())
+                            })
+                })
+                .try_for_each(|anchor| -> CryptoResult<()> {
+                    let anchor = PerDomainTrustAnchor::try_from(anchor)?;
+                    // the domain will obviously be in the context and will be already applied by
+                    // other client, so the domain uniquennes should not be validated here
```
beltram, line 161:
```suggestion
                    // other client, so the domain uniqueness should not be validated here
```

crypto/src/mls/credential/trust_anchor.rs
```diff
@@ -0,0 +1,1647 @@
...
+                    old_group_context.extensions().per_domain_trust_anchors().is_none()
+                        || old_group_context
+                            .extensions()
+                            .per_domain_trust_anchors()
+                            .is_some_and(|old_anchors| {
+                                !old_anchors
+                                    .iter()
+                                    .any(|old_anchor| old_anchor.certificate_chain() == anchor.certificate_chain())
+                            })
+                })
```
beltram, line 157: This reads better IMO. Also you could extract this into a `get_trust_anchor_diff()` method

```suggestion
if let Some(new_anchors) = commit_group_context.extensions().per_domain_trust_anchors() {
            // find new anchors
            new_anchors
                .iter()
                .filter(|new_anchor| {
                    let old_trust_anchor = old_group_context.extensions().per_domain_trust_anchors();

                    let no_old_trust_anchor = old_trust_anchor.is_none();
                    let empty_old_trust_anchor = old_trust_anchor.map(|a| a.is_empty()).unwrap_or_default();

                    let anchor_is_new = old_trust_anchor.is_some_and(|old_anchor| {
                        !old_anchor
                            .iter()
                            .any(|old_anchor| old_anchor.certificate_chain() == new_anchor.certificate_chain())
                    });

                    no_old_trust_anchor || empty_old_trust_anchor || anchor_is_new
                })
```

crypto/src/mls/credential/trust_anchor.rs
```diff
@@ -0,0 +1,1647 @@
...
+            .map(|anchors| {
+                anchors
+                    .iter()
+                    .map(PerDomainTrustAnchor::try_from)
+                    .collect::<CryptoResult<Vec<_>>>()
+            })
+            .unwrap_or_else(|| Ok(Vec::new()))?;
+
+        // check if all to remove exists
+        if remove_domain_names
```
beltram, line 195: You could simplify this by comparing `anchors` length before & after `retain` on L212

crypto/src/mls/credential/trust_anchor.rs
```diff
@@ -0,0 +1,1647 @@
...
+        }
+
+        // check if any new chain is already in the group's context
+        if anchors.iter().any(|a| {
+            add_trust_anchors
+                .iter()
+                .any(|n| n.intermediate_certificate_chain == a.intermediate_certificate_chain)
+        }) {
+            return Err(CryptoError::DuplicateCertificateChain);
+        }
```
beltram, line 209: You could factorize this check with this one a few lines below:

```rust
        // check for duplicate anchors to be added
        if anchors
            .iter()
            .any(|a| add_trust_anchors.iter().any(|n| a.domain_name == n.domain_name))
        {
            return Err(CryptoError::DuplicateDomainName);
        }
```

That way you only iterate once the lists. Use a `try_for_each()`

crypto/src/mls/credential/trust_anchor.rs
```diff
@@ -0,0 +1,1647 @@
...
+                    anchor.validate(backend, None)?;
+                    Ok(())
+                })?;
+        }
+        Ok(())
+    }
+
+    /// see [MlsCentral::update_trust_anchors_from_conversation]
+    #[cfg_attr(test, crate::durable)]
+    pub(crate) async fn update_trust_anchors(
```
beltram, line 171: You should really extract the validation part of this method into a dedicated method. This way we can reuse it if/when we allow to do the same in a GCE proposal

crypto/src/test_utils/central.rs
```diff
@@ -432,6 +435,20 @@ impl MlsCentral {
     pub async fn count_credentials_in_keystore(&self) -> usize {
         self.mls_backend.key_store().count::<MlsCredential>().await.unwrap()
     }
+
+    pub async fn add_per_domain_trust_anchor_unwchecked(
```
beltram, line 439: typo

crypto/src/test_utils/central.rs
```diff
@@ -445,6 +462,42 @@ impl MlsConversation {
...
+    pub fn per_domain_trust_anchors(&self) -> Vec<PerDomainTrustAnchor> {
+        self.extensions()
+            .per_domain_trust_anchors()
+            .unwrap()
+            .iter()
+            .map(|a| PerDomainTrustAnchor::try_from(a).unwrap())
+            .collect()
+    }
+
+    pub async fn add_per_domain_trust_anchor_unwchecked(
```
beltram, line 479: typo

beltram (changes requested):

crypto/src/mls/credential/trust_anchor.rs
```diff
@@ -0,0 +1,1647 @@
...
+                add_trust_anchors,
+                &self.mls_backend,
+            )
+            .await
+    }
+}
+
+pub(crate) fn extract_domain_name(certificate: &Certificate) -> CryptoResult<String> {
+    for attr in certificate.tbs_certificate.subject.0.iter().flat_map(|n| n.0.iter()) {
+        // according to the RFC implementations must be prepared to receive the domain component
```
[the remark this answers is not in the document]
augustocdias: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4

> In addition, implementations of this specification MUST be prepared
   to receive the domainComponent attribute, as defined in [[RFC4519](https://datatracker.ietf.org/doc/html/rfc4519)].
   The Domain Name System (DNS) provides a hierarchical resource
   labeling system.

OtaK (changes requested):

crypto-ffi/src/CoreCrypto.udl
```diff
@@ -179,6 +185,14 @@ enum CryptoError {
     "ClearingPendingCommitError",
     "SelfCommitIgnored",
     "UnmergedPendingGroup",
+   "X509CertDerError",
```
OtaK, line 188: Nit: indenting is wrong here (3 spaces, should be 4)

OtaK (approved):
Small nit to fix but LGTM

PR #349, opened 2023-07-10
feat!: catch duplicate messages [WPB-2788]

beltram:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

To prevent clients from failing with decryption errors whenever a message is replayed by the server.

**THIS DOES NOT DETECT DUPLICATES ACROSS EPOCHS ! It just does so for incoming messages with epoch == current_epoch - 1**

**This also does not work when handshake messages are encrypted (which we do not do ATM and probably never will)**

How it works:
* commits: compute the local group `ConfirmationTag` and compare it with the commit's one
* proposals: compare the proposal with the ones in store
* application message: just change the openmls error mapping 

----
##### References

### DIFF
--- a/crypto-ffi/src/CoreCrypto.udl
+++ b/crypto-ffi/src/CoreCrypto.udl
@@ -94,7 +94,7 @@ enum CryptoError {
     "UnauthorizedExternalAddProposal",
     "UnauthorizedExternalCommit",
     "InvalidHashReference",
-    "GenerationOutOfBound",
+    "DuplicateMessage",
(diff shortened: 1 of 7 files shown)


OtaK (changes requested):
Please mark the PR as breaking; The removal of the `GenerationOutOfBound` error marks a breaking change API-wise on the FFI.

Other than that, LGTM

OtaK (approved):

PR #352, opened 2023-07-11
feat!: prevent conversation overwrite [WPB-2977]

beltram:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Whenever a group is joined locally either by creating it, via a Welcome message or via an external commit ; we are now verifying beforehand if it does not overwrite an existing conversation in case the library is misused.

----
##### References


crypto/src/error.rs
```diff
@@ -22,6 +22,9 @@ pub enum CryptoError {
     /// This error is emitted when the requested conversation couldn't be found in our store
     #[error("Couldn't find conversation")]
     ConversationNotFound(crate::prelude::ConversationId),
+    /// This error is emitted when the requested conversation already exists with the given if
+    #[error("Conversation already exists")]
+    ConversationAlreadyExists(crate::prelude::ConversationId),
```
OtaK, line 27: Maybe we should output the conversation id in the error name; What do you think? Maybe we should bring it up to @rohan-wire, as there's a concern around metadata leakage in possibly logs.

Same maybe with the `ClientNotFound` error, where the error stores the ClientId but doesn't output it in its `Display` impl

OtaK (changes requested):
Unless I missed something, I don't see the safeguard being put in place in the `join_by_external_commit` method.

beltram:
> Unless I missed something, I don't see the safeguard being put in place in the join_by_external_commit method.

That's because you can use external commits to rejoin a conversation, which can still exist e.g. if you were absent for >28 days. So checking this here did not make sense everything considered

crypto/src/error.rs
beltram: When it comes to `ConversationId` & `ClientId` they are both present as path parameters in the backend API. So they are already public. I think it's fine

OtaK (approved):

PR #358, opened 2023-07-20
feat: buffer pending messages [WPB-2851]

beltram:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Buffer pending messages during join by external commit process to tolerate unordered messages.

----
##### References


OtaK (changes requested):
Please test the newly added keystore entity in `keystore/tests/z_entities.rs`. Otherwise we don't know if the backing storage will actually work.

crypto/src/mls/buffer_external_commit.rs
```diff
@@ -0,0 +1,191 @@
...
+                None
+            };
+            let callbacks = self.callbacks.as_ref().map(|boxed| boxed.as_ref());
+            let decrypted = conversation
+                .decrypt_message(m, parent_conversation, self.mls_client()?, &self.mls_backend, callbacks)
+                .await?;
+            decrypted_messages.push(decrypted);
+        }
+
+        let decrypted_messages = (!decrypted_messages.is_empty()).then_some(decrypted_messages);
```
OtaK, line 78: Shouldn't we delete the messages that were restored as well? It would make sense to avoid running into storage issues eventually.

OtaK (changes requested):

crypto/src/mls/buffer_external_commit.rs
beltram: Done

OtaK (approved):

PR #381, opened 2023-08-10
Clean Kotlin IDE completion, tests, topology & publication

beltram:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

TO DESCRIBE

----
##### References


crypto-ffi/bindings/jvm/src/main/kotlin/com/wire/crypto/CoreCrypto.kt
OtaK, line 1: Remove this from repo - this needs to be added to the .gitignore

crypto-ffi/Makefile.toml
```diff
@@ -140,10 +140,49 @@ args = [
...
+args = [
+    "run",
+    "--features", "uniffi/cli",
+    "--bin", "uniffi-bindgen",
+    "generate",
+    "--language", "kotlin",
+    "--out-dir", "./bindings/jvm/src/main/kotlin",
+    "--library", "../target/x86_64-unknown-linux-gnu/release/libcore_crypto_ffi.${LIBRARY_EXTENSION}"
+]
+
```
typfel, line 185: What's the purpose of a separate task for each target? Aren't they all identical?

OtaK (commented):
I have generally no clue about Kotlin so I'm handing it off to @typfel to actually review

The rust/tooling part LGTM so...

typfel (changes requested):

crypto-ffi/Makefile.toml
OtaK: Yeah they are - the library file UniFFI uses is just for the symbols so the platform actually doesn't matter...unless it does
armv7 is 32bits so any `usize` types will have a 32-bit symbol vs 64-bit on other platforms, so that's my only guess.

If it's not a problem I guess we can go back to using a single step to emit the bindings

beltram: @typfel @OtaK the only issue I had with that is that (on my machine) the binary is platform dependent i.e. it is not generated in ~~`/target/release/libcore_crypto_ffi.dylid`~~ but in `/target/aarch64-apple-darwin/release/libcore_crypto_ffi.dylid` so the command `cargo make jvm-aarch64-darwin` always fail.
That's why I added these platform specific tasks

typfel: Had the same problem on a different project. Switched to use the .a files for bindgen. Works on all plattforms.

beltram: I think we have a quiproquo here. The problem is not with the file but where it's generated.

On my machine, when I do `cargo build --release`, the binaries are generated in `target/aarch64-apple-darwin/release`.

[This commit](https://github.com/wireapp/core-crypto/pull/381/commits/e02a3d8256a1cf549c67b3660f94910a06d9559e) should fix that

OtaK: > On my machine, when I do cargo build --release, the binaries are generated in target/aarch64-apple-darwin/release

This looks like a cargo bug. If you're compiling for your own platform it should output in `target/release`
Kinda looks like cargo is running through Rosetta or something on Apple Silicon macOS

beltram: Yes you're probably right. I think it's safe replacing `../target` by `${CARGO_MAKE_CRATE_TARGET_DIRECTORY}` which could let me leverage `$CARGO_TARGET_DIR` instead to overcome this little annoyance

typfel (approved):

PR #383, opened 2023-08-10
feat!: buffer messages from future epochs [WPB-3786]

beltram:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Regular group members were also impacted by unordered messages. As a consequence, messages from future epochs are now buffered.

* `decryptMessage` might throw a `BufferedFutureMessage` error which you should catch & ignore
* `commitAccepted` now returns an optional list of decrypted messages

----
##### References

### DIFF
--- a/crypto-ffi/bindings/js/CoreCrypto.ts
+++ b/crypto-ffi/bindings/js/CoreCrypto.ts
@@ -919,7 +919,12 @@ export class CoreCrypto {
     }
 
     /**
-     * Decrypts a message for a given conversation
+     * Decrypts a message for a given conversation.
+     *
+     * Note: you should catch & ignore the following error reasons:
+     * * "We already decrypted this message once"
+     * * "You tried to join with an external commit but did not merge it yet. We will reapply this message for you when you merge your external commit"
+     * * "Incoming message is for a future epoch. We will buffer it until the commit for that epoch arrives"
      *
      * @param conversationId - The ID of the conversation
      * @param payload - The encrypted message buffer
@@ -1337,6 +1342,7 @@ export class CoreCrypto {
      * and deletes the temporary one. This step makes the group operational and ready to encrypt/decrypt message
      *
      * @param conversationId - The ID of the conversation
+     * @returns the messages from current epoch which had been buffered, if any
      */
     async mergePendingGroupFromExternalCommit(conversationId: ConversationId): Promise<DecryptedMessage[] | undefined> {
         return await CoreCryptoError.asyncMapErr(this.#cc.merge_pending_group_from_external_commit(conversationId));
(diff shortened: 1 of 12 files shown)


crypto-ffi/bindings/js/CoreCrypto.ts
```diff
@@ -1337,6 +1342,7 @@ export class CoreCrypto {
      * and deletes the temporary one. This step makes the group operational and ready to encrypt/decrypt message
      *
      * @param conversationId - The ID of the conversation
+     * @returns eventually the buffered messages from current epoch which had been buffered
```
OtaK, line 1345:
```suggestion
     * @returns the messages from current epoch which had been buffered, if any
```

crypto-ffi/bindings/js/CoreCrypto.ts
```diff
@@ -1358,8 +1364,9 @@ export class CoreCrypto {
      * into the local group state
      *
      * @param conversationId - The group's ID
+     * @returns eventually the buffered messages from current epoch which had been buffered
```
OtaK, line 1367:
```suggestion
     * @returns the messages from current epoch which had been buffered, if any
```

OtaK (changes requested):

OtaK (approved):
LGTM once tests pass

typfel:
@beltram sorry a bit late on this one but why do I get the buffered messages on `commitAccepted`? AFAIK `commitAccepted` is called when I myself have advanced the epoch so how can someone else possibly else send a message in this not yet existing epoch?

And what if I receive message out of order and they are buffered and I never advance the epoch? Then I never receive the buffered messages?

beltram:
> why do I get the buffered messages on commitAccepted? AFAIK commitAccepted is called when I myself have advanced the epoch so how can someone else possibly else send a message in this not yet existing epoch ?

You're right that does not make sense, this should be on `decryptMessage()`, I'll fix that.

> And what if I receive message out of order and they are buffered and I never advance the epoch? Then I never receive the buffered messages ?

Well in that case you are screwed. We cannot do anything about it

typfel:
> > why do I get the buffered messages on commitAccepted? AFAIK commitAccepted is called when I myself have advanced the epoch so how can someone else possibly else send a message in this not yet existing epoch ?
> 
> You're right that does not make sense, this should be on `decryptMessage()`, I'll fix that.
> 
> > And what if I receive message out of order and they are buffered and I never advance the epoch? Then I never receive the buffered messages ?
> 
> Well in that case you are screwed. We cannot do anything about it

I was thinking I when receive an incoming commit I would get the buffered messages for that commit.

PR #450, opened 2023-12-13
feat(wip): E2EI PKI environment support

OtaK:


beltram (changes requested):

crypto/src/e2e_identity/mod.rs
```diff
@@ -83,6 +100,184 @@ impl MlsCentral {
...
+        let inter_ca = PkiEnvironment::decode_pem_cert(cert_pem).map_err(|e| CryptoError::E2eiError(e.into()))?;
+        let (ski, aki) =
+            PkiEnvironment::extract_ski_aki_from_cert(&inter_ca).map_err(|e| CryptoError::E2eiError(e.into()))?;
+
+        let ski_aki_pair = format!("{ski}:{}", aki.unwrap_or_default());
+
+        // Validate it
+        let Some(pki_env) = self.e2ei_pki_env.as_ref() else {
+            unreachable!("Pki Env mysteriously disappeared despite being checked beforehand");
+        };
```
beltram, line 215: Move this to the beginning of the method instead of checking it first then panicking when it's not found

crypto/src/mls/external_commit.rs
```diff
@@ -262,6 +269,22 @@ impl MlsConversation {
...
+                .filter_map(|add_proposal| {
+                    let credential = add_proposal.add_proposal().key_package().leaf_node().credential();
+
+                    matches!(credential.credential_type(), CredentialType::X509).then(|| credential.clone())
+                })
+                .collect();
+            let state = compute_state(credentials.iter(), Some(pki_env), MlsCredentialType::X509);
+            if state != E2eiConversationState::Verified {
+                // FIXME: Uncomment when PKI env can be seeded - the computation is still done to assess performance and impact of the validations
+                // return Err(CryptoError::InvalidCertificateChain);
```
beltram, line 284: I guess this can be uncommented now ?

keystore/src/connection/platform/wasm/mod.rs
```diff
@@ -120,6 +120,17 @@ impl DatabaseConnection for WasmConnection {
                     .auto_increment(false)
                     .add_index(Index::new("id", "id").unique(true)),
             )
+            .add_object_store(
+                ObjectStore::new("e2ei_acme_ca")
+                    .auto_increment(false)
+                    .add_index(Index::new("id", "id").unique(true)),
+            )
+            .add_object_store(ObjectStore::new("e2ei_intermediate_certs").auto_increment(false))
```
beltram, line 128: Why not indexing `ski_aki_pair` ?

beltram (approved):

crypto/src/mls/external_commit.rs
OtaK: > I guess this can be uncommented now ?

This will make *all* tests fail. We need to uncomment it for sure only when we have a test environment for this.

crypto/src/e2e_identity/mod.rs
OtaK: Good catch

PR #481, opened 2024-02-02
feat!: `clientPublicKey` now also works for x509 credentials

beltram:


OtaK (approved):
To re rebased, but LGTM

PR #487, opened 2024-02-13
fix: Allow revoked Credentials in MLS operations

OtaK:


mls-provider/src/pki.rs
```diff
@@ -123,7 +114,8 @@ impl openmls_traits::authentication_service::AuthenticationServiceDelegate for P
                             PathValidationStatus::CertificateRevoked
                             | PathValidationStatus::CertificateRevokedEndEntity
                             | PathValidationStatus::CertificateRevokedIntermediateCa => {
-                                return CredentialAuthenticationStatus::Revoked;
+                                // ? Revoked credentials are A-OK. They still degrade conversations though.
+                                // return CredentialAuthenticationStatus::Revoked;
                             }
                             PathValidationStatus::InvalidNotAfterDate => {
                                 return CredentialAuthenticationStatus::Expired;
```
beltram, line 121: Same for expired certificates

beltram (changes requested):

mls-provider/src/pki.rs
OtaK: Expired certificates are allowed in any case on the MLS side [as recommended by the relevant section of RFC9420](https://www.rfc-editor.org/rfc/rfc9420.html#section-5.3.2-4) - it is just an indicative status https://github.com/wireapp/openmls/blob/9b724b76e0035a464216ac0dfcab412039013257/openmls/src/credentials/mod.rs#L315-L322

beltram (approved):

PR #563, opened 2024-05-31
refactor: crypto-ffi: statically link the runtime lib via Makefile.toml [WPB-8945]

istankovic:
On Android x86-64 we need to link the runtime lib statically because       
the compiler fails to do so due to -nodefaultlibs being given by           
default. This is not a problem on other Android platforms, though.         
                                                                           
More details at https://github.com/mozilla/application-services/issues/5436

### DIFF
--- a/crypto-ffi/Makefile.toml
+++ b/crypto-ffi/Makefile.toml
@@ -263,7 +263,9 @@ args = [
     "--crate-type=staticlib",
     "--release",
     "--",
-    "-C", "strip=symbols"
+    "-C", "strip=symbols",
+    "-l", "static=clang_rt.builtins-x86_64-android",
+    "-L", "${CLANG_RT_DIR}"
 ]
 dependencies = ["ffi-kotlin", "android-env"]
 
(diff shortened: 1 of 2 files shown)


vitorhugods:
Please elaborate.

This was added as android x86 was crashing due to missing clang libs during runtime.

See:
- #543

istankovic:
> Please elaborate.
> 
> This was added as android x86 was crashing due to missing clang libs during runtime.
> 
> See:
> 
>     * [fix: Use Mozilla's hack to fix Android on x86_64 [WPB-8945] #543](https://github.com/wireapp/core-crypto/pull/543)

@vitorhugods could you please try the updated branch here?
I've moved the linking logic to `Makefile.toml`, the result should be the same.

istankovic:
@typfel said it worked for him.

vitorhugods:
> @typfel said it worked for him.

I trust him more than I trust myself.

augustocdias (approved):

PR #581, opened 2024-06-13
test: pin future to heap in test with overflowing stack [WPB-9543]

SimonThormeyer:
# What's new in this PR
To fix the failing tests in CI, we pin the future to heap in test with overflowing stack

----

- The **PR Title**

### DIFF
--- a/crypto/src/mls/credential/mod.rs
+++ b/crypto/src/mls/credential/mod.rs
@@ -331,7 +331,10 @@ pub mod tests {
     #[apply(all_cred_cipher)]
     #[wasm_bindgen_test]
     async fn should_not_fail_but_degrade_when_basic_joins(case: TestCase) {
-        if case.is_x509() {
+        if !case.is_x509() {
+            return;
+        }
+        Box::pin(async {
             let mut x509_test_chain = X509TestChain::init_empty(case.signature_scheme());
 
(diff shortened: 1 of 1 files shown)


crypto/src/mls/credential/mod.rs
```diff
@@ -332,86 +332,89 @@ pub mod tests {
...
-            assert_eq!(
-                alice_central.e2ei_conversation_state(&id).await.unwrap(),
-                E2eiConversationState::Verified
-            );
-
-            assert_eq!(
-                bob_central.e2ei_conversation_state(&id).await.unwrap(),
-                E2eiConversationState::Verified
-            );
+            Box::pin(async {
```
istankovic, line 335: Nitpick: we could check if the case is not using x509 credentials and then do an early return, which would save us one level of indentation.

Alternatively, we could create a new test fixture, similar to `all_cred_cipher`, e.g. `all_cert_cipher`, which would use only x509 credentials. See [this](https://github.com/wireapp/core-crypto/blob/31e1d63d9ea412624ca469a9c4f75fb6270029a2/crypto/src/test_utils/fixtures.rs#L97). Such a test fixture would allow us to get rid of the above checks in a bunch of other places, too.

istankovic (approved):

augustocdias (approved):

PR #582, opened 2024-06-14
style: fix linting issues

SimonThormeyer:
# What's new in this PR
Some CI actions were failing due to linting errors or missing formatting. This PR fixes those issues.

----

- The **PR Title**

### DIFF
--- a/crypto/src/mls/buffer_external_commit.rs
+++ b/crypto/src/mls/buffer_external_commit.rs
@@ -30,7 +30,7 @@ impl MlsCentral {
 }
 
 #[cfg(test)]
-pub mod tests {
+mod tests {
     use crate::{test_utils::*, CryptoError};
     use wasm_bindgen_test::*;
 
@@ -38,7 +38,7 @@ pub mod tests {
 
     #[apply(all_cred_cipher)]
     #[wasm_bindgen_test]
-    pub async fn should_buffer_and_reapply_messages_after_external_commit_merged(case: TestCase) {
+    async fn should_buffer_and_reapply_messages_after_external_commit_merged(case: TestCase) {
         run_test_with_client_ids(
             case.clone(),
             ["alice", "bob", "charlie", "debbie"],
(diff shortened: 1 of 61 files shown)


crypto/src/mls/buffer_external_commit.rs
```diff
@@ -30,7 +30,7 @@ impl MlsCentral {
 }
 
 #[cfg(test)]
-pub mod tests {
```
istankovic, line 33: I think we should revisit all places like this that shouldn't be `pub` and just make them private.

augustocdias: although they don't need to be public, it doesn't really matter since this is only for test targets

SimonThormeyer: It's a style issue I guess... I changed it to private

crypto/src/mls/client/identities.rs
```diff
@@ -206,7 +206,7 @@ pub mod tests {
         }
     }
 
-    pub mod push {
+    pub(crate) mod push {
```
istankovic, line 209: Why do we need this if the parent mod is not pub? Similarly for `find` above.

SimonThormeyer: Suprised that cargo check didn't catch this. Updating this in a minute.

crypto/src/mls/conversation/commit.rs
```diff
@@ -374,7 +374,7 @@ pub mod tests {
 
     wasm_bindgen_test_configure!(run_in_browser);
 
-    pub mod add_members {
+    pub(crate) mod add_members {
```
istankovic, line 377: This should also be private. Likewise for the rest of the modules in `tests`.

interop/src/build/web/wasm.rs
```diff
@@ -19,7 +19,7 @@ use crate::TEST_SERVER_PORT;
 use color_eyre::eyre::Result;
 use std::net::SocketAddr;
 
-pub async fn find_wasm_file(deploy_path: &std::path::Path) -> Result<std::path::PathBuf> {
+pub(crate) async fn find_wasm_file(deploy_path: &std::path::Path) -> Result<std::path::PathBuf> {
```
istankovic, line 22: Why does this need to be `pub(crate)`? I don't see it used outside of this module.

istankovic (approved):

PR #593, opened 2024-06-19
 ci: bindings.yml: use 'build' rather than 'check' naming 

istankovic:


.github/workflows/bindings.yml
```diff
@@ -52,19 +52,28 @@ jobs:
...
-      - uses: actions-rust-lang/setup-rust-toolchain@v1  # this implicitly caches Rust tools and build artifacts
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           rustflags: ''
-          target: "aarch64-apple-ios,x86_64-apple-ios,aarch64-apple-ios-sim"
+          target: ${{ matrix.task.target }}
       - name: Setup cargo-make
         uses: davidB/rust-cargo-make@v1
       - run: |
           cd crypto-ffi
```
SimonThormeyer, line 75: Maybe we can add a `- name` here too? I find it confusing that in the overview this is called `cd crypto-ffi`

istankovic: Totally. :+1:  :-)

.github/workflows/bindings.yml
```diff
@@ -52,19 +52,28 @@ jobs:
...
-          cargo make ios-create-xcframework
```
SimonThormeyer, line 67: Do we want a downstream job that does `cargo make ios-create-xcframework` and
```
- needs: [ios-device, ios-simulator-x86, ios-simulator-arm]
```
?

istankovic: Yeah, there's going to be another job that collects all of the artifacts. Note this is still a draft. :wink:

.github/workflows/bindings.yml
istankovic: Done.

.github/workflows/bindings.yml
```diff
@@ -62,11 +62,12 @@ jobs:
...
       - name: Setup cargo-make
         uses: davidB/rust-cargo-make@v1
-      - run: |
+      - name: Build Swift package
+        run: |
           cd crypto-ffi
           cargo make ios-create-xcframework
 
-  check-wasm:
+  build-wasm:
```
SimonThormeyer, line 70: This could be named `build-and-test-wasm`, note `bun test` below

SimonThormeyer (approved):
Approved, just one minor thing regarding the wasm step

PR #600, opened 2024-07-03
feat: add log level to the callback [WPB-7260]

augustocdias:
# What's new in this PR

This adds the log level to the log callback

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/src/generic.rs
+++ b/crypto-ffi/src/generic.rs
@@ -30,7 +30,7 @@ use core_crypto::{
     },
     MlsError,
 };
-use tracing::level_filters::LevelFilter;
+use tracing::{level_filters::LevelFilter, Level};
 use tracing_subscriber::fmt::{self, MakeWriter};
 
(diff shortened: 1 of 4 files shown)


crypto-ffi/src/generic.rs
```diff
@@ -945,8 +973,8 @@ impl CoreCrypto {
     pub fn set_logger(&self, logger: std::sync::Arc<dyn CoreCryptoLogger>, level: CoreCryptoLogLevel) {
```
typfel, line 973: Is the logger is global and not bound to specific core crypto instance? Because me as a client call this on a core crypto instance but it will I think saw that it throws an error if try to do it again on a second core crypto instance.

augustocdias: yes it is global

typfel: Can't it be a static function then?

typfel (approved):

PR #605, opened 2024-07-10
chore: Update dependencies and fix android build on Mac [WPB-9740]

augustocdias:
# What's new in this PR

Update dependencies to latest versions, remove the pin on the openssl-sys crate to a fixed version and fix task to build android on mac os

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/Makefile.toml
+++ b/crypto-ffi/Makefile.toml
@@ -156,11 +156,21 @@ script = '''
     mv ./bindings/jvm/src/main/kotlin/com/wire/crypto/core_crypto_ffi.kt ./bindings/jvm/src/main/kotlin/com/wire/crypto/CoreCrypto.kt
 
     sed -i.bak \
-        -e 's/\CryptoException\b/CryptoError/g' \
+        -e 's/\bCryptoException\b/CryptoError/g' \
         -e 's/\bE2eIdentityException\b/E2eIdentityError/g' \
         ./bindings/jvm/src/main/kotlin/uniffi/core_crypto/core_crypto.kt
 
     rm ./bindings/jvm/src/main/kotlin/uniffi/core_crypto/core_crypto.kt.bak
+'''
+
+[tasks.ffi-kotlin.mac]
+script = '''
+    mv ./bindings/jvm/src/main/kotlin/com/wire/crypto/core_crypto_ffi.kt ./bindings/jvm/src/main/kotlin/com/wire/crypto/CoreCrypto.kt
+
+    sed -i '' \
+        -e 's/[[:<:]]CryptoException[[:>:]]/CryptoError/g' \
+        -e 's/[[:<:]]E2eIdentityException[[:>:]]/E2eIdentityError/g' \
+        ./bindings/jvm/src/main/kotlin/uniffi/core_crypto/core_crypto.kt
 '''
 
 [tasks.ffi]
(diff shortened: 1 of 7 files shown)


crypto-ffi/Makefile.toml
```diff
@@ -156,13 +156,23 @@ script = '''
...
+[tasks.ffi-kotlin.mac]
+script = '''
+    mv ./bindings/jvm/src/main/kotlin/com/wire/crypto/core_crypto_ffi.kt ./bindings/jvm/src/main/kotlin/com/wire/crypto/CoreCrypto.kt
+
+    sed -i '' \
+        -e 's/[[:<:]]CryptoException[[:>:]]/CryptoError/g' \
+        -e 's/[[:<:]]E2eIdentityException[[:>:]]/E2eIdentityError/g' \
+        ./bindings/jvm/src/main/kotlin/uniffi/core_crypto/core_crypto.kt
+'''
+
```
typfel, line 175: So we fixing UniFFI not following Swift conventions when generating errors types. Should we also open issue at the UniFFI project?

augustocdias: What you mean? Nothing is changing here in behaviors. This is there because MacOS's sed (which is bsd) has a different behavior than Linux's (which is gnu).

typfel: Ok, then I misunderstood the change.

keystore/Cargo.toml
```diff
@@ -57,10 +57,9 @@ git = "https://github.com/wireapp/proteus"
 branch = "2.x"
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
-async-fs = "2.0"
-blocking = "1.2"
-# Fixes the dynamic linking issue on Android with the atomic lib
-openssl-sys = "=0.9.92"
+async-fs = "2.1"
+blocking = "1.6"
+openssl-sys = "0.9"
```
istankovic, line 62: Is there a reason why we specify `openssl-sys` here explicitly? I don't think we are using it directly, but via one of our other dependencies so it  would be best to let those dependencies pick the version for us (and hopefully make dep resolver's work easier).

crypto-ffi/Makefile.toml
```diff
@@ -156,13 +156,23 @@ script = '''
...
 
+[tasks.ffi-kotlin.mac]
+script = '''
+    mv ./bindings/jvm/src/main/kotlin/com/wire/crypto/core_crypto_ffi.kt ./bindings/jvm/src/main/kotlin/com/wire/crypto/CoreCrypto.kt
+
+    sed -i '' \
+        -e 's/[[:<:]]CryptoException[[:>:]]/CryptoError/g' \
+        -e 's/[[:<:]]E2eIdentityException[[:>:]]/E2eIdentityError/g' \
+        ./bindings/jvm/src/main/kotlin/uniffi/core_crypto/core_crypto.kt
+'''
```
istankovic, line 174: If the BSD sed supports `\<` and `\>` for word boundaries, you could modify the original sed invocation to be e.g.
```
       sed -i.bak \
        -e 's/\<CryptoException\>/CryptoError/g' \
        -e 's/\<E2eIdentityException\>/E2eIdentityError/g' \
        ./bindings/jvm/src/main/kotlin/uniffi/core_crypto/core_crypto.kt
```
The `\<` and `\>` are supported by GNU sed so we might be able to use the same command on both platforms.

istankovic (approved):
Looks good! Thanks for doing this!

PR #615, opened 2024-07-22
fix: handle own commit after mls error [WPB-10105]

SimonThormeyer:
# What's new in this PR
See corresponing jira item

----

- The **PR Title**

### DIFF
--- a/crypto/src/mls/conversation/decrypt.rs
+++ b/crypto/src/mls/conversation/decrypt.rs
@@ -8,6 +8,7 @@
 //! | 0 pend. Proposal  | ✅              | ✅              |
 //! | 1+ pend. Proposal | ✅              | ✅              |
 
+use openmls::prelude::StageCommitError;
 use openmls::{
     framing::errors::{MessageDecryptionError, SecretTreeError},
     group::StagedCommit,
@@ -123,13 +124,17 @@ impl MlsConversation {
         callbacks: Option<&dyn CoreCryptoCallbacks>,
         restore_pending: bool,
     ) -> CryptoResult<MlsConversationDecryptMessage> {
-        // handles the crooked case where we receive our own commits.
-        // Since this would result in an error in openmls, we handle it here
-        if let Some(ct) = self.maybe_self_member_commit(&message)? {
-            return self.handle_self_member_commit(backend, ct).await;
-        }
-
-        let message = self.parse_message(backend, message).await?;
+        let message = match self.parse_message(backend, message.clone()).await {
+            // Handles the case where we receive our own commits.
+            Err(CryptoError::MlsError(MlsError::MlsMessageError(ProcessMessageError::InvalidCommit(
+                StageCommitError::OwnCommit,
+            )))) => {
+                let ct = self.extract_confirmation_tag_from_own_commit(&message)?;
+                return self.handle_own_commit(backend, ct).await;
+            }
+            Ok(processed_message) => Ok(processed_message),
+            Err(e) => Err(e),
+        }?;
 
         let credential = message.credential();
 
--- a/crypto/src/mls/conversation/own_commit.rs
+++ b/crypto/src/mls/conversation/own_commit.rs
@@ -11,35 +11,46 @@ use openmls::prelude::{
 };
 
 impl MlsConversation {
-    pub(crate) fn maybe_self_member_commit<'a>(
+    /// Returns the confirmation tag from a public message that is an own commit.
+    /// Returns an error if the confirmation tag in the own commit is missing.
+    pub(crate) fn extract_confirmation_tag_from_own_commit<'a>(
         &self,
-        msg: &'a MlsMessageIn,
-    ) -> CryptoResult<Option<&'a ConfirmationTag>> {
-        match msg.body_as_ref() {
+        own_commit: &'a MlsMessageIn,
+    ) -> CryptoResult<&'a ConfirmationTag> {
+        match own_commit.body_as_ref() {
             MlsMessageInBody::PublicMessage(msg) => {
                 let is_commit = matches!(msg.content_type(), ContentType::Commit);
                 let own_index = self.group.own_leaf_index();
                 let is_self_sent = matches!(msg.sender(), Sender::Member(i) if i == &own_index);
-                let is_self_commit = is_commit && is_self_sent;
+                let is_own_commit = is_commit && is_self_sent;
 
-                match is_self_commit.then_some(msg.body()) {
+                match is_own_commit.then_some(msg.body()) {
                     Some(FramedContentBodyIn::Commit(_)) => {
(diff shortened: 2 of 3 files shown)


crypto/src/mls/conversation/decrypt.rs
```diff
@@ -123,13 +124,19 @@ impl MlsConversation {
...
-        let message = self.parse_message(backend, message).await?;
+        let message = match self.parse_message(backend, message.clone()).await {
+            // Handles the case where we receive our own commits.
+            Err(CryptoError::MlsError(MlsError::MlsMessageError(ProcessMessageError::InvalidCommit(
+                StageCommitError::OwnCommit,
+            )))) => {
+                if let Some(ct) = self.maybe_self_member_commit(&message)? {
+                    return self.handle_self_member_commit(backend, ct).await;
+                }
+                Err(CryptoError::InternalMlsError)
```
SimonThormeyer, line 135: This is not expected to be ever reached. Should we `panic!()` instead?

istankovic: If this is indeed unreachable, there's `unreachable!()`.

istankovic:
Could you please split the test into its own commit, on top of the fix commit?

crypto/src/mls/conversation/decrypt.rs
```diff
@@ -123,13 +124,17 @@ impl MlsConversation {
         callbacks: Option<&dyn CoreCryptoCallbacks>,
         restore_pending: bool,
     ) -> CryptoResult<MlsConversationDecryptMessage> {
-        // handles the crooked case where we receive our own commits.
-        // Since this would result in an error in openmls, we handle it here
```
istankovic, line 127: Is this (old) comment still relevant? Which error do we get from openmls if we let own commits through?

SimonThormeyer: The error that we expect and handle below

istankovic: Ah, I assume this is exactly the error you're matching on below, right?

crypto/src/mls/conversation/own_commit.rs
```diff
@@ -11,35 +11,46 @@ use openmls::prelude::{
...
 
 impl MlsConversation {
-    pub(crate) fn maybe_self_member_commit<'a>(
+    /// Extract the confirmation tag from a public message that is an own commit.
+    ///
+    /// # Arguments
+    ///
+    /// * `own_commit`: The own commit.
+    ///
+    /// returns: Result<&ConfirmationTag, CryptoError>: The confirmation tag from the own commit or an error if the message is not an own commit.
```
istankovic, line 20: I think this last line and the whole `Arguments` area is unnecessary.
The former because we already have the function signature and we already describe what the function does at the top. The latter because it doesn't add any additional information on top of the already well-named argument to the function.

So I think just keeping the first line (14) should be sufficient in this case.
In general, I find the Rust's stdlib a good model for documentation, see e.g.
https://doc.rust-lang.org/std/string/struct.String.html#method.find

crypto/src/mls/conversation/own_commit.rs
```diff
@@ -11,35 +11,46 @@ use openmls::prelude::{
...
                             .confirmation_tag
                             .as_ref()
                             .ok_or(CryptoError::InternalMlsError)?;
-                        Ok(Some(ct))
+                        Ok(confirmation_tag)
                     }
-                    _ => Ok(None),
+                    // Not a self commit. Should never be reached if this function
+                    // is called correctly.
+                    _ => Err(CryptoError::ImplementationError),
```
istankovic, line 44: Why not use `unreachable!()`? Similarly for the other instance below.

SimonThormeyer: We could; however, I thought of a scenario where CC has a bug and this is reached. We we use `unreachable!()`, clients can not recover. With this, they can. And the desciption of the error `"We have done something terribly wrong and it needs to be fixed"` fits well in that case, so I thought it is the right thing to do.

istankovic: That's actually my problem with it. Are we sure that error will eventually get back to us?
If it's something recoverable and the client just proceeds, in most cases we will never get a report, meaning the bug would remain.

SimonThormeyer: OK. You convinced me, I'll change it to `unreachable!()`.

crypto/src/mls/conversation/own_commit.rs
```diff
@@ -281,4 +284,89 @@ mod tests {
...
+                        .pending_commit(&conversation_id)
+                        .await
+                        .is_some());
+
+                    let commit_serialized = &mut add_bob_message.commit.to_bytes().unwrap();
+
+                    // Tamper with the commit; this is the signature region, however,
+                    // the membership tag covers the signature, so this will result in an
+                    // invalid membership tag error emitted by openmls.
+                    let _ = std::mem::replace(&mut commit_serialized[355], 42);
```
istankovic, line 334: We don't really need `mem::replace` here, this would suffice:
```diff
-                    let _ = std::mem::replace(&mut commit_serialized[355], 42);               
+                    commit_serialized[355] = commit_serialized[355].wrapping_add(1);          
```

With `mem::replace` and `42` we also run the risk of a flaky test if the generated happens to be `42`.
By using `wrapping_add` or a similar operation we are guaranteed to change the tag.

crypto/src/mls/conversation/own_commit.rs
```diff
@@ -281,4 +284,89 @@ mod tests {
...
+                        .await
+                        .is_ok());
+
+                    // ...and has cleared the pending commit.
+                    assert!(alice_central
+                        .mls_central
+                        .pending_commit(&conversation_id)
+                        .await
+                        .is_none());
+                })
```
istankovic, line 367: Great test, thanks for writing it!

istankovic (approved):
Looks good, with a few comments.

crypto/src/mls/conversation/own_commit.rs
```diff
@@ -11,35 +11,46 @@ use openmls::prelude::{
...
                             .ok_or(CryptoError::InternalMlsError)?;
-                        Ok(Some(ct))
+                        Ok(confirmation_tag)
                     }
-                    _ => Ok(None),
+                    // Not an own commit. Should never be reached if this function
+                    // is called correctly.
+                    _ => unreachable!(
+                        "extract_confirmation_tag_from_own_commit() must always be called \
+                        with a an own commit."
```
istankovic, line 40: Nitpick: typo "a an" -> "an"

SimonThormeyer: Saw that a second too late

PR #618, opened 2024-07-24
tests: add cross signing tests [WPB-7264]

augustocdias:
# What's new in this PR

This adds 2 new tests for cross signing certificates. 

----

- The **PR Title**

### DIFF
--- a/crypto/src/e2e_identity/identity.rs
+++ b/crypto/src/e2e_identity/identity.rs
@@ -156,6 +156,7 @@ impl MlsConversation {
             return Err(CryptoError::ConsumerError);
         }
         let user_ids = user_ids.iter().map(|uid| uid.as_bytes()).collect::<Vec<_>>();
+
         self.members_with_key()
             .iter()
             .filter_map(|(id, c)| UserId::try_from(id.as_slice()).ok().zip(Some(c)))
@@ -177,7 +178,10 @@ impl MlsConversation {
 mod tests {
     use wasm_bindgen_test::*;
 
-    use crate::prelude::MlsCredentialType;
+    use crate::{
+        mls::MlsCentral,
+        prelude::{ClientId, ConversationId, MlsCredentialType},
+    };
     use crate::{
         prelude::{DeviceStatus, E2eiConversationState},
         test_utils::*,
@@ -185,6 +189,55 @@ mod tests {
     };
 
     wasm_bindgen_test_configure!(run_in_browser);
+
+    async fn all_identities_check<const N: usize>(
+        central: &mut MlsCentral,
+        id: &ConversationId,
+        user_ids: &[String; N],
+        expected_sizes: [usize; N],
+    ) {
+        let all_identities = central.get_user_identities(id, user_ids).await.unwrap();
+        assert_eq!(all_identities.len(), N);
+        for (expected_size, user_id) in expected_sizes.into_iter().zip(user_ids.iter()) {
+            let alice_identities = all_identities.get(user_id).unwrap();
+            assert_eq!(alice_identities.len(), expected_size);
+        }
+        // Not found
+        let not_found = central
+            .get_user_identities(id, &["aaaaaaaaaaaaa".to_string()])
+            .await
+            .unwrap();
+        assert!(not_found.is_empty());
+
+        // Invalid usage
+        let invalid = central.get_user_identities(id, &[]).await;
+        assert!(matches!(invalid.unwrap_err(), CryptoError::ConsumerError));
+    }
+
+    async fn check_identities_device_status<const N: usize>(
+        central: &mut MlsCentral,
+        id: &ConversationId,
+        client_ids: &[ClientId; N],
+        name_status: &[(&'static str, DeviceStatus); N],
+    ) {
+        let mut identities = central.get_device_identities(id, client_ids).await.unwrap();
+
+        for j in 0..N {
+            let client_identity = identities.remove(
+                identities
+                    .iter()
+                    .position(|i| i.x509_identity.as_ref().unwrap().display_name == name_status[j].0)
+                    .unwrap(),
+            );
+            assert_eq!(client_identity.status, name_status[j].1);
+        }
+        assert!(identities.is_empty());
+
+        assert_eq!(
+            central.e2ei_conversation_state(id).await.unwrap(),
+            E2eiConversationState::NotVerified
+        );
+    }
 
     #[async_std::test]
     #[wasm_bindgen_test]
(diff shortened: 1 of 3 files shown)


crypto/src/test_utils/mod.rs
```diff
@@ -168,31 +174,153 @@ pub async fn run_test_with_deterministic_client_ids<const N: usize>(
...
+pub async fn run_cross_signed_tests_with_client_ids<const N: usize, const F: usize>(
     case: TestCase,
     client_ids: [[&'static str; 3]; N],
+    federated_client_ids: [[&'static str; 3]; F],
+    (domain1, domain2): (&'static str, &'static str),
+    test: impl FnOnce([MlsCentral; N], [MlsCentral; F]) -> std::pin::Pin<Box<dyn std::future::Future<Output = ()> + 'static>>
+        + 'static,
+) {
+    assert!(case.is_x509(), "This is only supported for x509 test cases");
+    let _ = pretty_env_logger::try_init();
```
istankovic, line 203: Why not just use `pretty_env_logger::init()`, if we're discarding the return value?

crypto/src/test_utils/mod.rs
```diff
@@ -168,31 +174,153 @@ pub async fn run_test_with_deterministic_client_ids<const N: usize>(
...
+    (params1, params2): (CertificateParams, CertificateParams),
+    revoked_display_names: &'static [&'static str],
+) -> (X509TestChain, X509TestChain) {
+    let mut chain1 = init_x509_test_chain(case, &client_ids, revoked_display_names, params1);
+    let mut chain2 = init_x509_test_chain(case, &federated_client_ids, revoked_display_names, params2);
+    chain1.cross_sign(&mut chain2);
+    (chain1, chain2)
+}
+
+pub async fn run_cross_signed_tests_with_client_ids<const N: usize, const F: usize>(
```
istankovic, line 194: We should be consistent with other `run_test_` functions, e.g.
```
run_test_with_client_ids_cross_signed_certs
```
(note also: we're running just 1 test, not multiple tests)

crypto/src/test_utils/mod.rs
```diff
@@ -168,31 +174,153 @@ pub async fn run_test_with_deterministic_client_ids<const N: usize>(
...
+pub fn init_federated_x509_test_chains<const N: usize, const F: usize>(
+    case: &TestCase,
+    client_ids: [[&'static str; 3]; N],
+    federated_client_ids: [[&'static str; 3]; F],
+    (params1, params2): (CertificateParams, CertificateParams),
+    revoked_display_names: &'static [&'static str],
+) -> (X509TestChain, X509TestChain) {
+    let mut chain1 = init_x509_test_chain(case, &client_ids, revoked_display_names, params1);
+    let mut chain2 = init_x509_test_chain(case, &federated_client_ids, revoked_display_names, params2);
+    chain1.cross_sign(&mut chain2);
```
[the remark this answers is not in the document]
augustocdias: It is. If you check the `cross_sign` function it adds the intermediates in both of them

crypto/src/test_utils/mod.rs
augustocdias: because it can panic if another one is set before

istankovic: Hm, can we ensure that `pretty_env_logger` is initialized only once, say, in some toplevel test fixture or something like that? I see it only being used in two places.

crypto/src/test_utils/x509.rs
```diff
@@ -79,15 +80,50 @@ pub struct X509TestChain {
...
+pub struct X509TestChainArgs {
     pub root_params: CertificateParams,
     pub local_ca_params: CertificateParams,
     pub signature_scheme: SignatureScheme,
-    pub federated_test_chains: &'a [X509TestChain],
     pub local_actors: Vec<X509TestChainActorArg>,
     pub dump_pem_certs: bool,
 }
 
+// Helps debugging certificate chaings by printing the PEM into the stdout
```
istankovic, line 91: Typo: `chaings` -> `chains`

crypto/src/test_utils/x509.rs
```diff
@@ -79,15 +80,50 @@ pub struct X509TestChain {
...
+                "{}",
+                &actor
+                    .certificate
+                    .certificate
+                    .to_pem(x509_cert::der::pem::LineEnding::LF)
+                    .unwrap()
+            )
+        })
+    }
+}
```
istankovic, line 125: Very useful to have this! :+1:

crypto/src/test_utils/x509.rs
```diff
@@ -253,41 +287,43 @@ impl X509TestChain {
...
+        other_chain.crls = self.crls.clone();
 
-                intermediates.push(cross_signed_intermediate);
-            }
-            let mut federated_actors = federated_chain.actors.clone();
-            for actor in &mut federated_actors {
-                actor.certificate.is_federated = true;
-            }
+        let mut self_new_intermediates = vec![];
+        for fed_intermediate in &other_chain.intermediates {
```
istankovic, line 305: `fed_intermediate` can be just `intermediate`.

crypto/src/test_utils/x509.rs
```diff
@@ -253,41 +287,43 @@ impl X509TestChain {
...
+            let cross_signed_intermediate = self.trust_anchor.cross_sign_intermediate(fed_intermediate);
+            self_new_intermediates.push(cross_signed_intermediate);
+        }
 
-            actors.extend(federated_actors);
+        let mut other_new_intermediates = vec![];
+        for fed_intermediate in &self.intermediates {
```
istankovic, line 311: Similarly here.

crypto/src/test_utils/mod.rs
istankovic: Ah, good point. :+1:

crypto/src/e2e_identity/identity.rs
```diff
@@ -259,6 +259,114 @@ mod tests {
         .await
     }
 
+    // this test is a duplicate of its counterpart but taking federation into account
+    // The heavy lifting of cross-signing the certificates is being done by the test utils.
+    #[async_std::test]
+    #[wasm_bindgen_test]
+    async fn should_read_revoked_device_cross_signed() {
```
istankovic, line 266: This test is exactly the same as `should_read_revoked_device`, the only thing different looks to be the name. It feels like something is missing.

crypto/src/e2e_identity/identity.rs
```diff
@@ -421,6 +529,119 @@ mod tests {
...
+                    let all_identities = alice_android_central
+                        .mls_central
+                        .get_user_identities(&id, &[alice_user_id.clone(), bob_user_id.clone()])
+                        .await
+                        .unwrap();
+                    assert_eq!(all_identities.len(), 2);
+                    let alice_identities = alice_identities.get(&alice_user_id).unwrap();
+                    assert_eq!(alice_identities.len(), 2);
+                    let bob_identities = bob_identities.get(&bob_user_id).unwrap();
+                    assert_eq!(bob_identities.len(), 1);
```
istankovic, line 626: This looks weird. And I realize this is just copied from the original test, but I think we should nevertheless address it.

Here we just check that the length of `all_identities` is 2.
But then, we repeat the same checks for Bob and Alice that we had above. So I'm not really sure what this is supposed to test. Maybe the original author intended to write something like the following?

```rust
let all_identities = ...
assert_eq!(all_identities.get(&alice_user_id).unwrap().len(), 2);
assert_eq!(all_identities.get(&bob_user_id).unwrap().len(), 1);
```
That is, to use `all_identities` for the checks.

crypto/src/e2e_identity/identity.rs
```diff
@@ -520,4 +741,83 @@ mod tests {
...
+                        .try_talk_to(&id, &mut alices_ios_central)
+                        .await
+                        .unwrap();
+
+                    // same server communication
+                    bob_android_central
+                        .try_talk_to(&id, &mut alices_ios_central)
+                        .await
+                        .unwrap();
+                })
```
istankovic, line 818: This is actually an important test, happy to see it! :-)

istankovic (changes requested):
There's still some work to do.

crypto/src/e2e_identity/identity.rs
```diff
@@ -156,20 +156,18 @@ impl MlsConversation {
...
                 let uid = String::try_from(uid);
-                let identities = group.into_iter().map(|(_, id)| id).collect::<CryptoResult<Vec<_>>>();
                 // could be simplified if `Result::zip` was available
-                uid.and_then(|uid| identities.map(|ids| (uid, ids)))
+                uid.and_then(|uid| identity.map(|id| (uid, id)))
             })
-            .collect::<CryptoResult<HashMap<_, _>>>()
+            .process_results(|iter| iter.into_group_map())
     }
 }
```
istankovic, line 172: Could you please factor this out into a separate commit, so that it is not mixed in with a bunch of unrelated changes? It would be good to mention in the commit message what was wrong with previous code.

istankovic (approved):
Looks good. :+1:

crypto/src/e2e_identity/identity.rs
```diff
@@ -186,6 +190,54 @@ mod tests {
...
+            .get_user_identities(id, &["aaaaaaaaaaaaa".to_string()])
+            .await
+            .unwrap();
+        assert!(not_found.is_empty());
+
+        // Invalid usage
+        let invalid = central.get_user_identities(id, &[]).await;
+        assert!(matches!(invalid.unwrap_err(), CryptoError::ConsumerError));
+    }
+    async fn check_identities_device_status<const N: usize>(
```
istankovic, line 213: Missing blank line above this one.

PR #650, opened 2024-08-14
fix: run ci to generate junit report on tags [WPB-10608]

augustocdias:
# What's new in this PR

CI needs to generate junit report on tags

----

- The **PR Title**

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,6 +8,8 @@ on:
   push:
     branches:
       - develop
+    tags:
+      - '*'
   pull_request:
 
 env:


.github/workflows/rust.yml
```diff
@@ -1,3 +1,4 @@
+---
```
istankovic, line 1: What does this do?

augustocdias: this marks a document start... 

https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.document_start

istankovic:
It looks there are unrelated purely formatting changes in the same commit that does the actual change.
I would prefer to tackle the formatting changes separately. It's also not clear to me why the formatting changes were done.

augustocdias:
> It looks there are unrelated purely formatting changes in the same commit that does the actual change. I would prefer to tackle the formatting changes separately. It's also not clear to me why the formatting changes were done.

this was my linter... I haven't noticed it did this...

.github/workflows/rust.yml
istankovic: But from what I can read there, it may or may not be present, depending on the setting. This is not something required by github, right?

istankovic:
Could you please just make the originally intended change and leave the formatting out.
Because ideally we would settle on a common yaml formatter and tooling configuration, to ensure we all have the same set of rules. Plus, we would need to reformat all yaml CI files, not just in this repo, but all repos we currently maintain, to ensure consistency.

istankovic (approved):

PR #654, opened 2024-08-15
chore: update proteus version to latest tag

augustocdias:
# What's new in this PR

Update proteus dependency in CC to latest tag.

----

- The **PR Title**


Cargo.toml
```diff
@@ -71,7 +71,7 @@ branch = "otak/fix-1.0.3"
 [patch.crates-io.proteus-traits]
 package = "proteus-traits"
 git = "https://github.com/wireapp/proteus"
-branch = "2.x"
+tag = "v2.1.1"
```
istankovic, line 74: Please get rid of the patches completely in our toplevel Cargo.toml.
The `otak/fix-1.0.3` thing should be pushed down to the only crate that actually uses it (and not as a patch). That crate does not use the workspace dependency anyway.

keystore/Cargo.toml
```diff
@@ -110,7 +110,7 @@ pretty_env_logger = "0.5"
 [dev-dependencies.proteus-wasm]
 version = "2.1"
 git = "https://github.com/wireapp/proteus"
-branch = "2.x"
+tag = "v2.1.1"
```
istankovic, line 113: Can we just use workspace dependencies for all cases like this?

istankovic (changes requested):
Some more changes are necessary, in particular with respect to the patches and workspace dependencies.

istankovic:
I would also appreciate if you isolated the changes to `Cargo.toml` and `Cargo.lock` in their own separate commit, apart from the code changes.

Cargo.toml
augustocdias: The patch for the otak/fix cannot be removed, otherwise it conflicts with the one used by cryptobox in the interop crate. They're pointing to different commits...

augustocdias:
> I would also appreciate if you isolated the changes to `Cargo.toml` and `Cargo.lock` in their own separate commit, apart from the code changes.

I don't agree with this. One doesn't work without the other.

Cargo.toml
```diff
@@ -59,6 +59,10 @@ openmls_traits = { git = "https://github.com/wireapp/openmls", rev = "6d2f5a2046
 openmls_basic_credential = { git = "https://github.com/wireapp/openmls", rev = "6d2f5a2046fc12aca602af207e650b94848f09ff", version = "0.2" }
 openmls_x509_credential = { git = "https://github.com/wireapp/openmls", rev = "6d2f5a2046fc12aca602af207e650b94848f09ff", version = "0.2" }
 
+# proteus
+proteus-traits = { package = "proteus-traits", git = "https://github.com/wireapp/proteus", tag = "v2.1.1" }
+proteus-wasm = { package = "proteus-wasm", git = "https://github.com/wireapp/proteus", tag = "v2.1.1" }
```
istankovic, line 64: I think the `package = ...` is unnecessary since you're not renaming the package, no? We're using the same names so just
```
proteus-wasm = { git = "https://github.com/wireapp/proteus", tag = "v2.1.1" }
```
should work, since Cargo just uses the requested dependency name (the one on the left side of `=`), right?

crypto/Cargo.toml
```diff
@@ -55,15 +55,13 @@ base64.workspace = true
...
+workspace = true
 features = ["hazmat"]
 optional = true
-git = "https://github.com/wireapp/proteus"
-branch = "2.x"
 
 [dependencies.proteus-traits]
 optional = true
-version = "2.0"
+workspace = true
```
istankovic, line 64: I would join both of these (`proteus-wasm` and `proteus-traits`) with the above table and make them inline, like we did with `uniffi` above and `proteus-traits` further below. I.e.
```
proteus-traits = { workspace = true, optional = true}
proteus-wasm =  { workspace = true, optional = true, features = ["hazmat"] }
```

Cargo.toml
istankovic: Why not update cryptobox to use the otak/fix-1.0.3 branch?

istankovic (approved):
Nice job!

PR #672, opened 2024-08-29
feat: include entity type in aad [WPB-10108]

SimonThormeyer:
# What's new in this PR
See title

----

- The **PR Title**


keystore/src/entities/mod.rs
```diff
@@ -134,6 +134,8 @@ impl EntityFindParams {
 pub trait EntityBase: Send + Sized + Clone + PartialEq + Eq + std::fmt::Debug {
     type ConnectionType: DatabaseConnection;
     type AutoGeneratedFields;
+    /// Beware: if you change the value of this constant on any WASM entity, you'll need to do a data migration
+    ///     not only because it is used as reference to the object store names but also for the value of the aad.
```
istankovic, line 138: Appreciate the comment. :+1:

keystore/src/entities/platform/wasm/mls/epoch_encryption_keypair.rs
```diff
@@ -67,14 +67,14 @@ impl Entity for MlsEpochEncryptionKeyPair {
...
-        self.keypairs = Self::encrypt_data(cipher, self.keypairs.as_slice(), self.aad())?;
+        self.keypairs = Self::encrypt_data(cipher, self.keypairs.as_slice(), &self.aad()?)?;
         Self::ConnectionType::check_buffer_size(self.keypairs.len())?;
 
         Ok(())
     }
 
     fn decrypt(&mut self, cipher: &aes_gcm::Aes256Gcm) -> CryptoKeystoreResult<()> {
-        self.keypairs = Self::decrypt_data(cipher, self.keypairs.as_slice(), self.aad())?;
+        self.keypairs = Self::decrypt_data(cipher, self.keypairs.as_slice(), &self.aad()?)?;
```
istankovic, line 77: I'm wondering, if all the entities use the very same AAD, can we pull the `aad()` call out of entities and into `EntityBase`?

istankovic:
Commit `389ba18ffd61c12fa483df7c801d4f076c73b453` has message
```
refactor: MlsKeyPackage::find_all() in fetch_keypackages
```
It feels like a `use ` is missing in front of `MlsKeyPackage`.

keystore/src/connection/platform/wasm/storage.rs
```diff
@@ -131,29 +129,90 @@ impl WasmEncryptedStorage {
...
+                            match (key, value) {
+                                (Some(key), Some(value)) => {
+                                    result.push((key, value));
+                                    cursor.next(None).await?;
+                                }
+                                _ => break,
+                            }
+                        }
+                    }
+                }
```
istankovic, line 186: I know this code is copied from Rexie, but I can't shake the feeling that we could do better.

For instance, something like the below:
```rust
    let it = std::iter::from_fn(|| {
         let key = cursor.key()?;
         let value = cursor.value()?;
         
         if key.is_some() {
            cursor.next(None).await?
            Some((key, value))
         } else {
            None
         }
    });
    if let Some(limit) = limit {
        it.take(limit).collect::<Vec<_>>()
    } else {
        it.collect::<Vec<_>>()
    }
```

keystore/src/connection/platform/wasm/storage.rs
```diff
@@ -131,29 +129,90 @@ impl WasmEncryptedStorage {
         }
     }
 
+    /// Copied from Rexie.
+    pub(crate) async fn scan(
```
istankovic, line 133: Do we really need `pub` here?

keystore/src/connection/platform/wasm/storage.rs
```diff
@@ -242,12 +303,13 @@ impl WasmEncryptedStorage {
...
-                let transaction = rexie.transaction(&[collection], TransactionMode::ReadWrite)?;
-                let store = transaction.store(collection)?;
+            WasmStorageWrapper::Persistent(idb) => {
+                let transaction = idb.transaction(&[collection], TransactionMode::ReadWrite)?;
+                let store = transaction.object_store(collection)?;
                 for k in ids {
-                    let k = Uint8Array::from(k.as_ref());
-                    store.delete(k.into()).await?;
+                    let k = k.as_ref().to_vec();
+                    let request = store.delete(JsValue::from(k))?;
```
istankovic, line 311: Can a `JsValue` be constructed from a reference, to avoid creating a `Vec`?

.github/workflows/rust.yml
```diff
@@ -141,7 +141,7 @@ jobs:
       - run: |
           echo "CHROME_PATH=${{ steps.setup-chrome.outputs.chrome-path }}" >> $GITHUB_ENV
       - name: Run tests (wasm)
-        run: cargo test --target wasm32-unknown-unknown -p ${{ matrix.workspace }}
+        run: cargo test --target wasm32-unknown-unknown --workspace -p ${{ matrix.workspace }}
```
istankovic, line 144: What changed compared to before that we need this?

SimonThormeyer: We added the v1 keystore as a dependency

keystore/src/connection/platform/wasm/mod.rs
```diff
@@ -99,113 +101,29 @@ impl DatabaseConnection for WasmConnection {
         // - build: breaks after r9
         let version = version_number(version_major, version_minor, version_patch, version_pre);
```
istankovic, line 102: I realize this is out of scope for this PR, but nevertheless I'd like to mention here that we should completely decouple the db schema version from the core-crypto version. The db version should be a simple counter,  with no semantic connection to core-crypto version, like the generic platform has it.

keystore/src/connection/platform/wasm/migration.rs
```diff
@@ -0,0 +1,198 @@
...
+};
+use keystore_v_1_0_0::Connection as ConnectionV1_0_0;
+
+/// This is called from a while loop. The `from` argument represents the version the migration is performed from.
+/// The function will return the version number of the DB resulting from the migration.
+///
+/// To add a new migration, adjust the previous bottom match arm to return the current version,
+///     add a new match arm below that matches on that version, perform the migration workload
+///     and finally return `final_target`.
+pub(crate) async fn migrate(from: u32, final_target: u32, name: &str, key: &str) -> CryptoKeystoreResult<u32> {
```
istankovic, line 34: I'd rename `final_target` to `target_version`.

keystore/src/connection/platform/wasm/mod.rs
```diff
@@ -99,113 +101,29 @@ impl DatabaseConnection for WasmConnection {
...
+            // Migration is needed
+            existing_db.close();
+
+            while migrated_version < version {
+                migrated_version = migrate(migrated_version, version, &name, key).await?;
+            }
+
+            let open_request = factory.open(&name, Some(version))?;
+            open_request.await?
+        };
```
istankovic, line 123: I think this block that iteratively calls `migrate` should be better placed in the `migration` module (also: consider renaming the module to `migrations` -- if it's likely there will be more than one :wink: ).

So here we wouldn't need to be concerned about how migrations are performed, we just delegate that the migration module and continue with our work:
```rust

        let db = factory.open(&name, None)?.await?;
        let db = migrate(db); // takes ownership of `db`
        ...
```
In this case the pub `migrate` function would be the "driver" of all migrations and the looping is just a migration implementation detail.

keystore/src/connection/platform/wasm/migration.rs
```diff
@@ -0,0 +1,198 @@
...
+    const VERSION_NUMBER_V1_0_2: u32 = version_number(1, 0, 2, 0);
+    match from {
+        // The latest version that results from a migration must always map to "final_target"
+        //      to ensure convergence of the while loop this is called from.
+        0..=VERSION_NUMBER_V1_0_2 => {
+            // The version passed into this function must be the same as the one returned by this match arm.
+            // Will need to be adjusted once you add a new migration.
+            migrate_to_post_v_1_0_2(name, key, final_target).await?;
```
istankovic, line 42: A particular migration function, like `migrate_to_post_v_1_0_2` here, implicitly knows about the version it's migrating from, as well as the version it's migrating to. Given that, there should be no `final_target` in the function argument list -- the function should construct the version _internally_, and return the new db version (because the caller really shouldn't need to care about it).

Something like
```rust
   match from {
        0..=VERSION_NUMBER_V1_0_2 => migrate_from_1_0_2(name, key, ).await,
        VERSION_NUMBER_V1_1_1 => migrate_from_1_1_1(name, key, ).await,
        ...
    }
```
Note that I changed the migration function naming to mention the "from version" (which is known), instead of the "to version" (which we don't know). The rough idea here is, the migration function
- knows how to migrate _from_ a version
- it tells you which version it migrated _to_

This approach reduces the need to coordinate version numbers in several places.

keystore/src/connection/platform/wasm/migration.rs
```diff
@@ -0,0 +1,198 @@
...
+use keystore_v_1_0_0::Connection as ConnectionV1_0_0;
+
+/// This is called from a while loop. The `from` argument represents the version the migration is performed from.
+/// The function will return the version number of the DB resulting from the migration.
+///
+/// To add a new migration, adjust the previous bottom match arm to return the current version,
+///     add a new match arm below that matches on that version, perform the migration workload
+///     and finally return `final_target`.
+pub(crate) async fn migrate(from: u32, final_target: u32, name: &str, key: &str) -> CryptoKeystoreResult<u32> {
+    const VERSION_NUMBER_V1_0_2: u32 = version_number(1, 0, 2, 0);
```
istankovic, line 35: This should be pushed down to the particular migration function (see below comment).

keystore/src/connection/platform/wasm/migration.rs
```diff
@@ -193,6 +273,99 @@ where
     }
 }
 
+trait WasmMigrationUniqueExt: UniqueEntity {
+    type EntityTypeV1_0_0: UniqueEntityV1_0_0<ConnectionType = KeystoreDatabaseConnectionV1_0_0>;
+    async fn convert_from_v_1_2_0_or_earlier(
```
istankovic, line 278: Nitpick: would be nice to have one empty line above `async fn`.

keystore/src/connection/platform/wasm/migration.rs
```diff
@@ -40,23 +50,92 @@ pub(crate) async fn migrate(from: u32, final_target: u32, name: &str, key: &str)
...
+        .save(ProteusPrekey::COLLECTION_NAME, &mut proteus_prekeys)
+        .await?;
+    new_storage
+        .save(ProteusIdentity::COLLECTION_NAME, &mut proteus_identities)
+        .await?;
+    new_storage
+        .save(ProteusSession::COLLECTION_NAME, &mut proteus_sessions)
+        .await?;
+    // Migration finished
     new_storage.close()?;
```
istankovic, line 139: Is this transactional? What guarantees do we have that if there is an error in the middle of the above block of code, that database content is still valid, i.e. not partially migrated?

istankovic, line 139: I think it would be nice to consolidate the above two big blocks, perhaps via a macro, to reduce potential for mistakes.

Something like
```rust
migrate_entity!(ProteusPrekey);
migrate_entity!(ProteusIdentity);
migrate_entity!(ProteusSession);
...
```
This way you would only need to mention the entity type once.

keystore/src/connection/platform/wasm/migration.rs
```diff
@@ -193,6 +273,99 @@ where
...
+    type EntityTypeV1_0_0 = MlsKeyPackageV1_0_0;
+}
+
+impl WasmMigrationExt for PersistedMlsGroup {
+    type EntityTypeV1_0_0 = PersistedMlsGroupV1_0_0;
+}
+
+impl WasmMigrationExt for PersistedMlsPendingGroup {
+    type EntityTypeV1_0_0 = PersistedMlsPendingGroupV1_0_0;
+}
```
istankovic, line 347: I don't understand why we have a bunch of impls of `WasmMigrationExt` for entity types that don't have `find_unique`. Could you please explain how this works?

keystore/src/entities/platform/wasm/mls/epoch_encryption_keypair.rs
SimonThormeyer: Not sure what you mean? `aad()` is a method on `Entity` (which is a trait defined for SQLite and wasm seprately, and `aad()` is needed only on wasm).

istankovic: Since `decrypt_data` is always called with the result of a call to `instance.aad()`, I thought it'd be more robust to just make that call inside of `Entity::decrypt_data` instead of the callers.

istankovic: But feel free to ignore this, as it's orthogonal to the topic of this PR.

keystore/src/connection/platform/wasm/storage.rs
SimonThormeyer: Looks like this would only work with async iterators, right? Unfortunately they're still unstable

keystore/src/connection/platform/wasm/migrations.rs
```diff
@@ -0,0 +1,406 @@
...
+        0..=DB_VERSION_0 => migrate_to_version_1(name, key).await,
+        _ => Err(CryptoKeystoreError::MigrationNotSupported(from)),
+    }
+}
+
+/// With the current feature set of stable rust macros, we're not aware how to construct an identifier
+///     for each entity inside the macro.
+/// We need it for a variable to store the conversion result.
+/// So we have to take an identifier for each entity as an argument.
+/// Can be done better once std::concat_idents is stabilized, or we can use tuple indexing when ${index()} is stabilized (https://github.com/rust-lang/rust/pull/122808).
```
istankovic, line 83: Could you please break this line into two? It's currently 169 characters long, which is a bit too much.

keystore/src/connection/platform/wasm/migrations.rs
```diff
@@ -0,0 +1,406 @@
...
+        open_request.await.map_err(Into::into)
+    }
+}
+
+/// This is called from a while loop. The `from` argument represents the version the migration is performed from.
+/// The function will return the version number of the DB resulting from the migration.
+///
+/// To add a new migration, add a new match arm below the latest one.
+/// It must match on the version it migrates from, and call a function that performs the migration workload,
+///     which returns the version it migrates to, which is the same value as TARGET_VERSION in the function above at the time the migration is added.
```
istankovic, line 66: Nitpick: this line is unnecessarily indented and also a bit long, please consider breaking it.

keystore/src/connection/platform/wasm/migrations.rs
```diff
@@ -0,0 +1,406 @@
...
+    match from {
+        // The version that results from the latest migration must match TARGET_VERSION
+        //      to ensure convergence of the while loop this is called from.
+        0..=DB_VERSION_0 => migrate_to_version_1(name, key).await,
+        _ => Err(CryptoKeystoreError::MigrationNotSupported(from)),
+    }
+}
+
+/// With the current feature set of stable rust macros, we're not aware how to construct an identifier
+///     for each entity inside the macro.
```
istankovic, line 80: Nitpick: unnecessary indent.

keystore/src/connection/platform/wasm/migrations.rs
```diff
@@ -0,0 +1,406 @@
...
+            let mut storage_during_migration = WasmEncryptedStorage::new($key, wrapper_during_migration);
+
+            // See docstring. Here, alternatively, we'd use the identifiers constructed above for each entity
+            // or use tuple indexing once one of the required language features is stable.
+            let ( $( $conversion_result, )* ) = conversion_results;
+
+            $(
+                match $conversion_result {
+                    Ok(mut collection) => storage_during_migration.save(<$entity>::COLLECTION_NAME, &mut collection).await?,
+                    // If this error occurs, we assume (for now) that it is due to the old aad not working because the entity has already been migrated.
```
istankovic, line 113: Please consider breaking this line.

keystore/src/connection/platform/wasm/migrations.rs
```diff
@@ -0,0 +1,406 @@
...
+            .get_all::<Self::EntityTypeV1_0_0>(Self::COLLECTION_NAME, None)
+            .await?
+            .iter()
+            .map(|old_record| {
+                let serialized = postcard::to_stdvec(old_record)?;
+                postcard::from_bytes::<Self>(&serialized).map_err(Into::into)
+            })
+            .collect::<Vec<Result<Self, CryptoKeystoreError>>>()
+            .into_iter()
+            .collect::<Result<Vec<Self>, CryptoKeystoreError>>()?;
```
istankovic, line 298: Can we get rid of the first `collect` and `into_iter` here?

keystore/src/connection/platform/wasm/migrations.rs
```diff
@@ -0,0 +1,406 @@
...
+            ObjectStoreBuilder::new(ProteusSession::COLLECTION_NAME)
+                .auto_increment(false)
+                .add_index(IndexBuilder::new("id".into(), KeyPath::new_single("id")).unique(true)),
+        );
+    #[cfg(feature = "idb-regression-test")]
+    let idb_builder = idb_builder.add_object_store(ObjectStoreBuilder::new("regression_check").auto_increment(false));
+    idb_builder
+}
+
+pub trait WasmMigrationExt: Entity<ConnectionType = KeystoreDatabaseConnection>
```
istankovic, line 276: This trait does not need to be `pub`, right?

keystore/src/connection/platform/wasm/migration.rs
SimonThormeyer: We can now recover from partial migration caused by interruptions, everything else will be handled in WPB-10934

keystore/src/connection/platform/wasm/migrations.rs
```diff
@@ -0,0 +1,416 @@
...
+            (identifier_14, E2eiIntermediateCert),
+            (identifier_15, E2eiCrl),
+            (identifier_16, ProteusPrekey),
+            (identifier_17, ProteusIdentity),
+            (identifier_18, ProteusSession)
+        ]
+    )
+}
+
+fn get_builder(name: &str, version: u32) -> DatabaseBuilder {
```
istankovic, line 177: As I see it, `get_builder` defines the schema for the db. As such, I'm wondering whether the argument `version` makes sense.

For example, consider that we want to add another db version (say, 2), and a migration for it.
Suppose that version 2 replaces 1 db store with 2 new stores. Without embedding extra info about the new db version into `get_builder`, we cannot just call `get_builder(..., 2)`
because that would return an incorrect schema.

I suggest to have the version hardcoded with the builder. For instance,

```rust
fn get_builder_v0(name: &str) -> DatabaseBuilder {
    let idb_builder = DatabaseBuilder::new(name)
        .version(0)
    // a bunch of add_object_store calls
}

fn get_builder_v1(name: &str) -> DatabaseBuilder {
    let idb_builder = DatabaseBuilder::new(name)
        .version(1)
    // a bunch of add_object_store calls
}

fn get_builder_v2(name: &str) -> DatabaseBuilder {
    let idb_builder = DatabaseBuilder::new(name)
        .version(2)
    // a bunch of add_object_store calls
}
```
Now, one could reduce duplication by factoring out "a bunch of add_object_store calls" into some utility functions,  though we would need to be extra-careful when making changes to those utility functions in order to avoid breaking old migrations.

In cases you have a migration that merely adds another store, one could simply do
```rust
fn get_builder_v3(name: &str) -> DatabaseBuilder {
    let idb_builder = get_builder_v2(name).
        .version(3)
        .add_object_store(...)
}
```

istankovic (approved):
Massive work, good job!

PR #690, opened 2024-09-18
feat: create a keystore transaction struct to be used in the context [WPB-10914]

augustocdias:
# What's new in this PR

This creates the Keystore transaction for each platform in order to implement the context. Some questionable choices were made to either avoid having to redesign and rewrite the whole crate or conform with both WASM and other platforms at the same time.  Support for Proteus will come later and it will still be using the current implementation.

Comments were added into the PR to explain some approaches taken. There will be new changes in this crate being done in the next PRs as the need arrises and they won't be back ported to this PR. The intention is providing a base API for working with the transactions and enhancing it as needed.

----

- The **PR Title**


keystore/src/connection/platform/generic/mod.rs
```diff
@@ -27,8 +29,28 @@ pub struct SqlCipherConnection {
...
+    type Target = Transaction<'conn>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.transaction
+    }
+}
+
 unsafe impl Send for SqlCipherConnection {}
 unsafe impl Sync for SqlCipherConnection {}
+unsafe impl<'conn> Send for TransactionWrapper<'conn> {}
```
augustocdias, line 52: I don't like this, but since we're using the transaction in a single operation and afterwards dropping it, I don't think it should be a problem

keystore/src/entities/mod.rs
```diff
@@ -158,8 +153,19 @@ pub trait EntityBase: Send + Sized + Clone + PartialEq + Eq + std::fmt::Debug {
 
         Ok(ret)
     }
+    // FIXME: the unimplemented panics are intentional for now to help identifying where it needs
+    // to be adjusted to use the transaction instead
```
augustocdias, line 157: This is overwritten only for proteus. We should delete these in the future.

keystore/src/entities/mod.rs
```diff
@@ -172,6 +178,17 @@ cfg_if::cfg_if! {
             id: Vec<u8>,
         }
 
+        #[async_trait::async_trait(?Send)]
```
augustocdias, line 181: we should move this trait to the `EntityBase` when we support proteus.

keystore/src/entities/mod.rs
```diff
@@ -158,8 +153,19 @@ pub trait EntityBase: Send + Sized + Clone + PartialEq + Eq + std::fmt::Debug {
...
     }
+    // FIXME: the unimplemented panics are intentional for now to help identifying where it needs
+    // to be adjusted to use the transaction instead
+    async fn save(&self, _: &mut Self::ConnectionType) -> CryptoKeystoreResult<()> {
+        unimplemented!("This is implemented just for proteus as of now")
+    }
+    async fn insert(&self, _: &mut Self::ConnectionType) -> CryptoKeystoreResult<Self::AutoGeneratedFields> {
+        unimplemented!("This is implemented just for proteus as of now")
+    }
+    async fn delete(_: &mut Self::ConnectionType, _: StringEntityId<'_>) -> CryptoKeystoreResult<()> {
```
augustocdias, line 164: The delete originally took an array of ids to delete. We delete a single item in every usage of this function, so it was refactored to take a single id.

keystore/src/lib.rs
```diff
@@ -75,7 +79,7 @@ pub mod dummy_entity {
             _conn: &mut Self::ConnectionType,
             _id: &StringEntityId,
         ) -> CryptoKeystoreResult<Option<Self>> {
-            Ok(Some(DummyStoreValue))
+            Ok(Some(DummyStoreValue { _phantom: PhantomData }))
```
augustocdias, line 82: Forgot to fix this. It will be fixed in the next PR

keystore/src/transaction.rs
```diff
@@ -0,0 +1,340 @@
...
+use crate::{
+    connection::{Connection, DatabaseConnection, TransactionWrapper},
+    entities::{
+        E2eiAcmeCA, E2eiCrl, E2eiEnrollment, E2eiIntermediateCert, E2eiRefreshToken, EntityMlsExt, MlsCredential,
+        MlsEncryptionKeyPair, MlsEpochEncryptionKeyPair, MlsHpkePrivateKey, MlsKeyPackage, MlsPendingMessage,
+        MlsPskBundle, MlsSignatureKeyPair, PersistedMlsGroup, PersistedMlsPendingGroup, StringEntityId,
+    },
+    CryptoKeystoreError, CryptoKeystoreResult,
+};
+
```
augustocdias, line 17: Both the below enums were created to avoid having to rewrite and re-design the whole entity structure. There are several very strict restrictions imposed by the `EntityBase` trait that forbids us to use trait objects for example for a more elegant solution. What I had to end up doing is a bunch of from-into functions to convert types between each other. Some of those restrictions came from uniffi.

keystore/src/transaction.rs
```diff
@@ -0,0 +1,340 @@
...
+            id @ EntityId::E2eiEnrollment(_) => E2eiEnrollment::mls_delete(tx, id.as_id()).await,
+            id @ EntityId::E2eiRefreshToken(_) => E2eiRefreshToken::mls_delete(tx, id.as_id()).await,
+            id @ EntityId::E2eiAcmeCA(_) => E2eiAcmeCA::mls_delete(tx, id.as_id()).await,
+            id @ EntityId::E2eiIntermediateCert(_) => E2eiIntermediateCert::mls_delete(tx, id.as_id()).await,
+            id @ EntityId::E2eiCrl(_) => E2eiCrl::mls_delete(tx, id.as_id()).await,
+        }
+    }
+
+    #[cfg(target_family = "wasm")]
+    fn scan_collection_name(&self) -> &'static str {
```
augustocdias, line 161: Indexed DB needs the name of the collections that will be affected in the transaction before hand (🤯), so had to implement this exclusively for WASM.

keystore/src/entities/mod.rs
```diff
@@ -172,6 +178,17 @@ cfg_if::cfg_if! {
             id: Vec<u8>,
         }
 
+        #[async_trait::async_trait(?Send)]
+        pub trait EntityMlsExt: Entity {
+            async fn mls_delete<'a>(tx: &crate::connection::storage::WasmStorageTransaction<'a>, id: StringEntityId<'a>) -> CryptoKeystoreResult<()> {
+                tx.delete(Self::COLLECTION_NAME, id.as_slice()).await
+            }
+            async fn mls_save<'a>(&'a self, tx: &crate::connection::storage::WasmStorageTransaction<'a>) -> CryptoKeystoreResult<()>;
```
SimonThormeyer, line 186: Can we provide a default implementation here that will be used in most or all entities to avoid some code duplication?

keystore/src/entities/mod.rs
```diff
@@ -172,6 +178,17 @@ cfg_if::cfg_if! {
...
+            async fn mls_insert<'a>(&'a self, _tx: &crate::connection::storage::WasmStorageTransaction<'a>) -> CryptoKeystoreResult<Self::AutoGeneratedFields> {
+                return Err(CryptoKeystoreError::NotImplemented);
```
SimonThormeyer, line 188:
```suggestion
                Err(CryptoKeystoreError::NotImplemented)
```

keystore/src/connection/platform/wasm/storage.rs
```diff
@@ -26,9 +26,83 @@ use crate::{
...
+                    debug_assert!(result.is_some());
+                });
+            }
+        }
+        Ok(())
+    }
+
+    pub(crate) async fn save<R: Entity<ConnectionType = WasmConnection> + 'static>(
+        &self,
+        collection_name: &'static str,
```
SimonThormeyer, line 80: This argument is unecessary, you can just use `R::COLLECTION_NAME` below

keystore/src/transaction.rs
```diff
@@ -0,0 +1,340 @@
...
+    /// save, insert and delete functions in the `EntityBase` trait
+    /// FIXME: implement a transaction wrapper for the wasm platform to await on the transaction
+    pub async fn commit(&self) -> Result<(), CryptoKeystoreError> {
+        // we don't necessarily need the write lock here, but since we don't want additional
+        // operations being added to the queue, we get an exclusive lock
+        let mut operations = self.operations.write().await;
+        let mut conn = self.conn.borrow_conn().await?;
+        cfg_if::cfg_if! {
+            if #[cfg(target_family = "wasm")] {
+                let tables = operations.iter().map(|op| op.scan_collection_name()).collect::<Vec<_>>();
```
SimonThormeyer, line 240: What happens if this vec contains the same string twice? Is this okay for idb?

keystore/src/entities/mod.rs
augustocdias: For this specific one unfortunately no. I've tried. The compiler can't figure out the type for the second parameter.

keystore/src/entities/mod.rs
augustocdias: I've removed this function entirely later since there's no one using it and the few that were, don't fit in the transaction model

keystore/src/entities/mod.rs
augustocdias: Actually I've managed to make it work. Fixed in: https://github.com/wireapp/core-crypto/pull/695/commits/af6abef688cd89ab886d6aee49cb1f089c7dc3ba

keystore/src/connection/platform/wasm/storage.rs
augustocdias: Fixed in: https://github.com/wireapp/core-crypto/pull/695/commits/af6abef688cd89ab886d6aee49cb1f089c7dc3ba

keystore/src/transaction.rs
augustocdias: Fixed in: https://github.com/wireapp/core-crypto/pull/695/commits/9d7bae311457113a73dffd873dd9c78c8e9e23fc

keystore/src/entities/mod.rs
SimonThormeyer: Nice! Now to avoid repeating an empty impl block for every entity you can, AFAICT, do instead once 
```rust
impl<T: Entity> EntityMlsExt for T {}
```

keystore/src/connection/platform/wasm/storage.rs
```diff
@@ -26,9 +26,83 @@ use crate::{
...
+pub enum WasmStorageTransaction<'a> {
+    Persistent {
+        tx: idb::Transaction,
+        cipher: &'a aes_gcm::Aes256Gcm,
+    },
+    InMemory {
+        db: Rc<RefCell<InMemoryDB>>,
+        cipher: &'a aes_gcm::Aes256Gcm,
+    },
+}
```
SimonThormeyer, line 46: The comment reads like we don't really need the lifetime on a field, just on the type itself. 
Would it be better then to just do the following then? 

```rust
pub enum WasmStorageTransaction<'_a> {
    Persistent {
        tx: idb::Transaction,
        cipher: aes_gcm::Aes256Gcm,
    },
    InMemory {
        db: Rc<RefCell<InMemoryDB>>,
        cipher: aes_gcm::Aes256Gcm,
    },
}
```

keystore/src/connection/platform/wasm/storage.rs
```diff
@@ -116,7 +190,7 @@ impl WasmEncryptedStorage {
                 }
             }
             WasmStorageWrapper::InMemory(map) => {
-                if let Some(store) = map.get(collection) {
```
SimonThormeyer, line 119: Since we are touching the piece of code anyway, maybe we wanna get rid of this deep nesting which makes it hard to understand and use this more linear version instead? If you agree that it's a better style

```rust
WasmStorageWrapper::InMemory(map) => {
                let Some(store) = map.borrow().get(collection) else {
                    return Ok(None);
                };
                let Some(js_value) = store.get(id.as_ref()).cloned() else {
                    return Ok(None);
                };
                let Some(mut entity) = serde_wasm_bindgen::from_value::<Option<R>>(js_value)? else {
                    return Ok(None);
                };
                entity.decrypt(&self.cipher)?;
                Ok(Some(entity))
            }
 ```

File: keystore/src/entities/mod.rs (hunk 8 above)
Comment (augustocdias) [reply keystore/src/entities/mod.rs:186]: Good point. I'll try it out

Comment (augustocdias) [reply keystore/src/entities/mod.rs:186]: The problem is the few ones that just throw errors... what should we do in this case? We can't have it implementing the same struct twice.

File: keystore/src/connection/platform/wasm/storage.rs (hunk 12 above)
Comment (augustocdias) [reply keystore/src/connection/platform/wasm/storage.rs:46]: This doesn't compile. Rust doesn't accept specifying generic types or lifetimes that are not used.

File: keystore/src/connection/platform/wasm/storage.rs (hunk 13 above)
Comment (augustocdias) [reply keystore/src/connection/platform/wasm/storage.rs:119]: Fixed in: https://github.com/wireapp/core-crypto/pull/695/commits/a5d0af601d22f843c8706b15d3641fba14c72229

File: keystore/src/entities/mod.rs (hunk 8 above)
Comment (SimonThormeyer) [reply keystore/src/entities/mod.rs:186]: Hmm, right... I guess let's leave it with the repition for now, the refactor needed for that is out of scope then 

This throwing errors is such a design flaw. If they only throw errors the type shouldn't claim that it implements the trait :)

File: keystore/src/entities/platform/generic/proteus/identity.rs [hunk 14]
```diff
@@ -143,7 +143,7 @@ impl EntityBase for ProteusIdentity {
 
     async fn delete(
         conn: &mut Self::ConnectionType,
-        _ids: &[crate::entities::StringEntityId],
+        _ids: crate::entities::StringEntityId<'_>,
```
SimonThormeyer, line 146: this should be `id` if it's just one

keystore/src/entities/platform/generic/proteus/prekey.rs
```diff
@@ -131,18 +131,14 @@ impl EntityBase for ProteusPrekey {
 
     async fn delete(
         conn: &mut Self::ConnectionType,
-        ids: &[crate::entities::StringEntityId],
+        ids: crate::entities::StringEntityId<'_>,
```
SimonThormeyer, line 134: this should be id if it's just one

keystore/src/entities/platform/generic/proteus/session.rs
```diff
@@ -172,16 +172,12 @@ impl EntityBase for ProteusSession {
         Ok(conn.query_row("SELECT COUNT(*) FROM proteus_sessions", [], |r| r.get(0))?)
     }
 
-    async fn delete(conn: &mut Self::ConnectionType, ids: &[StringEntityId]) -> crate::CryptoKeystoreResult<()> {
+    async fn delete(conn: &mut Self::ConnectionType, ids: StringEntityId<'_>) -> crate::CryptoKeystoreResult<()> {
```
SimonThormeyer, line 175: again just one id

keystore/src/entities/platform/wasm/mls/credential.rs
```diff
@@ -38,14 +41,30 @@ impl EntityBase for MlsCredential {
...
+#[cfg_attr(not(target_family = "wasm"), async_trait::async_trait)]
+impl EntityMlsExt for MlsCredential {
+    async fn mls_save<'a>(&'a self, storage: &WasmStorageTransaction<'a>) -> crate::CryptoKeystoreResult<()> {
+        let _ = self.mls_insert(storage).await?;
+        Ok(())
+    }
 
+    async fn mls_insert<'a>(
+        &'a self,
+        storage: &WasmStorageTransaction<'a>,
```
SimonThormeyer, line 66: Can you make this parameter name consistent with the one in the trait?

SimonThormeyer (approved):
Looks good, with just some minor things to resolve

keystore/src/entities/platform/wasm/mls/credential.rs
augustocdias: It is auto implemented everywhere, so it is not applicable anymore

keystore/src/entities/platform/generic/proteus/session.rs
augustocdias: fixed by https://github.com/wireapp/core-crypto/pull/695/commits/7beae609399c01513b60bc51267a1be369c2048e

keystore/src/entities/platform/generic/proteus/prekey.rs
augustocdias: fixed by https://github.com/wireapp/core-crypto/pull/695/commits/7beae609399c01513b60bc51267a1be369c2048e

keystore/src/entities/platform/generic/proteus/identity.rs
augustocdias: fixed by https://github.com/wireapp/core-crypto/pull/695/commits/7beae609399c01513b60bc51267a1be369c2048e

PR #714, opened 2024-10-09
feat: expose API for upgrading from basic to x509 credentials for a conversation WPB-10161

typfel:
# What's new in this PR

Expose API for upgrading from basic to x509 credentials for a conversation. This is useful for a clients to address the race condition, which can happen if they got added to a conversation with their basic credentials while upgrading to x509 credentials. 

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/bindings/js/CoreCrypto.ts
+++ b/crypto-ffi/bindings/js/CoreCrypto.ts
@@ -1560,6 +1560,47 @@ export class CoreCrypto {
     }
 
     /**
+     * Creates an update commit which replaces your leaf containing basic credentials with a leaf node containing x509 credentials in the conversation.
+     *
+     * NOTE: you can only call this after you've completed the enrollment for an end-to-end identity, calling this without
+     * a valid end-to-end identity will result in an error.
+     *
+     * **CAUTION**: {@link CoreCrypto.commitAccepted} **HAS TO** be called afterward **ONLY IF** the Delivery Service responds
+     * '200 OK' to the {@link CommitBundle} upload. It will "merge" the commit locally i.e. increment the local group
+     * epoch, use new encryption secrets etc...
+     *
+     * @param conversationId - The ID of the conversation
+     *
+     * @returns A {@link CommitBundle}
+     */
+    async e2eiRotate(
+        conversationId: ConversationId
+    ): Promise<CommitBundle> {
+        try {
+            const ffiRet: CoreCryptoFfiTypes.CommitBundle =
+                await CoreCryptoError.asyncMapErr(
+                    this.#cc.e2ei_rotate(conversationId)
+                );
+
+            const gi = ffiRet.group_info;
+
+            const ret: CommitBundle = {
+                welcome: ffiRet.welcome,
+                commit: ffiRet.commit,
+                groupInfo: {
+                    encryptionType: gi.encryption_type,
+                    ratchetTreeType: gi.ratchet_tree_type,
+                    payload: gi.payload,
+                },
+            };
+
+            return ret;
+        } catch (e) {
+            throw CoreCryptoError.fromStdError(e as Error);
+        }
+    }
+
+    /**
      * Commits the local pending proposals and returns the {@link CommitBundle} object containing what can result from this operation.
      *
      * **CAUTION**: {@link CoreCrypto.commitAccepted} **HAS TO** be called afterwards **ONLY IF** the Delivery Service responds
--- a/crypto-ffi/bindings/js/CoreCrypto.ts
+++ b/crypto-ffi/bindings/js/CoreCrypto.ts
@@ -1235,14 +1235,13 @@ export class CoreCrypto {
                 custom?.keyRotationSpan,
                 custom?.wirePolicy
             );
-            const ret = await CoreCryptoError.asyncMapErr(
+            return await CoreCryptoError.asyncMapErr(
                 this.#cc.create_conversation(
                     conversationId,
                     creatorCredentialType,
                     config
                 )
             );
-            return ret;
         } catch (e) {
             throw CoreCryptoError.fromStdError(e as Error);
         }
@@ -1284,7 +1283,7 @@ export class CoreCrypto {
 
             const identity = mapWireIdentity(ffiDecryptedMessage.identity);
 
-            const ret: DecryptedMessage = {
+            return {
                 message: ffiDecryptedMessage.message,
                 proposals: ffiDecryptedMessage.proposals,
                 isActive: ffiDecryptedMessage.is_active,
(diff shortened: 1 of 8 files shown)


crypto-ffi/bindings/js/CoreCrypto.ts
```diff
@@ -1559,6 +1559,47 @@ export class CoreCrypto {
...
+                welcome: ffiRet.welcome,
+                commit: ffiRet.commit,
+                groupInfo: {
+                    encryptionType: gi.encryption_type,
+                    ratchetTreeType: gi.ratchet_tree_type,
+                    payload: gi.payload,
+                },
+            };
+
+            return ret;
```
SimonThormeyer, line 1597: Let's return this directly without the intermediate variable :)

SimonThormeyer (approved):
Just a nitpick, otherwise it looks good!

crypto-ffi/bindings/js/CoreCrypto.ts
typfel: Sure I could but all other functions use this pattern. I'd say either I change all functions or do nothing.

typfel: Alright went a head updated all functions

crypto/src/e2e_identity/rotate.rs
```diff
@@ -1067,5 +1067,89 @@ pub(crate) mod tests {
...
+                            let commit = alice_central.mls_central.e2ei_rotate(&id, None).await.unwrap();
+
+                            // Bob decrypts the commit...
+                            let decrypted = bob_central
+                                .mls_central
+                                .decrypt_message(&id, commit.commit.to_bytes().unwrap())
+                                .await
+                                .unwrap();
+                            // ...and verifies that now Alice is represented with her new identity
+                            alice_central.mls_central.verify_sender_identity(&case, &decrypted);
```
istankovic, line 1132: Hm, here it is Alice, not Bob that is doing sender verification -- is this intentional?

crypto/src/e2e_identity/rotate.rs
```diff
@@ -1067,5 +1067,89 @@ pub(crate) mod tests {
...
+                                .verify_local_credential_rotated(&id, new_handle, new_display_name)
+                                .await;
+
+                            let final_count = alice_central.mls_central.count_entities().await;
+                            assert_eq!(init_count.encryption_keypair, final_count.encryption_keypair);
+                            assert_eq!(
+                                init_count.epoch_encryption_keypair,
+                                final_count.epoch_encryption_keypair
+                            );
+                            assert_eq!(init_count.key_package, final_count.key_package);
```
istankovic, line 1147: This means a search for a basic credential key package would fail now, right?

crypto-ffi/bindings/gradle/libs.versions.toml
```diff
@@ -11,7 +11,7 @@ android-junit = "1.1.5"
 android-logback = "2.0.0"
 android-tools = "8.0.0"
 sdk-compile = "33"
-sdk-min = "26"
+sdk-min = "33"
```
istankovic, line 14: Why this change? If this is unrelated to the other changes in this commit, I'd prefer if this was moved into its own, separate commit, together with a message describing the "why".

istankovic:
The hunk(s) modifying crypto-ffi/src/generic.rs in commit 961e61a368dca1c1c76392377ebf5fdf695ad5bf should really be incorporated into commit b8b5ef92c1dbde014afa817e84888050ca849896, which introduces that code.

crypto-ffi/bindings/gradle/libs.versions.toml
typfel: When building the generated android bindings the compiler said it was using things SDK level 33. But you're I'll alright try to reproduce this on main and make separate PR for it.

crypto-ffi/bindings/jvm/src/main/kotlin/com/wire/crypto/client/MLSClient.kt
```diff
@@ -307,6 +307,20 @@ class MLSClient(private val cc: com.wire.crypto.CoreCrypto) {
...
     suspend fun updateKeyingMaterial(id: MLSGroupId) = cc.updateKeyingMaterial(id.lower()).lift()
 
+    /**
+     * Creates an update commit which replaces your leaf containing basic credentials with a leaf node containing x509 credentials in the conversation.
+     *
+     * NOTE: you can only call this after you've completed the enrollment for an end-to-end identity, calling this without
+     * a valid end-to-end identity will result in an error.
+     *
+     * **CAUTION**: [commitAccepted] **HAS TO** be called afterward **ONLY IF** the Delivery Service responds'200 OK' to the [CommitBundle] upload.
+     * It will "merge" the commit locally i.e. increment the local group epoch, use new encryption secrets etc...
```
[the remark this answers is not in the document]
typfel: Sure, I was just copy pasted from other similar methods which also generate commit bundles though.

crypto/src/e2e_identity/rotate.rs
typfel: Hmm, I inherited it from the test I based this test on but it looks wrong. I'll investigate

crypto/src/e2e_identity/rotate.rs
typfel: I mean this count doesn't distinguish. In this test the key packages aren't touched. 10 key packages are generated when the centrals are created but that's it.

I honestly now that I'm looking at it I don't see the point of asserting that the `key_package` or `epoch_encryption_keypair` count hasn't changed.

crypto/src/e2e_identity/rotate.rs
typfel: So yes, because `verify_sender_identity` verifies that the sender identity present in the decrypted message matches the identity in the central's credential bundle. 

Since it was Alice who sent the message, `verify_sender_identity` must be called on Alice's central.

crypto/src/e2e_identity/rotate.rs
istankovic: Yeah, then I would omit those checks in order to avoid confusion. :+1:

crypto-ffi/bindings/gradle/libs.versions.toml
typfel: Ok, we should definitely not raise the `sdk-min` to 33.

This comes from https://github.com/mozilla/uniffi-rs/issues/2226

I'll fix this in separate PR, the cleaner implementation is supposed to fallback to something else on older phones. I'll investigate a bit more if this is just linter issue or coming from the fact that we share code between android and java.

typfel: Ok, reverted this and issue was fixed in https://github.com/wireapp/core-crypto/pull/715

istankovic (approved):

PR #721, opened 2024-10-21
chore: update rusty-jwt-tools to latest release, 0.11.0

istankovic:
While updating to rusty-jwt-tools 0.11., we encountered failings tests due to
- version 0.11 correctly doing certificate validation (this was not the case before)
- incorrect construction of `CertificateBundle` instances, which were missing intermediate CA's certificate

This PR fixes the latter.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ openmls_x509_credential = { git = "https://github.com/wireapp/openmls", rev = "6
 proteus-traits = { git = "https://github.com/wireapp/proteus", tag = "v2.1.1" }
 proteus-wasm = { git = "https://github.com/wireapp/proteus", tag = "v2.1.1" }
 
-wire-e2e-identity = { git = "https://github.com/wireapp/rusty-jwt-tools", rev = "751f281ec75a6ae113584659b4b48940407c50f8", version = "0.9" }
+wire-e2e-identity = { git = "https://github.com/wireapp/rusty-jwt-tools", tag = "v0.11.0" }
 
 [profile.release]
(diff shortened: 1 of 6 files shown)


SimonThormeyer (approved):

Cargo.toml
```diff
@@ -61,7 +61,7 @@ openmls_x509_credential = { git = "https://github.com/wireapp/openmls", rev = "6
 proteus-traits = { git = "https://github.com/wireapp/proteus", tag = "v2.1.1" }
 proteus-wasm = { git = "https://github.com/wireapp/proteus", tag = "v2.1.1" }
 
-wire-e2e-identity = { git = "https://github.com/wireapp/rusty-jwt-tools", rev = "751f281ec75a6ae113584659b4b48940407c50f8", version = "0.9" }
+wire-e2e-identity = { git = "https://github.com/wireapp/rusty-jwt-tools", tag = "v0.11.0" }
```
coriolinus, line 64: This changes from specifying a minor version to specifying a patch version. That's probably fine, but I don't know if there's a standard policy or convention for how specifically we specify our dependency versions.

coriolinus (approved):

Cargo.toml
istankovic: I think we follow the usual `version = ...` pattern for things published on crates.io, however none of our internal crates are published (e.g. rusty-jwt-tools, proteus-related things, openmls fork...). For those we just point to a tag, like above.

SimonThormeyer (approved):
Good job!

PR #727, opened 2024-10-22
refactor: use log crate instead of tracing crate WPB-11542

typfel:
# What's new in this PR

The tracing log framework is not compatible with wasm if you use spans, which is main feature of the tracing system. I'm therefore I'm switching to collecting logs using the standard log crate.

----

- The **PR Title**


crypto-ffi/src/generic.rs
```diff
@@ -725,11 +723,9 @@ pub trait CoreCryptoCallbacks: std::fmt::Debug + Send + Sync {
 /// WARNING: This is a global setting. Calling it twice will cause errors.
 #[uniffi::export]
 pub fn set_logger(logger: std::sync::Arc<dyn CoreCryptoLogger>, level: CoreCryptoLogLevel) {
-    fmt::fmt()
-        .json()
-        .with_writer(CoreCryptoLoggerWrapper { logger, level })
-        .with_max_level(LevelFilter::from(level))
-        .init()
+    let logger = CoreCryptoLoggerWrapper { logger };
+    log::set_boxed_logger(Box::new(logger)).unwrap();
```
typfel, line 727: I assume we shouldn't unwrap here, but I guess that means we need to start returning an error?

crypto-ffi/src/wasm.rs
```diff
@@ -1122,40 +1113,22 @@ impl CoreCryptoWasmLogger {
...
-    type Writer = Self;
-
-    fn make_writer(&self) -> Self::Writer {
-        self.clone()
+    fn log(&self, record: &Record) {
+        let message = format!("{}", record.args());
+        let level: CoreCryptoLogLevel = self.level.into();
+        self.logger
+            .call2(&self.ctx, &JsValue::from(level), &JsValue::from(message))
+            .unwrap();
```
typfel, line 1128: I can't return an error here so should I simply swallow the error?

crypto-ffi/src/wasm.rs
```diff
@@ -1520,6 +1493,12 @@ impl CoreCrypto {
         }
     }
 
+    pub fn set_logger(logger: CoreCryptoWasmLogger, level: CoreCryptoLogLevel) {
+        log::set_boxed_logger(Box::new(logger))
+            .map(|()| log::set_max_level(level.into()))
+            .unwrap();
```
typfel, line 1499: Same thing goes here, we should start returning and error I guess.

crypto-ffi/src/generic.rs
istankovic: Hm, yeah, good question.
Generally, it is frowned upon to just panic inside a library when something bad happens and instead one is encouraged to return an error so the caller can handle it.

On the other hand, in this case the error can only happen if a client calls this function twice, which is a programmer error and we anyway expect this to be replaced in the next iteration of our logging mechanism so I guess it's fine to leave it as-is.

crypto/src/mls/external_commit.rs
```diff
@@ -76,21 +75,19 @@ impl MlsCentral {
...
-    /// serialized `GroupInfo` object
+    /// * `group_info` - a GroupInfo wrapped in a MLS message. it can be obtained by deserializing a TLS serialized `GroupInfo` object
     /// * `custom_cfg` - configuration of the MLS conversation fetched from the Delivery Service
     /// * `credential_type` - kind of [openmls::prelude::Credential] to use for joining this group.
-    /// If [MlsCredentialType::Basic] is chosen and no Credential has been created yet for it,
-    /// a new one will be generated. When [MlsCredentialType::X509] is chosen, it fails when no
-    /// [openmls::prelude::Credential] has been created for the given Ciphersuite.
+    ///   If [MlsCredentialType::Basic] is chosen and no Credential has been created yet for it,
+    ///   a new one will be generated. When [MlsCredentialType::X509] is chosen, it fails when no
+    ///   [openmls::prelude::Credential] has been created for the given Ciphersuite.
```
istankovic, line 83: Were these formatting changes made accidentally?

crypto-ffi/src/wasm.rs
istankovic: I'm not sure what kind of errors may pop up here, but we could perhaps consider using `web_sys::console` facilities to at least print out something in case this errors.

typfel: Unsure about printing things to console I think it's best to swallow the error.

crypto/src/mls/external_commit.rs
istankovic: Was that `cargo fmt` or something else?

crypto-ffi/src/wasm.rs
istankovic: Eh, the problem with swallowing errors is, if something unexpected happens, you're only going to see unexpected results on some higher layer, and the logs will be empty, leaving you wondering what really happened.

Admittedly, here it's a bit hard to come up with a scenario where this would fail, but OTOH, this does go through the wasm stack and if, for example, clients have a bug in their logging function, this is going to silently fail, and clients will wonder why they don't see anything in the logs. So it would be nice if the call fails here to dump the error to the JS console, perhaps it might even contain the traceback.

typfel: But the error will be happening in the web app, they can catch and log the error themselves. Sure maybe they forget and it's will nice inform them about it.

I'll go ahead add the console log fallback.

typfel: Ok, finally this this change and I added a test for it.

crypto/src/mls/external_commit.rs
typfel: I don't remember if it was clippy or fmt

istankovic (approved):
Looks fine, thanks!

PR #732, opened 2024-10-23
feat: implement transaction [WPB-11400] [WPB-10917]

SimonThormeyer:
# What's new in this PR
Transaction implementation

----

- The **PR Title**


crypto/src/lib.rs
```diff
@@ -20,7 +20,7 @@
 //! The goal is provide a easier and less verbose API to create, manage and interact with MLS
 //! groups.
 #![doc = include_str!("../../README.md")]
-#![deny(missing_docs)]
+#![cfg_attr(not(test), deny(missing_docs))]
```
istankovic, line 23: Why add this now?

istankovic:
Do we have a test for `ContextState::Invalid` i.e. a test that causes the context to become invalid and then checks that we return proper errors when attempting to use the context further?

keystore/src/entities/platform/generic/proteus/identity.rs
```diff
@@ -110,17 +111,15 @@ impl EntityBase for ProteusIdentity {
...
-        let transaction = conn.transaction()?;
-
+#[cfg_attr(target_family = "wasm", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_family = "wasm"), async_trait::async_trait)]
+impl EntityTransactionExt for ProteusIdentity {
+    async fn save(&self, transaction: &TransactionWrapper<'_>) -> crate::CryptoKeystoreResult<()> {
         use rusqlite::ToSql as _;
         transaction.execute(
-            "INSERT INTO proteus_identities (sk, pk) VALUES (?, ?)",
+            "INSERT INTO proteus_identities (sk, pk) SELECT ?, ? WHERE (SELECT COUNT(*) FROM proteus_identities) = 0",
```
istankovic, line 122: What is the reason for this change?

SimonThormeyer: To keep the logic that was present on current main, which would do an early return in this function:
```
if Self::count(conn).await? == 1 {
            return Ok(());
        }
```

keystore/src/entities/mod.rs
```diff
@@ -261,6 +261,12 @@ cfg_if::cfg_if! {
 
         pub trait Entity: EntityBase {
             fn id_raw(&self) -> &[u8];
+            
+            /// The query results that are obtained during a transaction
+            /// from the transaction cache and the database are merged by this key.
```
istankovic, line 297: Appreciate the comment!

crypto/src/lib.rs
SimonThormeyer: No idea, removed it in [chore: fix docs](https://github.com/wireapp/core-crypto/pull/732/commits/fd93bce72ef8b8867e76b3f37466dc9cd8f30088).

keystore/src/entities/platform/generic/proteus/identity.rs
SimonThormeyer: Just realized that this won't work as expected if there is a record in the persisted DB but none in the cache.

keystore/src/transaction.rs
```diff
@@ -80,379 +75,396 @@ impl EntityId {
...
-            MlsEntityId::EncryptionKeyPair => Self::EncryptionKeyPair(id.into()),
-            MlsEntityId::EpochEncryptionKeyPair => Self::EpochEncryptionKeyPair(id.into()),
-            MlsEntityId::GroupState => Self::PersistedMlsGroup(id.into()),
+            EntityId::ProteusIdentity(vec) => vec.as_slice().into(),
+            EntityId::ProteusSession(id) => id.as_slice().into(),
+            EntityId::ProteusPrekey(vec) => vec.as_slice().into(),
         }
     }
 
     fn from_collection_name(entity_id: &'static str, id: &[u8]) -> CryptoKeystoreResult<Self> {
```
istankovic: I feel this would be a good candidate for `const fn`.

keystore/src/transaction.rs
```diff
@@ -80,379 +75,396 @@ impl EntityId {
...
+        Ok(())
     }
-}
 
-impl KeystoreTransaction {
-    pub fn new(conn: Connection) -> Self {
-        Self {
-            conn,
-            operations: Default::default(),
+    pub(crate) async fn find<E>(&self, id: &[u8]) -> CryptoKeystoreResult<Option<Option<E>>>
```
istankovic: Why do we have two `Option`s in a row in the result?

keystore/src/transaction.rs
```diff
@@ -80,379 +75,396 @@ impl EntityId {
...
+            .merge_records(cached_records, persisted_records, EntityFindParams::default())
+            .await)
     }
 
-    pub async fn child_groups<
-        E: crate::entities::Entity<ConnectionType = KeystoreDatabaseConnection>
-            + crate::entities::PersistedMlsGroupExt
-            + Sync,
-    >(
+    async fn merge_records<E: crate::entities::Entity<ConnectionType = KeystoreDatabaseConnection>>(
```
istankovic: It would be nice to add a comment describing how the merging is actually performed, i.e. how you get from input records to the output.

istankovic:
We need to resolve code duplication in `src/transaction/dynamic_dispatch/{mls.rs,proteus_and_mls.rs}`.

keystore/src/connection/mod.rs
```diff
@@ -192,8 +194,31 @@ impl Connection {
...
+    ) -> CryptoKeystoreResult<Vec<E>> {
+        let mut conn = self.conn.lock().await;
+        let persisted_records = entity.child_groups(conn.deref_mut()).await?;
+
+        let transaction_guard = self.transaction.lock().await;
+        let Some(transaction) = transaction_guard.as_ref() else {
+            return Ok(persisted_records);
+        };
+        transaction.child_groups(entity, persisted_records).await
     }
```
istankovic, line 225: I don't understand this function, could you please described what it is supposed to do?

keystore/src/connection/mod.rs
```diff
@@ -227,10 +277,21 @@ impl FetchFromDatabase for Connection {
...
+        };
+        transaction.find_many(persisted_records, ids).await
     }
 
     async fn count<E: Entity<ConnectionType = KeystoreDatabaseConnection>>(&self) -> CryptoKeystoreResult<usize> {
+        if self.transaction.lock().await.is_some() {
+            // Unfortunately, we have to do this because of possible record id overlap 
+            // between cache and db.
+            return Ok(self.find_all::<E>(Default::default()).await?.len());
+        };
```
istankovic, line 330: This looks suspicious. Isn't `find_all` going to try to acquire the same lock that we got here?

keystore/src/entities/platform/generic/mls/credential.rs
```diff
@@ -33,6 +33,13 @@ impl Entity for MlsCredential {
     fn id_raw(&self) -> &[u8] {
         self.id.as_slice()
     }
+
+    fn merge_key(&self) -> Vec<u8> {
+        let mut merge_key = self.id_raw().to_vec();
+        // Credentials are unique by id and type, the type is contained in these bytes.
+        merge_key.extend(self.credential.to_vec());
```
istankovic, line 40: Can we just iterate over `self.credential` in  order to avoid creating a new `Vec`, just so we can iterate over it and then immediately discard it?

crypto/src/mls/client/identities.rs
```diff
@@ -5,106 +5,120 @@ use crate::{
...
-        self.0.get(&sc)?.iter().find(|c| {
+    ) -> Option<CredentialBundle> {
+        let credential_bundles_by_signature_schemes = self.0.lock().await;
+        let credential_bundles = credential_bundles_by_signature_schemes.get(&sc)?;
+        credential_bundles.iter().find(|c| {
             let ct_match = ct == c.credential.credential_type().into();
             let pk_match = c.signature_key.public() == pk.as_slice();
             ct_match && pk_match
-        })
+        }).cloned()
```
istankovic: Why do we clone bundles now?

keystore/src/transaction.rs
SimonThormeyer: Thanks for pointing out that this may need clarification! I'm adding the following docstring
```rust
/// The result of this function will have different contents for different scenarios: 
/// * `Some(Some(E))` - the transaction cache contains the record
/// * `Some(None)` - the deletion of the record has been cached
/// * `None` - there is no information about the record in the cache 
```

keystore/src/transaction.rs
SimonThormeyer: Thanks for pointing this out.
Added this docstring

```rust
    /// Build a single list of unique records from two potentially overlapping lists.
    /// In case of overlap, records in `records_a` are prioritized.
    /// Identity from the perspective of this function is determined by the output of [crate::entities::Entity::merge_key].
    ///
    /// Further, the output list of records is built with respect to the provided [EntityFindParams]
    /// and the deleted records cached in this [Self] instance.     

```

keystore/src/entities/platform/generic/proteus/identity.rs
SimonThormeyer: Not an issue because the calling site takes care of the case that is handled here. Resolving.

keystore/src/connection/mod.rs
SimonThormeyer: It collects all records that have the  id of `entity` as `parent_id`. I don't know in what context this may be needed, AFAICT this function isn't used anywhere. Should we remove it in this PR or in another?

keystore/src/connection/mod.rs
SimonThormeyer: The tests run and execute this codepath without a deadlock. Calling `.is_some()` probably drops the lock.

crypto/src/mls/client/identities.rs
SimonThormeyer: Couldn't figure out a way to return references out of the mutex; Is there one?

keystore/src/entities/platform/generic/mls/credential.rs
istankovic: Perfect!

SimonThormeyer:
> Do we have a test for `ContextState::Invalid` i.e. a test that causes the context to become invalid and then checks that we return proper errors when attempting to use the context further?

Added a test in the typescript wrapper test.

istankovic:
If an error occurs during `context.finish()`, we will propagate the error to the client, however we won't set the context's state to invalid. This opens up a possibility for the client to continue using the context (e.g. remembering the context variable in the scope outside transaction).

crypto-ffi/src/wasm/context/mod.rs
```diff
@@ -18,6 +18,8 @@ use wasm_bindgen::prelude::wasm_bindgen;
 use wasm_bindgen::JsValue;
 use wasm_bindgen_futures::future_to_promise;
 
+mod e2ei;
```
istankovic, line 21: Why is the the one from `generic` `pub`, and this one not?

SimonThormeyer:
> If an error occurs during `context.finish()`, we will propagate the error to the client, however we won't set the context's state to invalid. This opens up a possibility for the client to continue using the context (e.g. remembering the context variable in the scope outside transaction).

fixed in [chore: fix transactions in bindings](https://github.com/wireapp/core-crypto/pull/732/commits/307a045dfbd4ebd5a14b6b2d1f7cb04fc2638dea).

crypto-ffi/src/wasm/context/mod.rs
SimonThormeyer: Good catch, made it consistent to be pub in both places.

interop/src/clients/corecrypto/ffi.rs
```diff
@@ -101,85 +92,118 @@ impl<'a> EmulatedMlsClient for CoreCryptoFfiClient<'a> {
...
     async fn session_from_prekey(&mut self, session_id: &str, prekey: &[u8]) -> Result<()> {
-        let _ = self.cc.proteus_session_from_prekey(session_id, prekey)?;
+        let _ = self
+            .cc
+            .proteus_session_from_prekey(session_id.to_string(), prekey.to_vec())
+            .await?;
         Ok(())
     }
 
+    #[allow(deprecated)]
```
istankovic, line 182: Why do we need this?

keystore/src/transaction/mod.rs
```diff
@@ -283,30 +282,58 @@ impl KeystoreTransaction {
...
+                        (identifier_10, MlsPendingMessage),
+                        (identifier_11, E2eiEnrollment),
+                        (identifier_12, E2eiRefreshToken),
+                        (identifier_13, E2eiAcmeCA),
+                        (identifier_14, E2eiIntermediateCert),
+                        (identifier_15, E2eiCrl)
+                    ]
+                );
+            }
+        }
```
istankovic, line 336: Can we make this more concise by moving the Proteus-specific parts under their own `cfg_if` block?

interop/src/clients/corecrypto/ffi.rs
SimonThormeyer: WPB-12132

crypto-ffi/bindings/jvm/src/main/kotlin/com/wire/crypto/client/CoreCryptoCentral.kt
```diff
@@ -34,11 +34,19 @@ private class Callbacks : CoreCryptoCallbacks {
...
+            try {
+                result = block(CoreCryptoContext(context))
+            } catch (e: Throwable) {
+                error = e
+            }
         }
     })
+    if (error != null) {
+        throw error as Throwable
+    }
```
istankovic, line 49: Do we have a test for this?

crypto-ffi/bindings/js/CoreCrypto.ts
```diff
@@ -1043,15 +1043,27 @@ export class CoreCrypto {
...
+                        error = CoreCryptoError.fromStdError(
+                            e as Error
+                        ) as CoreCryptoError;
+                    }
+                }
+            },
+        });
+        if (error !== null) {
+            throw error;
+        }
```
istankovic, line 1066: Do we have a test for this?

crypto-ffi/bindings/js/test/CoreCrypto.test.js
```diff
@@ -225,6 +225,52 @@ test("can use groupInfo enums", async () => {
...
+                1
+            );
+        } catch (e) {
+            error = e;
+        }
+
+        return error;
+    });
+
+    expect(error.rustStackTrace).toBe("CryptoError(InvalidContext)");
```
istankovic, line 268: Thank you for this test! :purple_heart:

crypto-ffi/bindings/jvm/src/main/kotlin/com/wire/crypto/client/CoreCryptoCentral.kt
SimonThormeyer: Test was added

crypto-ffi/bindings/js/CoreCrypto.ts
SimonThormeyer: Test was added

crypto-ffi/bindings/js/test/CoreCrypto.test.js
```diff
@@ -315,6 +315,58 @@ test("JS Error is propagated by transaction", async () => {
     await page.close();
     await ctx.close();
 });
+test("Transaction is rolled back after error", async () => {
```
istankovic, line 318: Nit: missing one empty line between tests.

istankovic (approved):
Massive work, great job. :partying_face:

PR #739, opened 2024-10-30
feat: structured logging [WPB-11544]

typfel:
# What's new in this PR

* Change the logging system to support structured data
* Add debug logs for the decryption path in MLS and include relevant contextual information when available.

----

- The **PR Title**

### DIFF
--- a/crypto/src/obfuscate.rs
+++ b/crypto/src/obfuscate.rs
@@ -1,0 +1,86 @@
+use crate::prelude::{ClientId, ConversationId};
+use derive_more::{Constructor, From};
+use hex;
+use log::kv::{ToValue, Value};
+use openmls::framing::Sender;
+use openmls::group::QueuedProposal;
+use openmls::prelude::Proposal;
+use sha2::{Digest, Sha256};
+use std::fmt::{Debug, Formatter};
+
+pub(crate) trait Obfuscate {
+    fn obfuscate(&self, f: &mut Formatter<'_>) -> core::fmt::Result;
+}
+
+impl Obfuscate for &ConversationId {
+    fn obfuscate(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        f.write_str(hex::encode(compute_hash(self)).as_str())
+    }
+}
+
+impl Obfuscate for &ClientId {
+    fn obfuscate(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        f.write_str(hex::encode(compute_hash(self)).as_str())
+    }
+}
+
+impl Obfuscate for &Proposal {
+    fn obfuscate(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        f.write_str(match self {
+            Proposal::Add(_) => "Add",
+            Proposal::Update(_) => "Update",
+            Proposal::Remove(_) => "Remove",
+            Proposal::PreSharedKey(_) => "PreSharedKey",
+            Proposal::ReInit(_) => "ReInit",
+            Proposal::ExternalInit(_) => "ExternalInit",
+            Proposal::AppAck(_) => "AppAck",
+            Proposal::GroupContextExtensions(_) => "GroupContextExtensions",
+        })
+    }
+}
+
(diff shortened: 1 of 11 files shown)


augustocdias (approved):

crypto/src/obfuscate.rs
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+}
+
+pub(crate) struct Obfuscated<T> {
+    inner: T,
+}
+
+impl<T> Obfuscated<T> {
+    pub(crate) fn from(value: T) -> Self {
```
coriolinus, line 72: We're already using `derive_more`, right?  We could derive `derive_more::From`, `derive_more::Constructor`, or both.

crypto/src/obfuscate.rs
```diff
@@ -0,0 +1,99 @@
...
+            Sender::External(external_sender_index) => f.write_fmt(format_args!("External{:?}", external_sender_index)),
+            Sender::NewMemberProposal => f.write_str("NewMemberProposal"),
+            Sender::NewMemberCommit => f.write_str("NewMemberCommit"),
+        }
+    }
+}
+
+pub(crate) struct Obfuscated<T> {
+    inner: T,
+}
```
coriolinus, line 69: It's not wrong, but I'm curious why you used a named struct instead of a tuple struct here? This acts like a newtype, for which pattern I think tuple structs are the more common implementation.

crypto/src/obfuscate.rs
```diff
@@ -0,0 +1,99 @@
...
+impl<T> ToValue for Obfuscated<T>
+where
+    T: Obfuscate,
+{
+    fn to_value(&self) -> Value {
+        Value::from_debug(self)
+    }
+}
+
+fn compute_hash(bytes: &[u8]) -> Vec<u8> {
```
coriolinus, line 95: Microoptimization: we could save an allocation here by returning `[u8; 10] here instead of a vector. It also makes it more obvious that we're truncating the hash to discard even more information, to get just an opaque identifier.

Something like

```rust
let mut output = [0; 10];
output.copy_from_slice(&hasher.finalize().as_slice()[0..10]);
output
```

crypto/src/obfuscate.rs
```diff
@@ -0,0 +1,99 @@
...
+            Proposal::ExternalInit(_) => "ExternalInit",
+            Proposal::AppAck(_) => "AppAck",
+            Proposal::GroupContextExtensions(_) => "GroupContextExtensions",
+        })
+    }
+}
+
+impl Obfuscate for &QueuedProposal {
+    fn obfuscate(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        f.write_str(match self.proposal {
```
coriolinus, line 43: Could reduce code duplication here:

```rust
self.proposal.obfuscate(f)
```

I assume it's intentional that you can't tell from the `Obfuscate` implementation whether this is a `Proposal` instance or a `QueuedProposal`?

crypto/src/obfuscate.rs
```diff
@@ -0,0 +1,99 @@
...
+            Proposal::AppAck(_) => "AppAck",
+            Proposal::GroupContextExtensions(_) => "GroupContextExtensions",
+        })
+    }
+}
+
+impl Obfuscate for &Sender {
+    fn obfuscate(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Sender::Member(leaf_node_index) => f.write_fmt(format_args!("Member{}", leaf_node_index)),
```
coriolinus, line 59: The `f.write_fmt(format_args!(...))` style isn't wrong, but I'm curious why not just use the `write!` macro? It expands to essentially this same thing, as I understand it, but it's easier to read.

Like

```rust
write!(f, "Member{leaf_node_index}")
```

coriolinus (approved):
I did have some questions and comments, but no objections if you want to merge as-is.

crypto/src/obfuscate.rs
typfel: Great! Didn't know

PR #747, opened 2024-11-04
feat: support changing loggers at runtime WPB-11541

typfel:
# What's new in this PR

Support changing loggers and max log level at runtime.

also renamed `initLogger` in the bindings to `setLogger` since IMO init is not suitable anymore
when we allow replacing a logger.

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/bindings/js/CoreCrypto.ts
+++ b/crypto-ffi/bindings/js/CoreCrypto.ts
@@ -898,11 +898,14 @@ export enum CoreCryptoLogLevel {
 }
 
 /**
- * Initializes the global logger for Core Crypto and registers the callback. Can be called only once
+ * Initializes the global logger for Core Crypto and registers the callback.
+ *
+ * **NOTE:** you must call this after `await CoreCrypto.init(params)` or `await CoreCrypto.deferredInit(params)`.
+ *
+ * @deprecated use {@link CoreCrypto.setLogger} instead.
+ *
  * @param logger - the interface to be called when something is going to be logged
  * @param level - the max level that should be logged
- *
- * NOTE: you must call this after `await CoreCrypto.init(params)` or `await CoreCrypto.deferredInit(params)`.
  **/
 export function initLogger(
     logger: CoreCryptoLogger,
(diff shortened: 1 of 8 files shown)


istankovic:
@typfel this is a breaking change, right?

crypto-ffi/src/generic.rs
```diff
@@ -719,13 +721,27 @@ pub trait CoreCryptoCallbacks: std::fmt::Debug + Send + Sync {
     ) -> bool;
 }
 
+static INIT_LOGGER: Once = Once::new();
+static LOGGER: LazyLock<ReloadLog<CoreCryptoLoggerWrapper>> = LazyLock::new(|| {
```
augustocdias, line 725: This `ReloadLog` struct has a blocking lock inside. Are we sure we want to use it?

crypto-ffi/src/generic.rs
```diff
@@ -719,13 +721,27 @@ pub trait CoreCryptoCallbacks: std::fmt::Debug + Send + Sync {
...
+
 /// Initializes the logger
-/// WARNING: This is a global setting. Calling it twice will cause errors.
 #[uniffi::export]
-pub fn set_logger(logger: std::sync::Arc<dyn CoreCryptoLogger>, level: CoreCryptoLogLevel) {
-    let logger = CoreCryptoLoggerWrapper { logger };
-    log::set_boxed_logger(Box::new(logger)).unwrap();
-    log::set_max_level(LevelFilter::from(level));
+pub fn set_logger(logger: Arc<dyn CoreCryptoLogger>) {
+    LOGGER.handle().replace(CoreCryptoLoggerWrapper { logger }).unwrap();
```
augustocdias, line 734: Shouldn't we properly handle the error? I know its lock poison error, but better safe than sorry

typfel: Yes, we should 👍

typfel:
> @typfel this is a breaking change, right?

I kept the old API around in the wrapper but for iOS it would be a breaking change. I can fix that though.

crypto-ffi/src/generic.rs
typfel: How would you suggest we handle it? Start returning an error in this method? That would be a breaking change.

typfel:
> > @typfel this is a breaking change, right?
> 
> I kept the old API around in the wrapper but for iOS it would be a breaking change. I can fix that though.

I've re-introduced the old API  and re-named the new API. 

**old:** `set_logger`
**new:** `set_logger_only`

In the next release where we break the API we'll delete old method and rename the new API.

crypto-ffi/src/generic.rs
typfel: Added a commented to explain the unwrapping.

crypto-ffi/src/generic.rs
typfel: As discussed offline this should be fine.

crypto-ffi/bindings/js/CoreCrypto.ts
```diff
@@ -898,19 +898,44 @@ export enum CoreCryptoLogLevel {
...
+ * Initializes the global logger for Core Crypto and registers the callback.
+ *
+ * **NOTE:** you must call this after `await CoreCrypto.init(params)` or `await CoreCrypto.deferredInit(params)`.
+ *
+ * @deprecated use {@link CoreCrypto.setLogger} instead.
+ *
  * @param logger - the interface to be called when something is going to be logged
  * @param level - the max level that should be logged
- *
- * NOTE: you must call this after `await CoreCrypto.init(params)` or `await CoreCrypto.deferredInit(params)`.
```
coriolinus, line 905: I think this comment is still relevant, no?

PR #749, opened 2024-11-06
refactor: keep client initialization state internal [WPB-12004]

SimonThormeyer:
# What's new in this PR
See above

----

- The **PR Title**

### DIFF
--- a/crypto/src/mls/client/mod.rs
+++ b/crypto/src/mls/client/mod.rs
@@ -24,35 +24,38 @@ use crate::{
     mls::credential::{ext::CredentialExt, CredentialBundle},
     prelude::{
         identifier::ClientIdentifier, key_package::KEYPACKAGE_DEFAULT_LIFETIME, CertificateBundle, ClientId,
-        CryptoError, CryptoResult, MlsCentral, MlsCiphersuite, MlsCredentialType, MlsError,
+        CryptoError, CryptoResult, MlsCiphersuite, MlsCredentialType, MlsError,
     },
 };
-use async_lock::RwLockReadGuard;
+use async_lock::RwLock;
 use core_crypto_keystore::{connection::FetchFromDatabase, Connection, CryptoKeystoreError};
 use log::debug;
 use openmls::prelude::{Credential, CredentialType};
 use openmls_basic_credential::SignatureKeyPair;
 use openmls_traits::{crypto::OpenMlsCrypto, types::SignatureScheme, OpenMlsCryptoProvider};
 use std::collections::HashSet;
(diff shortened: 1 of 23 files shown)


typfel:
@SimonThormeyer set the target branch on the PR to `feat/transaction` so that the diff makes sense.

crypto/src/mls/client/mod.rs
```diff
@@ -192,12 +213,17 @@ impl Client {
...
             id: client_id.clone(),
             identities: ClientIdentities::new(stored_skp.len()),
             keypackage_lifetime: KEYPACKAGE_DEFAULT_LIFETIME,
         };
 
+        let mut inner_lock = self.state.lock().await;
+        *inner_lock = Some(client);
+        // Drop the lock because it will be acquired below when the identities are saved.
+        drop(inner_lock);
+
```
coriolinus, line 226: You use this pattern several times. We can automate it a bit:

```rust
async fn replace_inner(&self, new_inner: ClientInner) {
    let mut inner_lock = self.state.lock().await;
    *inner_lock = Some(new_inner);
}
```

Which lets us simplify the usage:

```rust
self.replace_inner(ClientInner{
    id: client_id.clone(),
    identities: ClientIdentities::new(stores_skp.len()),
    keypackage_lifetime: KEYPACKAGE_DEFAULT_LIFETIME,
}).await;
```

crypto/src/mls/client/mod.rs
```diff
@@ -454,14 +515,6 @@ impl Client {
     }
 }
 
-impl PartialEq for Client {
```
coriolinus, line 457: No reason not to keep implementing `PartialEq` and `Eq` for `Client`:

```rust
let Some(ClientInner{id: sid,..}) = self.state.lock().await.deref() else { return false };
let Some(ClientInner{id: oid,..}) = other.state.lock().await.deref() else { return false };
sid == oid
```

coriolinus (commented):
Not approving yet because I think swapping out the `Mutex` for a `RwLock` is actually important. The rest of the comments can be considered advisory/recommendations but not critical. But overall LGTM.

crypto/src/mls/client/mod.rs
coriolinus: I was wrong here: of course we can't `.await` within a non-async function, such as `PartialEq`.

coriolinus (approved):
Looks great! We can defer the question of replacing `async-lock` with `tokio` until later. If we happen to start using surrealdb, and if that brings in `tokio` already, then it's relatively much easier to make that switch.

PR #757, opened 2024-11-12
feat: implement set_client_data() [WPB-10919]

SimonThormeyer:
# What's new in this PR
See title and corresponding Jira item

----

- The **PR Title**


keystore/src/transaction/mod.rs
```diff
@@ -341,12 +341,13 @@ impl KeystoreTransaction {
...
                 (identifier_14, E2eiIntermediateCert),
-                (identifier_15, E2eiCrl)
+                (identifier_15, E2eiCrl),
+                (identifier_16, ConsumerData)
             ],
             proteus_types: [
-                (identifier_16, ProteusPrekey),
-                (identifier_17, ProteusIdentity),
-                (identifier_18, ProteusSession)
+                (identifier_17, ProteusPrekey),
```
coriolinus, line 348: I don't know how these identifiers are used. Are these identifiers persisted at all? Can anything go wrong by moving `identifier_16` from `ProteusPrekey` to `ConsumerData`?

coriolinus (approved):
Had some questions, but I believe this should be good to go.

keystore/src/connection/platform/generic/migrations/V13__consumer_data.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE consumer_data (
+    id INTEGER PRIMARY KEY CHECK ( id = 0 ),
```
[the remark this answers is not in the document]
SimonThormeyer: Yes, exactly. This pattern can also be seen in other tables where we only ever expect one record, such as `e2ei_acme_ca`.

keystore/src/transaction/mod.rs
SimonThormeyer: These identifiers are merely used to hold the data from the in-memory cache before it is committed. So it doesn't matter what their name is/previously was at all. 

If we can create them internally in the macro instead of passing them in, I would like doing so!

coriolinus: Creating idents like this within a macro is tricky but not impossible. 

In the future, it will be easier: combine the [`${index()}` metafunction](https://github.com/rust-lang/rust/pull/122808) with the [`paste!` macro](https://crates.io/crates/paste) and we should be able to generate all these idents automatically. 

For now, `${index()}` is still not stable, so you have to use [recursive macro tricks](https://veykril.github.io/tlborm/decl-macros/building-blocks/counting.html#bit-twiddling) to build up a unary number for each index and then count how many symbols appear in that number. This approach is complicated and opaque enough I'd recommend not reworking the existing macro until `${index()}` stabilizes.

PR #771, opened 2024-11-20
feat: implement binding specific errors [WPB-14352]

coriolinus:
# What's new in this PR

Clients only care about a few error types, so let's surface those at the client interface and hide the rest.

(We can log the full detailed errors at the conversion point, but that's a separate task.)

- [x] binding-specific errors for WASM targets
- [x] binding-specific errors for uniffi targets
----

- The **PR Title**


crypto-ffi/src/wasm/mod.rs
```diff
@@ -78,96 +78,195 @@ pub extern "C" fn __stack_chk_fail() {
...
+// This implementation is intended to be temporary; we're going to be completely restructuring the way we handle
+// errors in `core-crypto` soon. We can replace this with better error patterns when we do.
+//
+// Certain error mappings could apply to both MLS and Proteus. In all such cases, we map them to the MLS variant.
+// When we redesign the errors in `core-crypto`, these ambiguities should disappear anyway.
+impl From<CryptoError> for InternalError {
+    fn from(value: CryptoError) -> Self {
+        #[cfg(feature = "proteus")]
+        if value.proteus_error_code().is_some() {
+            return ProteusError::Other(value.proteus_error_code()).into();
```
typfel, line 173: Hmm so right we only return `ProteusError:Other` but we could map on the error code code see `ProteusErrorKind`

coriolinus: Good point! Made that change.

crypto-ffi/bindings/js/test/CoreCrypto.test.ts
```diff
@@ -380,13 +382,22 @@ describe("proteus", () => {
...
+                message
+            );
+        } catch (e) {
+            didfail = true;
+            const error = e as Error;
+            expect(error.message).toBe(
+                "ProteusErrorOther: Another Proteus error occurred but the details are probably irrelevant to clients"
+            );
+        }
+        expect(didfail).toBeTruthy();
```
SimonThormeyer, line 400: The pattern before was more concise IMO, and should be equivalent.
So you could keep that pattern and just replace the expected error with
```
new Error("ProteusErrorOther: Another Proteus error occurred but the details are probably irrelevant to clients")
```

SimonThormeyer (approved):
Looks good! Just one thing I would have done differently in the tests

crypto-ffi/Cargo.toml
```diff
@@ -29,7 +29,12 @@ tls_codec.workspace = true
 async-lock.workspace = true
 log.workspace = true
 log-reload.workspace = true
+<<<<<<< HEAD
 serde_json.workspace = true
+=======
+proteus-wasm = { workspace = true, optional = true }
+core-crypto-keystore = { workspace = true, optional = true}
+>>>>>>> 75bea838 (feat: refactor WASM error types)
```
istankovic: Somehow we ended up with conflict markers in this commit.
Could you please fix this?

README.md
```diff
@@ -171,29 +171,56 @@ wasm-pack test --headless --chrome ./<crate-folder-to-test>
...
-### Kotlin/Android
+Build the JVM target every timee the Rust code changes:
 
-* Take the steps to compile for Kotlin/Android
-* Then:
-```ignore
-cd crypto-ffi/bindings
-./gradlew test
+```sh
+core-crypto/crypto-ffi$ cargo make jvm-linux
```
SimonThormeyer, line 179: or, on mac os,
`cargo make jvm-darwin`.

crypto-ffi/bindings/js/test/CoreCrypto.test.ts
coriolinus: Huh, you're right. At the time I thought I had a good reason to change it, but your suggestion works, so I'm going to use that.

crypto-ffi/src/wasm/mod.rs
```diff
@@ -79,96 +79,201 @@ pub extern "C" fn __stack_chk_fail() {
...
-    E2eError(#[from] E2eIdentityError),
+    ProteusError(#[from] ProteusError),
+    #[error("End to end identity error")]
+    E2eiError,
     #[error(transparent)]
     SerializationError(#[from] serde_wasm_bindgen::Error),
-    #[error("Failed lifting an enum")]
-    EnumError,
+    #[error("Unknown cyphersuite identifier")]
+    UnknownCyphersuite,
```
istankovic, line 183: I'm also a fan of British English, but would suggest to stick to the `ciphersuite` spelling, since that's what the rest of the code uses, both in Wire and the external projects. :-)

crypto-ffi/src/wasm/mod.rs
```diff
@@ -79,96 +79,201 @@ pub extern "C" fn __stack_chk_fail() {
...
+// errors in `core-crypto` soon. We can replace this with better error patterns when we do.
+//
+// Certain error mappings could apply to both MLS and Proteus. In all such cases, we map them to the MLS variant.
+// When we redesign the errors in `core-crypto`, these ambiguities should disappear anyway.
+impl From<CryptoError> for InternalError {
+    fn from(value: CryptoError) -> Self {
+        #[cfg(feature = "proteus")]
+        if let Some(error_code) = value.proteus_error_code() {
+            return match error_code {
+                102 => ProteusError::SessionNotFound, // reviewers take note: is this correct?
```
istankovic: Yes, I would say this corresponds to what's on
https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/689045561/Proteus+decryption+error+code+explanation

and the code in proteus repo, so feel free to remove the `reviewers ...` comment.

crypto-ffi/src/wasm/mod.rs
```diff
@@ -79,96 +79,201 @@ pub extern "C" fn __stack_chk_fail() {
...
-    fn proteus_error_code(&self) -> u32 {
-        let WasmError::CryptoError(e) = &self.0 else {
-            return 0;
-        };
+impl InternalError {
+    fn variant_name(&self) -> String {
+        let mut out = self.as_ref().to_string();
+        match self {
+            Self::MlsError(mls) => out += mls.as_ref(),
+            Self::ProteusError(proteus) => out += proteus.as_ref(),
```
istankovic, line 232: How does the resulting `out` string look like? This code just concatenates two strings, without any separator.
Can you provide a few examples?

crypto-ffi/src/wasm/mod.rs
```diff
@@ -79,96 +79,201 @@ pub extern "C" fn __stack_chk_fail() {
...
+        fn recurse(stack: &mut Vec<String>, err: &dyn std::error::Error) {
+            stack.push(err.to_string());
+            if let Some(source) = err.source() {
+                recurse(stack, source);
+            }
+        }
+
+        let mut stack = Vec::new();
+        recurse(&mut stack, self);
+        stack
```
istankovic, line 248: How about a simple iterative approach that avoids... the stack? :-)
```rust
    fn stack(&self) -> Vec<String> {
        let mut stack = Vec::new();
        let mut err = self;
        while let Some(source) = err.source() {
            stack.push(err.to_string());
            err = source;
        }
        stack
    }
```

crypto-ffi/src/generic/mod.rs
```diff
@@ -88,22 +88,143 @@ pub fn build_metadata() -> BuildMetadata {
...
+            102 => Self::SessionNotFound, // reviewers take note: is this correct?
+            204 => Self::RemoteIdentityChanged,
+            209 => Self::DuplicateMessage,
+            _ => Self::Other(Some(code)),
+        }
+    }
+
+    pub fn error_code(&self) -> Option<u16> {
+        match self {
+            ProteusError::SessionNotFound => Some(102), // reviewers: probably???
```
istankovic: Yes, this is correct, feel free to drop the comment.

crypto-ffi/src/generic/mod.rs
```diff
@@ -88,22 +88,143 @@ pub fn build_metadata() -> BuildMetadata {
...
+
+impl ProteusError {
+    pub fn from_error_code(code: impl Into<Option<u16>>) -> Option<Self> {
+        let code = code.into()?;
+        if code == 0 {
+            return None;
+        }
+
+        match code {
+            102 => Self::SessionNotFound, // reviewers take note: is this correct?
```
istankovic, line 141: Yes, this is correct, feel free to drop the comment.

README.md
```diff
@@ -171,29 +171,57 @@ wasm-pack test --headless --chrome ./<crate-folder-to-test>
...
+Build the WASM target every timee the Rust code changes:
+
+```sh
+core-crypto/crypto-ffi$ cargo make wasm
+```
+
+Then run the tests each time the wrapper or wrapper tests change
+
+```sh
+core-crypto/crypto-ffi$ bun run wdio
```
istankovic, line 218: It might be useful to mention the need for `bun run install` somewhere, since that's required before running tests.

istankovic (approved):
Left a few comments, looks good. :+1:

PR #777, opened 2024-11-22
feat: refactor the Kotlin wrapper and hide the generated uniffi api [WPB-14473]

typfel:
# What's new in this PR

- The previous Kotlin wrapper has been re-written to be more close the web wrapper.
- The deprecated non-transactional methods have been removed.
- The generated UniFFI Kotlin bindings have been moved into their own library so that we can avoid exposing them to consumer of our library.

----

- The **PR Title**


crypto-ffi/bindings/jvm/src/main/kotlin/com/wire/crypto/CoreCryptoException.kt
```diff
@@ -0,0 +1,346 @@
...
+//        fun fromCoreCryptoException(proteusLastErrorCode: UShort?, err: CoreCryptoException): ProteusException {
+//            var code = ProteusException.Code.UNKNOWN_ERROR
+//            if (proteusLastErrorCode != null) {
+//                code = ProteusException.fromProteusCode(proteusLastErrorCode)
+//            }
+//
+//            return ProteusException(err.message, code, err.cause)
+//        }
+//    }
+//}
```
coriolinus, line 346: Any reason to keep a whole bunch of dead code here? Seems like we could just remove it.

SimonThormeyer:
@typfel this is a breaking change, right? Can you append append to the conventional commit prefix an !, where applicable?
e.g., `refactor!: ...`. (It feels a bit like a category error that a refactor can be a breaking change, maybe it's rather a `build!: ...`?

coriolinus (approved):
This makes sense to me! I have to admit I did not look into the gradle stuff too deeply--I haven't really worked with that at all--but I assume that the build/test step passing on CI means that it has succeeded.

PR #785, opened 2024-11-28
chore: add transaction bench and web bench

SimonThormeyer:
# What's new in this PR
Adds benches and runs them in CI. The benches were used for an investigation regarding transactions and group store.

----

- The **PR Title**

### DIFF
--- a/crypto/benches/transaction.rs
+++ b/crypto/benches/transaction.rs
@@ -1,0 +1,71 @@
+use criterion::async_executor::FuturesExecutor;
+use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
+use openmls_traits::types::Ciphersuite;
+use rand::distributions::{Alphanumeric, DistString};
+
+pub mod utils;
+
+use utils::*;
+fn decrypt_transaction(c: &mut Criterion) {
+    const MESSAGE_LENGTH: usize = 100;
+    const MESSAGE_COUNT: usize = 1000;
+
+    let mut group = c.benchmark_group(format!("Decrypting {MESSAGE_COUNT} messages of {MESSAGE_LENGTH} bytes"));
+    let (ciphersuite, credential, in_memory) = (
+        Ciphersuite::MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519.into(),
+        None,
+        false,
+    );
+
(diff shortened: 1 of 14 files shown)


crypto/benches/transaction.rs
```diff
@@ -0,0 +1,74 @@
...
+pub mod utils;
+
+use utils::*;
+fn decrypt_transaction(c: &mut Criterion) {
+    const MESSAGE_LENGTH: usize = 100;
+    const MESSAGE_COUNT: usize = 1000;
+
+    let mut group = c.benchmark_group(format!(
+        "Decrypting {MESSAGE_COUNT} messages, transaction vs. no transaction"
+    ));
```
coriolinus, line 15: This group name shows up in the output, and it's not obvious which case is running for any given results. Recommend creating the group slightly later within the `transactional` loop, and naming it like `Decrypting {MESSAGE_COUT} messages: {id}`.

crypto/benches/transaction.rs
```diff
@@ -0,0 +1,74 @@
...
+        });
+    }
+    group.finish();
+}
+
+criterion_group!(
+    name = decryption;
+    config = criterion();
+    targets =
+    decrypt_transaction,
```
coriolinus, line 72: This would be easier to read if it were on a single line.

coriolinus (approved):
I suspect the CI failures are spurious, but it would be good after https://github.com/wireapp/core-crypto/pull/789#pullrequestreview-2469742941 is merged to rebase this atop that to ensure that it gets clean runs.

Approving anyway for now; the code looked right, except for some nits below.

crypto/benches/transaction.rs
SimonThormeyer: @coriolinus the output looks like this:
`Decrypting 1000 messages, transaction vs. no transaction/transactional` and
`Decrypting 1000 messages, transaction vs. no transaction/non-transactional`, where the first is transactional sending and the second non-transactional.
I'd like to keep the group creation in that place, as I follow the pattern of the other existing benches. I admint those might need an overhaul, though.

However, I made the output a bit more concise and added the information of message length in bytes.

coriolinus: Yes, that's clearer now, thanks!

PR #786, opened 2024-11-28
fix: stop exposing wipe() and unload() since they are broken in Kotlin [WPB-14514]

typfel:
# What's new in this PR

Stop exposing `wipe()` and `unload()`

Clients should instead call close() which will close the  database connection, it's then safe to delete the core crypto database.

----

- The **PR Title**

### DIFF
--- a/interop/src/clients/corecrypto/ffi.rs
+++ b/interop/src/clients/corecrypto/ffi.rs
@@ -18,7 +18,6 @@ use color_eyre::eyre::Result;
 use core_crypto_ffi::{
     context::TransactionHelper, ClientId, CoreCrypto, CustomConfiguration, MlsCredentialType, UniffiCustomTypeConverter,
 };
-use std::sync::Arc;
 
 use crate::{
     clients::{EmulatedClient, EmulatedClientProtocol, EmulatedClientType, EmulatedMlsClient},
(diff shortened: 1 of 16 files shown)


istankovic:
Should we also remove related functions from the `crypto` crate?

typfel:
> Should we also remove related functions from the `crypto` crate?

It's still used by the wasm implementation

interop/src/clients/corecrypto/ffi.rs
```diff
@@ -33,15 +32,18 @@ pub(crate) struct CoreCryptoFfiClient {
     prekey_last_id: u16,
 }
 
+const KEY_STORE_PATH: &str = "path";
+const KEY_STORE_KEY: &str = "path";
```
coriolinus, line 36: This used to be `"key"` when it was in the `CoreCrypto::new` invocation as a literal.

coriolinus (approved):

interop/src/clients/corecrypto/ffi.rs
```diff
@@ -29,11 +30,11 @@ pub(crate) struct CoreCryptoFfiClient {
...
     client_id: Vec<u8>,
     #[cfg(feature = "proteus")]
-    prekey_last_id: u16,
+    prekey_last_id: Cell<u16>,
 }
 
-const KEY_STORE_PATH: &str = "path";
-const KEY_STORE_KEY: &str = "path";
+const KEY_STORE_PATH: &str = "ffi_keystore";
+const KEY_STORE_KEY: &str = "key";
```
istankovic, line 37: Nit: this hunk should be moved to the previous commit, since that's the one that introduced the wrong values.

istankovic:
I noticed that the keystore still has `wipe` -- assuming it's not being used anymore, shouldn't we remove that too?

typfel:
> I noticed that the keystore still has `wipe` -- assuming it's not being used anymore, shouldn't we remove that too?

It's still used extensively in our tests which use on disk key stores, like the proteus tests.

istankovic (approved):

PR #791, opened 2024-11-29
refactor: errors per module [WPB-14354]

coriolinus:
# What's new in this PR

Refactors `core-crypto` such that each logical module defines its own internal error type, and adds context whenever wrapping another module's error type.

- [x] `core_crypto::e2e_identity`
- [x] `core_crypto::mls::client`
- [x] `core_crypto::mls::conversation`
- [x] `core_crypto::mls::credential`
- [x] `core_crypto::mls`
- [x] `core_crypto`

Logical modules are assumed to be the set of peer modules at a particular level in the directory tree.

### Verbose `{context, source}` variants

There's quite a lot of manual coding here, and the usage pattern is _ok_, but it could be improved. I put together a [macro library (`context-err`)](https://github.com/coriolinus/context-err) a while ago which is intended to improve the situation, but I don't recommend it for production use yet. Right now the macro is very verbose, but still restrictive: it doesn't allow for boxing or `'static str` contexts. We should think about whether we want to take this more in that direction.

### General-purpose variants which should be collapsed

There are at least a few general-purpose `{context, source}` variants which could usefully be collapsed into more specific variants. I haven't yet done the analysis to identify all of these. In general the rule I will use is, if there are two or fewer distinct contexts in use, then they should just be variants of their own. If there are three or more, we should use the more general form.

### Overall factoring

I've factored down the error types as finely as I think I reasonably can. Usage is relatively straightforward, and we end up with nice error stacks with a lot of context.

The downside of this is that when you actually get into the weeds of it, the error model in this crate is now insanely complex. Lots of simple pieces attaching to each other and converting between each other in non-simple ways. 

I think this is the best we can do for now, but if we decide to do a Big Refactor after the Whiteboard Week in January, then we should definitely refactor and simplify the error model at that time as well. 

----

- The **PR Title**


coriolinus:
Currently there are errors when running `cargo nextest run -p core-crypto --no-fail-fast`:

- `mls::conversation::leaf_node_validation::tests::stages::should_validate_leaf_node_when_adding` (x3): also appears to fail in `main`

~10 other errors. The set of failing cases changes each time I run the test command, and running these cases individually always seems to produce a passing output. So I'm going to mark this down to flakiness and not worry too much about fixing them.

crypto/src/error/legacy.rs
```diff
@@ -27,7 +23,7 @@ use rexie;
 pub enum CryptoError {
```
typfel, line 23: Is this error still in use after the refactoring?

crypto-ffi/src/generic/mod.rs
```diff
@@ -189,52 +192,150 @@ pub enum CoreCryptoError {
...
+                        v
+                    }
+                )*
+                else {
+                    $default
+                }
+            }};
+        }
+
+        match_heterogenous!(innermost => {
```
typfel, line 252: Should we also have `OrphanWelcome` case here?

crypto-ffi/src/wasm/mod.rs
```diff
@@ -196,46 +203,146 @@ pub(crate) enum InternalError {
...
+                        v
+                    }
+                )*
+                else {
+                    $default
+                }
+            }};
+        }
+
+        match_heterogenous!(innermost => {
```
typfel, line 263: Missing `OrphanWelcome` case here?

crypto/src/mls/conversation/commit.rs
```diff
@@ -92,13 +116,24 @@ impl CentralContext {
     /// If the conversation can't be found, an error will be returned. Other errors are originating
     /// from OpenMls and the KeyStore
     #[cfg_attr(test, crate::idempotent)]
-    pub async fn update_keying_material(&self, id: &ConversationId) -> CryptoResult<MlsCommitBundle> {
-        let client = self.mls_client().await?;
+    pub async fn update_keying_material(&self, id: &ConversationId) -> Result<MlsCommitBundle> {
+        let client = self
+            .mls_client()
+            .await
+            .map_err(RecursiveError::root("getting mls client"))?;
```
typfel, line 123: This repeated in quite a few places. Are we ever getting calling `mls_client()` with another intention than "getting mls client"?

typfel: Same question also applies for `mls_provider()`

crypto/src/mls/conversation/renew.rs
```diff
@@ -110,7 +114,7 @@ impl MlsConversation {
                 Proposal::Add(add) => self.propose_add_member(client, backend, add.key_package.into()).await?,
                 Proposal::Remove(remove) => self.propose_remove_member(client, backend, remove.removed()).await?,
                 Proposal::Update(update) => self.renew_update(client, backend, Some(update.leaf_node())).await?,
-                _ => return Err(CryptoError::ImplementationError),
+                _ => return Err(Error::PropopsalVariantCannotBeRenewed),
```
typfel, line 117:
```suggestion
                _ => return Err(Error::ProposalVariantCannotBeRenewed),
```

crypto/src/mls/credential/mod.rs
```diff
@@ -228,17 +230,37 @@ mod tests {
...
+            .unwrap_err();
+        // assert!(innermost_source_matches!(err, Error::InvalidIdentity));
+        //
+        // The above should work, in this case. But it fails
+        // for mysterious reasons that I do not understand.
+        // Digging into it, the `downcast_ref` internal to the `innermost_source_matches`
+        // macro produces `None`, so Rust thinks that it's a different error internally.
+        // But changing the debug/display implementations on `Error::InvalidIdentity`
+        // also changes the implementations on the inner type; I'm reasonably confident
+        // that it is in fact this enum and variaint.
```
typfel, line 244:
```suggestion
        // that it is in fact this enum and variant.
```

crypto/src/mls/credential/mod.rs
```diff
@@ -228,17 +230,37 @@ mod tests {
...
-        assert!(matches!(
-            try_talk(&case, Some(&x509_test_chain), alice_identifier, bob_identifier)
-                .await
-                .unwrap_err(),
-            CryptoError::InvalidIdentity
-        ));
+        let err = try_talk(&case, Some(&x509_test_chain), alice_identifier, bob_identifier)
+            .await
+            .unwrap_err();
+        // assert!(innermost_source_matches!(err, Error::InvalidIdentity));
```
typfel, line 236: This is still a mystery?

crypto/src/mls/error.rs
```diff
@@ -0,0 +1,35 @@
...
+    #[error(
+        "You tried to join with an external commit but did not merge it yet. We will reapply this message for you when you merge your external commit"
+    )]
+    UnmergedPendingGroup,
+    /// The ciphersuite identifier presented does not map to a known ciphersuite.
+    #[error("Unknown ciphersuite")]
+    UnknownCiphersuite,
+    #[error("The callbacks needed for CoreCrypto to operate were not set")]
+    CallbacksNotSet,
+    #[error("External add proposal validation failed: only users already in the group are allowed")]
```
typfel, line 21: These are deleted on main

crypto/src/mls/client/key_package.rs
```diff
@@ -161,10 +162,15 @@ impl Client {
...
         ciphersuite: MlsCiphersuite,
         credential_type: MlsCredentialType,
-    ) -> CryptoResult<usize> {
-        let kps: Vec<MlsKeyPackage> = backend.key_store().find_all(EntityFindParams::default()).await?;
+    ) -> Result<usize> {
+        let kps: Vec<MlsKeyPackage> = backend
+            .key_store()
+            .find_all(EntityFindParams::default())
+            .await
+            .map_err(Error::keystore("finding all key stores"))?;
```
SimonThormeyer: Not sure if I get what you mean, but shouldn't this be "finding all keypackages"?

coriolinus: Yes, agree. Updating.

crypto/src/mls/client/key_package.rs
```diff
@@ -292,25 +307,31 @@ impl Client {
...
-            CryptoResult::Ok(acc)
-        })?;
+        let kps = kps
+            .into_iter()
+            .map(|raw_kp| -> Result<_> {
+                let kp = core_crypto_keystore::deser::<KeyPackage>(&raw_kp.keypackage)
+                    .map_err(Error::keystore("deserializing keypackage"))?;
+                Ok((raw_kp, kp))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
SimonThormeyer: Really like this refactoring, much easier to understand code

crypto/src/mls/conversation/commit.rs
coriolinus: No, but I don't see a better way to structure this.

The `.mls_client()` method (and `mls_provider()` as well) returns a root error: `core_crypto::Error`, implemented in `crypto/src/error/mod.rs`. But we call that method from all over the place, and we need to create all kinds of different errors from that. In this case what the function actually returns is a conversation error: `core_crypto::mls::conversation::Error`, defined in `crypto/src/mls/conversation/error.rs`. 

You can see for yourself what's going on here: `RecursiveError::root` returns a function which wraps the root error and adds the context line, and then the `?` operator uses the `From<RecursiveError>` implementation (from the `#[from]` annotation) to actually convert the error. But the point of the context field is to distinguish the cases. Quite likely, this "getting mls client" error is not the top-level error. It's common to get a MLS client and provider very near each other, and both operations will end up being mapped with `RecursiveError::root`. Having this context allows us to distinguish what in particular is being returned there.

Maybe you're thinking that the answer would be to have the `mls_client` method do its own wrapping: instead of returning `Result<Client, core_crypto::Error>`, it might return `Result<Client, E> where RecursiveError: Into<E>`. That would allow us to inline the context at least, avoid explicitly specifying that we're getting the mls client each time we get the mls client. Unfortunately, there are problems with that approach:

1. It breaks the pattern that each group of module always returns the internal error type for that group of modules. Now we're returning _something_ based on this internal error type, but we can't say exactly what until we know the context of our caller.
2. It forces us to wrap everything everywhere in `RecursiveError`. That makes it more difficult, not less, to determine what the internal error type is.
3. It breaks the `?` operator. The reason for this is that this defines the returned type in terms of `Into`. But the `?` operator also applies an `Into` conversion. So desugaring everything looks like:
    ```rust
    let client = match self.mls_client() {
      Ok(client) => client,
      Err(e) => return e.into(),
    };
    ```
    But Rust can't infer what type `e` should have there, because any number of types might potentially satisfy both `into` conditions. 

Maybe I'm missing something. But I do think that this somewhat-repetitive pattern is the best option we have for now.

(But it's worth mentioning that avoiding this kind of mess is a big driver of my desire to do a big refactoring that we can talk about next month! I don't have the domain knowledge yet to come into that week with a plan for what precisely we should do. But my hope is by the end of the week we will have designed a structure that lets avoid `RecursiveError` entirely.)

crypto-ffi/src/generic/mod.rs
```diff
@@ -189,52 +192,151 @@ pub enum CoreCryptoError {
...
 }
 
 // This implementation is intended to be temporary; we're going to be completely restructuring the way we handle
 // errors in `core-crypto` soon. We can replace this with better error patterns when we do.
 //
 // Certain error mappings could apply to both MLS and Proteus. In all such cases, we map them to the MLS variant.
 // When we redesign the errors in `core-crypto`, these ambiguities should disappear anyway.
-impl From<CryptoError> for CoreCryptoError {
-    fn from(value: CryptoError) -> Self {
+impl From<core_crypto::Error> for CoreCryptoError {
```
SimonThormeyer, line 273: Does the comment above this still apply?

crypto-ffi/src/generic/mod.rs
```diff
@@ -189,52 +192,151 @@ pub enum CoreCryptoError {
...
+            ||=> MlsError::Other(error.innermost_error_message()).into(),
+        })
+    }
 }
 
 // This implementation is intended to be temporary; we're going to be completely restructuring the way we handle
 // errors in `core-crypto` soon. We can replace this with better error patterns when we do.
 //
 // Certain error mappings could apply to both MLS and Proteus. In all such cases, we map them to the MLS variant.
 // When we redesign the errors in `core-crypto`, these ambiguities should disappear anyway.
```
SimonThormeyer, line 272: Does this comment still apply?

crypto-ffi/src/wasm/mod.rs
```diff
@@ -196,46 +203,147 @@ pub(crate) enum InternalError {
...
+            core_crypto::mls::conversation::Error::MessageEpochTooOld => MlsError::MessageEpochTooOld.into(),
+            core_crypto::mls::conversation::Error::SelfCommitIgnored => MlsError::SelfCommitIgnored.into(),
+            core_crypto::mls::conversation::Error::StaleCommit => MlsError::StaleCommit.into(),
+            core_crypto::mls::conversation::Error::StaleProposal => MlsError::StaleProposal.into(),
+            core_crypto::mls::conversation::Error::UnbufferedFarFutureMessage => MlsError::WrongEpoch.into(),
+            core_crypto::mls::conversation::Error::OrphanWelcome => MlsError::OrphanWelcome.into(),
+            core_crypto::mls::Error::UnmergedPendingGroup => MlsError::UnmergedPendingGroup.into(),
+            ||=> MlsError::Other(error.innermost_error_message()).into(),
+        })
+    }
```
SimonThormeyer, line 276: Looks like this entire code block is shared between wasm and uniffi bindings.
Can we de-duplicate it somehow?

crypto/src/e2e_identity/rotate.rs
```diff
@@ -280,24 +343,24 @@ impl MlsRotateBundle {
...
-                    acc.push(kp.tls_serialize_detached().map_err(MlsError::from)?);
-                    CryptoResult::Ok(acc)
-                })?;
+        let new_key_packages = self.new_key_packages.into_iter().try_fold(
+            Vec::with_capacity(kp_size),
+            |mut acc, kp| -> Result<Vec<Vec<u8>>> {
+                acc.push(kp.tls_serialize_detached()?);
+                Ok(acc)
+            },
+        )?;
```
SimonThormeyer, line 363: Just because we're touching this code anyway, if you want, let's refactor this to a `map()`/`collect()

SimonThormeyer (approved):
Amazing, impressive work :)

crypto/src/e2e_identity/rotate.rs
coriolinus: The reason I left these was the `with_capacity` calls. The [`Vec` documentation](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#impl-FromIterator%3CT%3E-for-Vec%3CT%3E) states it clearly:

> Vec may use any or none of the following strategies, depending on the supplied iterator:
> 
> - preallocate based on [Iterator::size_hint()](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.size_hint)

So it's not quite as simple as `map`/`collect`, unfortunately. Needs to be a `with_capacity`/`map`/`extend` setup to retain the existing performance characteristics, and then a wrapper so that we don't leave it mutable after the fact. In the end, I decided that retaining `try_fold` here was the simplest option.

crypto-ffi/src/wasm/mod.rs
coriolinus: I mean, massive parts of the code are shared between wasm and uniffi bindings. The right way to deduplicate this is to unify those bindings and use conditional compilation to expose any differences we require. 

Deduplicating these macros means either putting them in `core-crypto`, or into some other crate which is parent to both `generic` and `wasm`. But that involves making them public. I'm happier if they stay limited to the context of this one very specific function.

crypto/src/e2e_identity/rotate.rs
SimonThormeyer: Good to see a use case where folding to a collection is a valid approach.

crypto/src/mls/credential/mod.rs
coriolinus: Not a mystery anymore, but I'm still figuring out how to solve it. In this particular case, we have `dbg!` output that looks like this:

```
[crypto/src/mls/credential/mod.rs:236:9] &err = Recursive(
    Mls {
        context: "initializing mls",
        source: Recursive(
            MlsClient {
                context: "initializing mls client",
                source: Recursive(
                    MlsCredential {
                        context: "getting client id",
                        source: InvalidIdentity,
                    },
                ),
            },
        ),
    },
)
```

What's going on here is that `RecursiveError` boxes up its sources, so that the size of the enum doesn't go infinite. So even though the representation of the innermost source is just `InvalidIdentity`, what's actually sitting there is a `Box<core_crypto::mls::credential::Error>` with the `InvalidIdentity` variant. 

[`<impl<T> Error for Box<T> where T: Error>::source`](https://doc.rust-lang.org/stable/src/alloc/boxed.rs.html#3028-3030) just defers to the inner `.source()` implementation. So in the case of this error, it returns `None`, because `InvalidIdentity` has no source error. 

The `innermost_source_matches!` macro uses that passed type--in this case, `core_crypto:mls::credential::Error`--to determine what `downcast_ref` should be looking for. And `downcast_ref` decides, correctly, that `core_crypto:mls::credential::Error` and `Box<core_crypto:mls::credential::Error>` are of different types, and refuses to downcast one as the other, which means of course that the match fails.

coriolinus: [Deref patterns](https://github.com/rust-lang/rust/issues/87121) would be really useful here.

crypto/src/mls/conversation/commit.rs
```diff
@@ -296,10 +312,16 @@ impl MlsConversationCreationMessage {
...
-        let welcome = self.welcome.tls_serialize_detached().map_err(MlsError::from)?;
-        let msg = self.commit.tls_serialize_detached().map_err(MlsError::from)?;
+        let welcome = self
+            .welcome
+            .tls_serialize_detached()
+            .map_err(Error::tls_serialize("serialize welcome"))?;
+        let msg = self
+            .commit
+            .tls_serialize_detached()
+            .map_err(Error::tls_serialize("serialize welcome"))?;
```
istankovic, line 369: This should probably be `"serialize commit"`.

crypto/src/mls/credential/mod.rs
coriolinus: fd40ffc6

crypto-ffi/src/generic/mod.rs
coriolinus: Hmm. Not as it was originally written, when it referred to this work. But now I'm hopeful that we can come up with a Big Refactor Plan in a month which leads to us getting better error-handling patterns in the future. So I guess I would still leave it in there for now.

crypto/src/mls/error.rs
coriolinus: Great point: https://github.com/wireapp/core-crypto/pull/791/commits/3ff26921e583e06d2a220ef45870e3221c64147f

istankovic:
I'm wondering if we can find a better name for `LeafError`. Leaf nodes are part of MLS and this error name may cause the reader to think that the error has something to do with those nodes, so it would be nice if we could avoid the ambiguity.

crypto/src/proteus.rs
```diff
@@ -462,10 +470,36 @@ impl ProteusCentral {
...
+        //    it avoids parametric virality. (`init_from_prekey` is itself only generic because it returns
+        //    this error type with a type-parametric variant, which the function never returns.)
+        //
+        // In this case, we could in principle use the unit type, because we never construct the error variant
+        // in question. But of course, that doesn't implement `Display`, which is required for deriving the
+        // wrapper's `Error` implementation.
+        //
+        // So instead we use our own error type, which is recursive and silly, but gets the job done.
+        let proteus_session = Session::init_from_prekey::<Error>(self.proteus_identity.clone(), prekey)
+            .map_err(KeystoreError::wrap("initializing session from prekey"))?;
```
istankovic, line 502: Is this `KeystoreError::wrap` a typo, i.e. shouldn't it be `ProteusError::wrap`?

coriolinus: Oops, you're right: https://github.com/wireapp/core-crypto/pull/791/commits/80ae45aa3fc87e88c9bad593782699737b648538

istankovic:
I would drop the two empty commits:
378e53de095fca72cb52cc05655b626ab3a9191a
4d791f57f67b5591f5d9eb1c483813341e787f9f

to avoid confusing whoever comes across them later.

istankovic (approved):
This was a heroic effort. Well done! :wine_glass:

PR #809, opened 2024-12-09
Ivan/rework js bindings infra

istankovic:


crypto-ffi/bindings/js/wdio.conf.ts
```diff
@@ -1,14 +1,18 @@
...
 
-const STATIC_SEVER_URL = "http://localhost:3000/";
-
-const staticPath = path.resolve(process.cwd(), `../platforms/web`);
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const STATIC_SERVER_URL = "http://localhost:3000/";
 const log = logger("wdio.conf.ts");
 
+let staticPath;
+if (staticPath === undefined) {
```
coriolinus, line 11: This looks like an unconditional branch?

crypto-ffi/bindings/js/wdio.conf.ts
```diff
@@ -1,14 +1,18 @@
...
+    staticPath = await fs.mkdtemp(path.join(tmpdir(), 'core-crypto-wdio-test-'));
```
coriolinus, line 12: Intentional trailing hyphen?

interop/src/clients/corecrypto/ffi.rs
```diff
@@ -29,29 +30,38 @@ use crate::{
 pub(crate) struct CoreCryptoFfiClient {
     cc: CoreCrypto,
     client_id: Vec<u8>,
+    temp_file: NamedTempFile,
```
coriolinus, line 40: Maybe also add an `#[allow(unused)]` here, along with a similar note to the one where this is initialized. 

Alternately, just naming this member `_temp_file` silences the warnings, but also preserves the lifetime so it doesn't drop until the struct does.

coriolinus (commented):
Waiting for CI to go green before approving, but this does look good to me. Hoping to get this merged soon so I can take advantage of it!

crypto-ffi/bindings/js/wdio.conf.ts
istankovic: > Intentional trailing hyphen?

Yes, those typically end up being of the form `core-crypto-wdio-test-XYZABC`, so having the hyphen makes it slightly easier to read, if there ever is a need to interact with those files.

interop/src/clients/corecrypto/ffi.rs
istankovic: > Alternately, just naming this member _temp_file silences the warnings, but also preserves the lifetime so it doesn't drop until the struct does.

Ah, nice, I'm going to use that, thanks!

crypto-ffi/bindings/js/wdio.conf.ts
istankovic: Thanks for asking about that. That was supposed to address the issue where `wdio.conf.ts` is imported by the main wdio process, as well as worker processes, which would then needlessly create multiple temporary directories.

However, as written, that didn't really work so I changed it a bit so we only really create one tempdir per wdio run. I also added a comment explaining what's going on.

coriolinus (approved):

SimonThormeyer (approved):
Amazing work!

PR #812, opened 2024-12-09
fix: wait for current transaction to finish when creating a new one

SimonThormeyer:
Previously, an error was thrown when a new transaction was started and there was one in progress. This was not the intended behavior.

# What's new in this PR
See above

----
##### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #813, opened 2024-12-11
fix: don't swallow transaction errors if they don't originate from the closure [WPB-14895]

typfel:
# What's new in this PR

We were assuming errors when executing a transaction can only come from the client provided closure. This is not true since errors can also come from core crypto, for example when we throw `TransactionInProgress` when there is an attempt to perform parallel transactions.

----

- The **PR Title**


SimonThormeyer:
Will be included in #812

PR #815, opened 2024-12-11
fix: silence verbose logs when performing a transaction [WPB-14953]

typfel:
# What's new in this PR

When creating transaction we also create an in-memory database as cache. When this database is created all the schema migration are performed and logged at the info level and they are very verbose. Silence them by mapping these logs to the Debug level instead.

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/src/generic/mod.rs
+++ b/crypto-ffi/src/generic/mod.rs
@@ -954,9 +954,20 @@ struct CoreCryptoLoggerWrapper {
     logger: std::sync::Arc<dyn CoreCryptoLogger>,
 }
 
+impl CoreCryptoLoggerWrapper {
+    fn adjusted_log_level(&self, metadata: &Metadata) -> Level {
+        match (metadata.level(), metadata.target()) {
+            // increase log level for refinery_core::traits since they are too verbose in transactions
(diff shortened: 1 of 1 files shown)


crypto-ffi/src/generic/mod.rs
```diff
@@ -954,20 +954,38 @@ struct CoreCryptoLoggerWrapper {
...
     }
 
     fn log(&self, record: &Record) {
         let kvs = record.key_values();
         let mut visitor = KeyValueVisitor(BTreeMap::new());
         let _ = kvs.visit(&mut visitor);
 
+        if !self.enabled(record.metadata()) {
+            return;
+        }
```
SimonThormeyer, line 980: Is this really necessary? I would have expected that the call to `enabled()` is done internally in the logger?

SimonThormeyer (approved):
Looks good, I just have a single question.

crypto-ffi/src/generic/mod.rs
typfel: One would think so but the docs say:
> This method isn’t called automatically by the log! macros. It’s up to an implementation of the Log trait to call enabled in its own log method implementation to guarantee that filtering is applied.

coriolinus (approved):

PR #823, opened 2024-12-19
build: use reusable workflows in ci [WPB-14852]

coriolinus:
# What's new in this PR

- [x] ensure docs get built in all cases, but only deployed when `main` does
- [x] identify all cases where multiple workflows do the same things and isolate those in workflows
- [x] use parallel jobs wherever possible

----

- The **PR Title**

### DIFF
--- a/.github/workflows/bindings.yml
+++ b/.github/workflows/bindings.yml
@@ -1,4 +1,4 @@
-name: Build bindings
+name: build bindings
 
 concurrency:
   group: "${{ github.workflow }}-${{ github.ref }}"
@@ -28,14 +28,14 @@ jobs:
     needs: build-android
     steps:
       - uses: actions/checkout@v4
-      - name: Set up JDK 17
+      - name: set up jdk 17
         uses: actions/setup-java@v4
         with:
           java-version: "17"
           distribution: "adopt"
-      - name: Gradle Setup
+      - name: gradle setup
         uses: gradle/actions/setup-gradle@v4
-      - name: Validate Gradle wrapper
+      - name: validate gradle wrapper
         uses: gradle/actions/wrapper-validation@v4
       - uses: actions/download-artifact@v4
         with:
@@ -45,12 +45,12 @@ jobs:
         with:
           name: android-${{ github.run_id }}
           path: crypto-ffi/bindings
-      - name: Enable KVM group perms
+      - name: enable kvm group perms
         run: |
           echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
           sudo udevadm control --reload-rules
           sudo udevadm trigger --name-match=kvm
-      - name: Android Instrumentation Tests
+      - name: android instrumentation tests
         uses: reactivecircus/android-emulator-runner@v2
         with:
           api-level: 32
@@ -64,14 +64,14 @@ jobs:
     needs: build-jvm-linux
     steps:
       - uses: actions/checkout@v4
-      - name: Set up JDK 17
+      - name: set up jdk 17
         uses: actions/setup-java@v4
         with:
           java-version: "17"
           distribution: "adopt"
-      - name: Gradle Setup
+      - name: gradle setup
         uses: gradle/actions/setup-gradle@v4
-      - name: Validate Gradle wrapper
+      - name: validate gradle wrapper
         uses: gradle/actions/wrapper-validation@v4
       - name: download linux library
         uses: actions/download-artifact@v4
@@ -83,7 +83,7 @@ jobs:
         with:
           name: jvm-linux-bindings-${{ github.run_id }}
           path: crypto-ffi/bindings
-      - name: Build and test JVM package
+      - name: build and test jvm package
         run: |
           cd crypto-ffi/bindings
           ./gradlew jvm:build -x lint -x lintRelease
@@ -99,7 +99,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Determine Rust target
+      - name: determine rust target
         id: rust-target
         run: |
           if [[ "${{ matrix.task }}" == "ios-device" ]]; then
@@ -114,15 +114,15 @@ jobs:
         with:
           target: ${{ env.target }}
 
-      - name: Setup cargo-make
+      - name: setup cargo-make
         uses: davidB/rust-cargo-make@v1
 
-      - name: Build ${{ matrix.task }}
+      - name: build ${{ matrix.task }}
         run: |
           cd crypto-ffi
           cargo make ${{ matrix.task }}
 
-      - name: Upload artifact
+      - name: upload artifact
         uses: actions/upload-artifact@v4
         with:
           name: ${{github.event.number}}-${{ matrix.task }}
@@ -130,7 +130,7 @@ jobs:
           retention-days: 1
           overwrite: 'true'
       # Only needs to be uploaded once, this step finishes fastest.
-      - name: Upload FFI artifact
+      - name: upload ffi artifact
         if: startsWith(matrix.task, 'ios-simulator-arm')
         uses: actions/upload-artifact@v4
         with:
@@ -144,18 +144,18 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Download artifacts for ios
+      - name: download artifacts for ios
         uses: actions/download-artifact@v4
         with:
           path: target
           pattern: ${{github.event.number}}-ios-*
           merge-multiple: 'true'
-      - name: Download FFI artifacts
+      - name: download ffi artifacts
         uses: actions/download-artifact@v4
         with:
           name: ${{github.event.number}}-swift-ffi
           path: crypto-ffi/bindings/swift/WireCoreCrypto
-      - name: Create xcframework
+      - name: create xcframework
         run: |
           cd crypto-ffi/bindings/swift
           ./build-xcframework.sh
@@ -181,12 +181,12 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 18
-      - name: Install wasm-pack
+      - name: install wasm-pack
         uses: taiki-e/install-action@v2
         with:
           tool: wasm-pack
       - uses: davidB/rust-cargo-make@v1
-      - name: build WASM
+      - name: build wasm
         run: |
           cd crypto-ffi
           cargo make wasm
@@ -195,15 +195,15 @@ jobs:
           cd crypto-ffi/bindings/js
           bun install
           bun run build
-      - name: Lint
+      - name: lint
         run: |
           cd crypto-ffi/bindings/js
           bun eslint . --max-warnings=0
-      - name: check all TS files
+      - name: check all ts files
         run: |
           cd crypto-ffi/bindings/js
           bun tsc --noEmit
-      - name: Test
+      - name: test
         run: |
           cd crypto-ffi/bindings/js
           bun run test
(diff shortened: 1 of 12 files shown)


coriolinus:
Actions to extract into common workflows:

- [ ] build rust target (?)
- [ ] build wasm target
- [ ] build kotlin/jvm target
- [ ] build android target
- [ ] build ios target
- [ ] rm either `documentation/rust` or `rust/docs` jobs; they do the same thing
- [ ] parallelize `rust/check` jobs with a matrix
- [ ] rm `rust/build` job as it provides no particular benefits or artifacts
- [ ] rm `rust/wasm-build` job as it provides no particular benefits or artifacts

coriolinus:
Some findings so far:

- bun keeps some kind of state outside the local directory and its contained `node_modules`. The implication of which is that packing the entire directory up as an artifact and then unpacking it later does not actually work.
- As a consequence, we don't try to do anything with `bun` directly downstream of the unarchive operation. Instead, we use `bunx`, which fetches necessary resources dynamically.
- `bunx` is not smart about following versioning information in a `package.json` file, so we need to make versioning explicit any time we don't want `latest`.
- `bunx` and `wdio` don't play nicely together.

crypto-ffi/bindings/js/tsconfig.json
```diff
@@ -18,11 +18,20 @@
...
-      "node_modules"
+    "node_modules",
+    "*.conf.ts",
+    "test",
+    "*.test.ts",
+    "benches",
+    "*.bench.ts"
   ]
-}
+}
```
istankovic, line 37: Why these changes?

istankovic:
Commit `5c0d6b6f104856cdbad0183fc71485d78fd7f936` has `wip` in its message -- is it still work in progress?

crypto-ffi/bindings/js/tsconfig.json
coriolinus: I'm 99% sure that's where there used to be trailing whitespace which got auto-stripped by my editor.

istankovic: Sorry, I mean the whole set of changes to `tsconfig.json`, not jus the newline.

coriolinus:
> Commit `5c0d6b6f104856cdbad0183fc71485d78fd7f936` has `wip` in its message -- is it still work in progress?

No, and I've adjusted it.

crypto-ffi/bindings/js/tsconfig.json
coriolinus: Ah. That's in two parts. The formatting changes were autoformat on the JSON, were automatic on the part of the editor, and are irrelevant. I can try to put them back if you think it's important.

Excluding the other files may not be necessary anymore. At the time, TS was complaining quite a lot about the contents of various test, bench, and config files that it could not figure out. Those seemed irrelevant, so I excluded them from consideration. I think you may have done some work improving that situation since then, though.

istankovic: > Ah. That's in two parts. The formatting changes were autoformat on the JSON, were automatic on the part of the editor, and are irrelevant. I can try to put them back if you think it's important.

Nah, no need. 

> Excluding the other files may not be necessary anymore. At the time, TS was complaining quite a lot about the contents of various test, bench, and config files that it could not figure out. Those seemed irrelevant, so I excluded them from consideration. I think you may have done some work improving that situation since then, though.

You could say I spent the last several weeks working on resolving _just_ those issues, which eventually resulted in the rework of the JS wrapper. :-)

This is why it's confusing to me to see these changes, as I expect no errors anymore.
(There should be ~7 doc warnings, but those should not fail CI for the time being and we have an item to address them.)

.github/workflows/bindings.yml
```diff
@@ -175,31 +172,38 @@ jobs:
...
-
+      - name: Setup bun
+        run: |
+          cd crypto-ffi/bindings/js
+          bun install
           bun run build
+      - name: Lint
+        run: |
+          cd crypto-ffi/bindings/js
+          bun eslint .. --max-warnings=0
```
istankovic, line 201: Shouldn't this be `bun eslint . --max-warnings=0`?
Note `.` instead of `..`.
The command is supposed to run on `bindings/js`, not its parent.

crypto-ffi/bindings/js/tsconfig.json
coriolinus: Fair enough. I've excluded that file from its commit.

.github/workflows/bindings.yml
coriolinus: Oops, yes! Fixed.

.github/workflows/build-android.yml
```diff
@@ -0,0 +1,60 @@
+name: Build Rust
```
istankovic, line 1: Shouldn't this be something like "Build Android"?

coriolinus: I suppose. The top-level name gets stripped out anyway in favor of the calling job's name, so I don't think it matters very much.

coriolinus: This particular top-level name gets stripped out because it is part of a reusable workflow.

.github/workflows/build-android.yml
```diff
@@ -0,0 +1,60 @@
...
+          ./gradlew android:build -x lint -x lintRelease
+      - uses: actions/upload-artifact@v4
+        with:
+          name: android-target-${{ github.run_id }}
+          path: target/*android*
+          retention-days: 1
+      - uses: actions/upload-artifact@v4
+        with:
+          name: android-${{ github.run_id }}
+          path: crypto-ffi/bindings
```
istankovic, line 59: Why not upload `crypto-ffi/bindings/android`?

.github/workflows/docs.yml
```diff
@@ -40,20 +62,72 @@ jobs:
...
+      - rustdoc
+      - tsdoc
+      - ktdoc
+
+    steps:
+    - uses: actions/download-artifact@v4.1.8
+      with:
+        path: target/doc
+        merge-multiple: true
+
```
istankovic, line 126: This is really nice! :+1:

.github/workflows/docs.yml
```diff
@@ -40,20 +62,72 @@ jobs:
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - uses: actions-rust-lang/setup-rust-toolchain@v1  # this implicitly caches Rust tools and build artifacts
+      with:
+        rustflags: ''
+        target: |
+          x86_64-unknown-linux-gnu
```
istankovic, line 92: I don't think specifying the target here is necessary, one only needs that for additional, non-system targets.

.github/workflows/build-android.yml
coriolinus: I ~~couldn't~~ didn't want to take the time to prove that it didn't require other components, like `uniffi-android`. Total size of this artifact is ~8Mb, which should be fine, I think.

.github/workflows/docs.yml
coriolinus: Thanks!

.github/workflows/bindings.yml
```diff
@@ -175,31 +175,38 @@ jobs:
...
         run: |
           cd crypto-ffi
           cargo make wasm
-          cd bindings/js
-          bun eslint . --max-warnings=0
-
-          # check all Typescript files (wdio.conf.ts, benches)
-          bunx tsc --noEmit
-
+      - name: Setup bun
```
istankovic, line 193: This should be named something like "Compile TS files", because bun is already set up by the time we are here. Here we are just installing dependencies and running the compiler + bundler.

.github/workflows/bindings.yml
```diff
@@ -175,31 +175,38 @@ jobs:
         id: setup-chrome
         with:
           chrome-version: stable
-      - uses: actions-rust-lang/setup-rust-toolchain@v1  # this implicitly caches Rust tools and build artifacts
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          rustflags: ''
```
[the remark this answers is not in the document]
coriolinus: If everything is passing right now (as it currently is), then we've cleaned up the code to the point where we are not raising any (denied) warnings anymore. I think this is a good thing!

.github/workflows/bindings.yml
coriolinus: Sure, will adjust.

.github/workflows/rust.yml
```diff
@@ -163,24 +163,23 @@ jobs:
...
+      - name: setup wasm-pack
         uses: taiki-e/install-action@v2
         with:
           tool: wasm-pack
       - uses: browser-actions/setup-chrome@latest
         id: setup-chrome
         with:
           chrome-version: stable
+      - run: |
+          echo "CHROME_PATH=${{ steps.setup-chrome.outputs.chrome-path }}" >> $GITHUB_ENV
```
istankovic, line 176: Why is this now necessary?

.github/workflows/build-android.yml
istankovic: Oof, I didn't see `uniffi-android`, good point. :+1:

.github/workflows/build-android.yml
coriolinus: Turns out we don't! So I've removed it.

.github/workflows/bindings.yml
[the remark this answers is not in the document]
istankovic: Absolutely! I was just surprised to see this, wasn't expecting us to reach that state so soon. :sweat_smile:

.github/workflows/build-jvm-linux.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+            name: jvm-linux-so-file-${{ github.run_id }}
+            path: target/x86_64-unknown-linux-gnu/release/*.so
+            retention-days: 1
+      - name: upload bindings
+        uses: actions/upload-artifact@v4
+        with:
+          name: jvm-linux-bindings-${{ github.run_id }}
+          path: crypto-ffi/bindings
+          retention-days: 1
```
istankovic, line 38: Why not invoke `upload-artifact` action just once, with multiple paths, so you get both native libraries and bindings in the same artifact?

.github/workflows/bindings.yml
```diff
@@ -69,14 +73,16 @@ jobs:
...
-      - name: Build JVM rust binaries
-        run: |
-          cd crypto-ffi
-          cargo make jvm-linux
+      - name: download linux library
+        uses: actions/download-artifact@v4
+        with:
+          name: jvm-linux-so-file-${{ github.run_id }}
+          path: target/x86_64-unknown-linux-gnu/release
+      - name: download linux bindings
```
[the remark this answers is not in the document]
coriolinus: Here you go: https://github.com/wireapp/core-crypto/pull/823/commits/6f2ad33fb3b87e5c36bb3175bea83f776e52b730

.github/workflows/docs.yml
istankovic: My reasoning here is, we should avoid unnecessary things like this, where we know omitting them would do just as well. Contrast that with `rustflags: ''` where we actually want to change the value.

coriolinus: Ok, adjusted it.

.github/workflows/docs.yml
```diff
@@ -32,29 +50,79 @@ jobs:
...
+    needs:
+      - rustdoc
+      - tsdoc
+      - ktdoc
+
+    steps:
+    - uses: actions/download-artifact@v4.1.8
+      with:
+        path: target/doc
+        merge-multiple: true
```
istankovic, line 123: Without an artifact name, wouldn't this download and merge all artifacts for the run?

coriolinus: Oh, good point! Fixing.

coriolinus: https://github.com/wireapp/core-crypto/pull/823/commits/73f1f2b5ff7ea02cf5c0c731df50040aef8371a6

PR #825, opened 2024-12-24
feat: commits producing apis use transport [WPB-12121]

SimonThormeyer:
# What's new in this PR

See Jira item and commit messages

----

- The **PR Title**


crypto-ffi/bindings/js/src/CoreCrypto.ts
```diff
@@ -1150,8 +1128,28 @@ export class CoreCrypto {
...
+                                commit: commitBundle.commit,
+                                groupInfo: {
+                                    encryptionType:
+                                        commitBundle.group_info.encryption_type,
+                                    ratchetTreeType:
+                                        commitBundle.group_info
+                                            .ratchet_tree_type,
+                                    payload: commitBundle.group_info.payload,
+                                },
+                            }
```
typfel, line 1144: Why can't we pass the commitBundle directly? Mismatching type? Shouldn't we also have ffi mapping function here as well then?

crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/MLSTest.kt
```diff
@@ -387,3 +391,30 @@ fun randomIdentifier(n: Int = 12): String {
         .map { kotlin.random.Random.nextInt(0, charPool.size).let { charPool[it] } }
         .joinToString("")
 }
+
+interface DeliveryService : MlsTransport {
+    suspend fun getLatestCommitBundle(): CommitBundle
+
+    suspend fun getLatestWelcome(): Welcome
+
+    suspend fun getLatestCommit(): MlsMessage
```
typfel, line 400: Shouldn't these all by optional?

crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/MLSTest.kt
```diff
@@ -387,3 +391,30 @@ fun randomIdentifier(n: Int = 12): String {
...
+}
+
+class  MlsTransportSuccessProvider : DeliveryService {
```
typfel, line 403:
```suggestion
class  MockMlsTransportSuccessProvider : DeliveryService {
```

Let's make it clear that this is mock implementation which we are using in the tests.

crypto-ffi/src/generic/context/mod.rs
```diff
@@ -414,7 +413,7 @@ impl CoreCryptoContext {
         &self,
         conversation_id: Vec<u8>,
         key_packages: Vec<Vec<u8>>,
-    ) -> CoreCryptoResult<MemberAddedMessages> {
+    ) -> CoreCryptoResult<Option<Vec<String>>> {
```
typfel, line 416: Does it make sense to introduce a typedef or something to indicate what the list of string that's optionally returned is?

crypto-ffi/src/generic/mod.rs
```diff
@@ -117,6 +117,16 @@ pub enum MlsError {
     /// with an External Commit instead
     #[error("Although this Welcome seems valid, the local KeyPackage it references has already been deleted locally. Join this group with an external commit")]
     OrphanWelcome,
+    /// This error is returned based on the abort reason returned in `MlsTransportResponse`. It is constructed when the reason is equal to `mls-client-mismatch`.
+    /// You should catch this error and restart the operation.
+    #[error("The message was rejected by the delivery service because there was a mismatch in mls clients. Restart the operation.")]
+    RestartOperation,
```
typfel, line 123:
```suggestion
    RetryOperation,
```

I'm not sure we should make decision for the client. Maybe it's better to keep as `OperationAborted` and let clients decide if they want to retry the operation or not.

crypto-ffi/src/generic/mod.rs
```diff
@@ -117,6 +117,16 @@ pub enum MlsError {
     /// with an External Commit instead
     #[error("Although this Welcome seems valid, the local KeyPackage it references has already been deleted locally. Join this group with an external commit")]
     OrphanWelcome,
+    /// This error is returned based on the abort reason returned in `MlsTransportResponse`. It is constructed when the reason is equal to `mls-client-mismatch`.
```
typfel, line 120: Why don't make this decision though, the clients parse the backend error `mls-client-mismatch` and tells us to abort the operation? In theory there could be other cases where a client decides to abort an operation.

typfel: Like for example because they reached the max amounts of retries.

crypto/src/mls/conversation/commit.rs
```diff
@@ -20,11 +21,70 @@ use crate::{
...
+                .send_commit_bundle(commit.clone())
+                .await
+                .map_err(RecursiveError::root("sending commit bundle"))?
+            {
+                MlsTransportResponse::Success => {
+                    return Ok(());
+                }
+                MlsTransportResponse::Abort { reason } if reason == MLS_CLIENT_MISMATCH_REASON => {
+                    return Err(Error::RestartOperation);
+                }
```
typfel, line 79: I'm not sure it's good idea do this? Now we rely on what clients puts in the reason string. 

IMO clients do this comparison on their own and decide if they want to restart the operation or not.

crypto-ffi/src/generic/mod.rs
SimonThormeyer: I was unsure about this, I'll just drop the corresponding commit and the extra error variant.

crypto/src/mls/conversation/commit.rs
SimonThormeyer: will just drop the commit, see above

crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/MLSTest.kt
SimonThormeyer: The functions are only intended to be called when a commit has already been sent. So I decided to make a non-null assertion and have the return type non-optional, to make test code a bit more concise (I did the same in rust tests, to avoid having to call `unwrap()` in tests everywhere.

I'm up for changing this if you think it would be better to keep make the (in case of unintended usage) possibilty of the buffered commit bundle being empty explicit

typfel: OK I see, yeah I guess it's nicer like this and not unwrap in all the tests.

typfel (approved):

PR #832, opened 2025-01-03
feat!: proposal producing APIs use mls transport [WPB-12135]

SimonThormeyer:
# What's new in this PR
See Jira item

----

- The **PR Title**


crypto/src/mls/conversation/proposal.rs
```diff
@@ -17,6 +17,78 @@ use crate::{
...
+                proposal
+                    .to_bytes()
+                    .map_err(MlsError::wrap("constructing byte vector of proposal"))?,
+            )
+            .await
+            .map_err(RecursiveError::root("sending mls message"))?
+        {
+            MlsTransportResponse::Success => Ok(()),
+            MlsTransportResponse::Abort { reason } => {
+                self.clear_all_pending_proposals(conversation).await?;
```
istankovic, line 45: Assume we're inside a client transaction where two proposals are sent for the same conversation.
Further assume the transport response for the first one returned success, while the second one resulted in an `Abort` here. Would the first proposal will still be in the pending set and would clearing all pending proposals also remove it too?

crypto/src/mls/proposal.rs
```diff
@@ -153,6 +153,8 @@ impl CentralContext {
                     .map_err(RecursiveError::root("getting mls provider"))?,
                 conversation.write().await,
             )
+            .await?;
+        self.send_and_merge_or_discard_proposal(conversation.deref_mut(), proposal)
```
istankovic, line 145: Wouldn't `&mut conversation` work here, instead of calling `deref_mut`?

crypto/src/mls/proposal.rs
```diff
@@ -107,20 +88,26 @@ impl MlsProposal {
...
-    pub async fn new_add_proposal(&self, id: &ConversationId, key_package: KeyPackage) -> Result<MlsProposalBundle> {
+    pub async fn new_add_proposal(
+        &self,
+        id: &ConversationId,
+        key_package: KeyPackage,
+    ) -> Result<NewCrlDistributionPoint> {
         self.new_proposal(id, MlsProposal::Add(key_package)).await
     }
 
     /// Creates a new Add proposal
```
istankovic, line 99: This comment is wrong.

crypto/src/mls/proposal.rs
```diff
@@ -144,18 +131,21 @@ impl CentralContext {
...
+        let mut conversation = conversation.write().await;
+        let (proposal, new_crl_distribution_point) = proposal
             .create(
                 client,
                 &self
                     .mls_provider()
                     .await
                     .map_err(RecursiveError::root("getting mls provider"))?,
-                conversation.write().await,
+                conversation.deref_mut(),
```
istankovic, line 142: Why not just `&mut conversation`?

crypto/src/mls/conversation/merge.rs
```diff
@@ -273,107 +218,30 @@ mod tests {
...
-                    alice_central
-                        .context
-                        .new_conversation(&id, case.credential_type, case.cfg.clone())
-                        .await
-                        .unwrap();
-                    assert!(alice_central.pending_proposals(&id).await.is_empty());
-                    let any_ref = MlsProposalRef::from(vec![0; case.ciphersuite().hash_length()]);
-                    let clear = alice_central.context.clear_pending_proposal(&id, any_ref.clone()).await;
-                    assert!(
-                        matches!(clear.unwrap_err(), Error::PendingProposalNotFound(prop_ref) if prop_ref == any_ref)
```
istankovic, line 370: I'm not sure if I understand the reason behind removing the `PendingProposalNotFound` enum variant and this test. Could you please provide more details on that?

I'm trying to imagine if there is a case where we get a proposal ref from another client and then try to find it locally.

crypto-ffi/src/generic/mod.rs
```diff
@@ -681,7 +678,14 @@ impl TryFrom<MlsConversationDecryptMessage> for DecryptedMessage {
...
             .proposals
             .into_iter()
-            .map(ProposalBundle::try_from)
+            .map(|mls_message| {
+                mls_message
+                    .to_bytes()
+                    .map(Proposal::from)
+                    .map_err(core_crypto::MlsError::wrap("converting mls message to bytes"))
+                    .map_err(core_crypto::Error::from)
+                    .map_err(CoreCryptoError::from)
```
istankovic, line 687: This looks funny. :-)

crypto/src/mls/conversation/proposal.rs
```diff
@@ -17,6 +17,78 @@ use crate::{
...
+                Err(Error::MessageRejected { reason })
+            }
+            MlsTransportResponse::Retry => self.commit_pending_proposals(conversation.id()).await,
```
istankovic, line 48: Why do we commit pending proposals when we get a `Retry`? Shouldn't that be done on `Success`?
I note the Jira item also has this in the pseudocode.

istankovic (changes requested):
This needs more work, in particular regarding the `Retry` handling, as returned by the clients.

SimonThormeyer:
replaced by
https://github.com/wireapp/core-crypto/pull/845

PR #848, opened 2025-01-13
refactor: rework JS wrapper, phase 1 [WPB-15318]

istankovic:
This PR breaks the import cycle in the wrapper by splitting the code into more modules and addresses several issues, see commit messages for details.

The rest of the work on the wrapper will come in a subsequent PR.


crypto-ffi/bindings/js/src/CoreCryptoCommon.ts
coriolinus, line 1: Any particular reason not to call this `CoreCryptoError.ts`? That's what it handles in practice.

crypto-ffi/bindings/js/src/CoreCrypto.ts
```diff
@@ -14,23 +14,14 @@
...
-    NewAcmeAuthz,
-    NewAcmeOrder,
-    MlsTransportProvider,
-} from "./core-crypto-ffi.js";
+export * from "./CoreCryptoCommon.js";
+export * from "./CoreCryptoContext.js";
+export * from "./CoreCryptoInstance.js";
+export * from "./CoreCryptoMLS.js";
+export * from "./CoreCryptoE2EI.js";
+export * from "./CoreCryptoProteus.js";
```
coriolinus, line 22: I know it's a lot of extra verbiage, but I do prefer to avoid glob imports (or in this case, exports). I've had TS editors which jump to the import line on the "jump to definition" action. When you have an explicit import, you can just press the button again to jump into the source file. With a glob, you have to go there manually according to what line the cursor it at after that action.

And of course, outside a language-server-enabled editor, you'd need explicit imports to have any hope of tracing back the precise definition site at all.

coriolinus (approved):
Approved with comments.

crypto-ffi/bindings/js/src/CoreCryptoCommon.ts
istankovic: Sure, I think previously it contained some more code, but now it just contains error-related code. :+1:

coriolinus (approved):
Looks good! (Pending CI going green, of course.)

PR #850, opened 2025-01-14
feat: iOS interop [WPB-15034]

typfel:
# What's new in this PR

Adds an iOS client to the e2e interop tests

----

- The **PR Title**

### DIFF
--- a/extras/webdriver-installation/src/webdriver.rs
+++ b/extras/webdriver-installation/src/webdriver.rs
@@ -222,7 +222,7 @@ impl WebdriverKind {
                     .take()
                     .unwrap()
                     .channels
-                    .remove(&ChromeDriverReleaseDetailsChannel::Stable)
+                    .remove(&ChromeDriverReleaseDetailsChannel::Canary)
                     .unwrap();
 
                 let download_url = channel
--- a/.github/workflows/bindings.yml
+++ b/.github/workflows/bindings.yml
@@ -133,7 +133,7 @@ jobs:
       - uses: browser-actions/setup-chrome@latest
         id: setup-chrome
         with:
-          chrome-version: stable
+          chrome-version: latest
       - run: |
           echo "CHROME_PATH=${{ steps.setup-chrome.outputs.chrome-path }}" >> $GITHUB_ENV
       - uses: oven-sh/setup-bun@v2


interop/src/clients/corecrypto/ios.rs
```diff
@@ -0,0 +1,373 @@
...
+#[derive(Error, Debug)]
+struct SimulatorDriverError {
+    msg: String,
+}
+
+impl Display for SimulatorDriverError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "simulator driver error: {}", self.msg)
+    }
+}
```
SimonThormeyer, line 57: You can use the `thiserror` crate to implement `Display`:

```suggestion
   #[error("simulator driver error: {msg}")]
    msg: String,
}
```

SimonThormeyer (approved):
Approved with a single comment

extras/webdriver-installation/src/webdriver.rs
```diff
@@ -222,7 +222,7 @@ impl WebdriverKind {
                     .take()
                     .unwrap()
                     .channels
-                    .remove(&ChromeDriverReleaseDetailsChannel::Stable)
+                    .remove(&ChromeDriverReleaseDetailsChannel::Canary)
```
istankovic, line 225: What was the reason for this change?

typfel: The web driver was crashing after switching to run a macOS runner, upgrading to the latest Chrome resolves the crash.

istankovic:
Could you please fold the fixup commit into the commit introducing the change the fixup commit is touching?

typfel:
> Could you please fold the fixup commit into the commit introducing the change the fixup commit is touching?

Of course, I thought it's nicer to leave them as fixups while reviewing so you can see the only the relevant changes I make.

interop/src/clients/InteropClient/InteropClient.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,390 @@
...
+			runOnlyForDeploymentPostprocessing = 0;
+		};
+/* End PBXCopyFilesBuildPhase section */
+
+/* Begin PBXFileReference section */
+		16029C902D371F0D007B20A0 /* WireCoreCrypto.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; path = WireCoreCrypto.framework; sourceTree = BUILT_PRODUCTS_DIR; };
+		166BD2BC2D28123F00E9FABB /* InteropClient.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = InteropClient.app; sourceTree = BUILT_PRODUCTS_DIR; };
+		166BD2E12D30296B00E9FABB /* WireCoreCrypto.xcodeproj */ = {isa = PBXFileReference; lastKnownFileType = "wrapper.pb-project"; name = WireCoreCrypto.xcodeproj; path = "../../../../crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto.xcodeproj"; sourceTree = SOURCE_ROOT; };
+/* End PBXFileReference section */
+
```
istankovic, line 33: It should be possible to script creation of these auto-generated project files, right?
Why not use something like https://github.com/yonaskolb/XcodeGen and avoid checking in auto-generated files into the repo?

extras/webdriver-installation/src/webdriver.rs
istankovic: That would have been a perfect commit message for a commit with just that change. ;-)

interop/src/clients/InteropClient/InteropClient.xcodeproj/project.pbxproj
typfel: Yep, that would probably work but that's another dependency on a third party tool. In this case I'm not sure if it's worth since I suspect we'll rarely change the project files.

I'm hoping soon it will be supported to build iOS app using the swift package manager.

extras/webdriver-installation/src/webdriver.rs
typfel: True, I'll extract it

interop/Cargo.toml
```diff
@@ -38,6 +38,8 @@ tempfile = { version = "3.15" }
 # Terminal support
 xshell = "0.2"
 spinoff = { version = "0.8", features = ["aesthetic"], default-features = false }
+serde = { version = "1.0.217", features = ["derive"] }
```
istankovic, line 41: Why are we using this specific version of `serde`, and not just using the one from workspace?

typfel: No reason, was added by the IDE

extras/webdriver-installation/src/webdriver.rs
typfel: I'll also see if I can use something less bleeding edge than Canary

interop/src/clients/corecrypto/ios.rs
```diff
@@ -0,0 +1,367 @@
...
+
+        Ok(Self {
+            driver,
+            client_id: client_id.into_bytes().into(),
+            #[cfg(feature = "proteus")]
+            prekey_last_id: Cell::new(0),
+        })
+    }
+
+    #[allow(dead_code)]
```
istankovic, line 163: Why is this necessary?

typfel: It isn't. Was implemented for the web client but we can actually delete it there as well.

interop/src/clients/InteropClient/InteropClient/InteropAction.swift
```diff
@@ -0,0 +1,259 @@
...
+            let ciphersuite = components?.queryItems?.first(where: {
+                $0.name == "ciphersuite"
+            })?.value.flatMap { UInt16($0) }
+            
+            if let ciphersuite  {
+                self = .getKeyPackage(ciphersuite: ciphersuite)
+            } else {
+                return nil
+            }
+        case "add-client":
```
istankovic, line 73: There seems to be inconsistent formatting here -- no blank line between case blocks, whereas in blocks below there is. Do we have a swift formatter/linter checking this on CI?

typfel: No we don't

istankovic:
Regarding commit sequencing, I think it would be better if we modify `b7197602d2da263b038dbe9987e7ee26d2f07a4e` (ci: build and install ios interop client in interop workflow) so that the changes touching `.github/workflows/rust.yml` are moved to the next commit, `04b3f1a88b62261d38043f344551127f93025876` (ci: re-use job for building ios artifacts).

That way, the first one would just add the `grant_permissions.sh` and `install_app.sh` scripts, and the second one would actually make use of those scripts in the CI.

interop/src/clients/InteropClient/InteropClient/InteropAction.swift
typfel: But good idea, I'll add ticket in JIRA

interop/src/clients/corecrypto/ios.rs
typfel: Deleted

.github/workflows/build-ios.yml
```diff
@@ -0,0 +1,63 @@
...
+          name: ${{github.event.number}}-${{ matrix.task }}
+          path: target
+          retention-days: 1
+          overwrite: 'true'
+      # Only needs to be uploaded once, this step finishes fastest.
+      - name: upload ffi artifact
+        if: startsWith(matrix.task, 'ios-simulator-arm')
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{github.event.number}}-swift-ffi
```
istankovic, line 60: Here we're using `github.event.number`, elsewhere we're using `github.run_id` to uniquely name artifacts. I think we should be consistent, and use `github.run_id` everywhere, unless there is a good reason to prefer `github.event.number` here.

interop/src/clients/corecrypto/ios.rs
```diff
@@ -0,0 +1,354 @@
...
+use std::process::{Child, ChildStdout, Command, Stdio};
+use std::rc::Rc;
+use thiserror::Error;
+use tls_codec::Deserialize;
+
+#[derive(Debug)]
+struct SimulatorDriver {
+    device: String,
+    process: Child,
+    output: Rc<RefCell<BufReader<ChildStdout>>>,
```
istankovic, line 36: Why do we need `Rc` here?

.github/workflows/build-ios.yml
typfel: I just moved the job into separate file maybe @coriolinus can give a reason.

coriolinus: I didn't touch that actually; it appears to come from https://github.com/wireapp/core-crypto/commit/1ce7d95b3e2267e08b30766644a58eadee349d30, which is @SimonThormeyer. 

I don't think there's a strong reason here to prefer either one here. TBH we'd probably be fine without a distinguishing number in either case; looking deeper into the download artifact action, it automatically partitions by run id anyway. Adding an event number or run id just helps us keep track of things if we happen to download some artifacts for manual inspection.

istankovic: > looking deeper into the download artifact action, it automatically partitions by run id anyway

Ah, didn't know that, thanks!

istankovic:
@typfel what went wrong [here](https://github.com/wireapp/core-crypto/actions/runs/12827927691/job/35772215680?pr=850)? Was that some sort of github CI timeout?

typfel:
> @typfel what went wrong [here](https://github.com/wireapp/core-crypto/actions/runs/12827927691/job/35772215680?pr=850)? Was that some sort of github CI timeout?

I was testing another chrome version. I guess it was unsuccessful

typfel:
> > > @typfel what went wrong [here](https://github.com/wireapp/core-crypto/actions/runs/12827927691/job/35772215680?pr=850)? Was that some sort of github CI timeout?
> > 
> > 
> > I was testing another chrome version. I guess it was unsuccessful
> 
> Was there a successful run of `build bindings / e2e-interop-test (pull_request) `? In the above run I can see the test took too long, as if it was stuck during migration. I assume that doesn't happen locally on your machine, does it?

I think the migration are unrelated that's coming from the the ffi client but afterwards the web client is initialised and that got stuck apparently.

istankovic (approved):
Looks good to me. :+1: 

I'm curious about the total CI time now, as github sometimes really takes a long time to give a macos runner.

typfel:
> Looks good to me. 👍
> 
> I'm curious about the total CI time now, as github sometimes really takes a long time to give a macos runner.

It can a really long time

PR #853, opened 2025-01-15
refactor: rework JS wrapper, phase 2 [WPB-15318]

istankovic:
This PR is a continuation of the work done in https://github.com/wireapp/core-crypto/pull/848.

It resolves the remaining warnings about conflicting Typescript types and does it by eliminating wrapper types and instead only exposing types defined on the Rust level.

### DIFF
--- a/crypto-ffi/src/wasm/mod.rs
+++ b/crypto-ffi/src/wasm/mod.rs
@@ -1128,8 +1128,14 @@ impl ConversationConfiguration {
 #[derive(Debug, Copy, Clone, serde::Serialize, serde::Deserialize)]
 /// see [core_crypto::prelude::MlsCustomConfiguration]
 pub struct CustomConfiguration {
-    key_rotation_span: Option<u32>,
-    wire_policy: Option<WirePolicy>,
+    ///  Duration in seconds after which we will automatically force a self-update commit
+    ///  Note: This isn't currently implemented
+    #[wasm_bindgen(js_name = keyRotationSpan)]
+    pub key_rotation_span: Option<u32>,
+    /// Defines if handshake messages are encrypted or not
+    /// Note: encrypted handshake messages are not supported by wire-server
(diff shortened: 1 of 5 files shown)


crypto-ffi/src/wasm/mod.rs
```diff
@@ -1082,11 +1082,16 @@ impl From<core_crypto::e2e_identity::E2eiDumpedPkiEnv> for E2eiDumpedPkiEnv {
...
 #[wasm_bindgen]
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+/// Configuration object for new conversations
 /// see [core_crypto::prelude::MlsConversationConfiguration]
 pub struct ConversationConfiguration {
-    ciphersuite: Option<Ciphersuite>,
+    #[wasm_bindgen(readonly)]
+    /// Conversation ciphersuite
+    pub ciphersuite: Option<Ciphersuite>,
     external_senders: Vec<Vec<u8>>,
```
coriolinus, line 1091: `external_senders` is not readonly?

coriolinus: I do see the other `externalSenders` getter implementation, but I would expect that right now the JS version of a conversation configuration has a read-only `externalSenders` getter and a writeable `external_senders` field.

istankovic: wasm-bindgen only generates getters and setters for `pub` items. Before this PR, we had an interface that was used to provide docs, and not much more than that.

istankovic: Note that there is a getter in the `impl` block, so we still have a getter, but not a setter -- essentialy what `readonly` would give you.

coriolinus: Ah, ok. You're right, I missed that it wasn't `pub`.

coriolinus (approved):

PR #858, opened 2025-01-16
feat: implement basic derive macro for entity trait [WPB-14952]

SimonThormeyer:
# What's new in this PR
see jira item

----

- The **PR Title**


crypto-attributes/src/entity_derive.rs
```diff
@@ -0,0 +1,567 @@
...
+
+        let blob_columns = self
+            .columns
+            .iter()
+            .filter_map(|column| {
+                if matches!(column.column_type, ColumnType::Bytes) {
+                    Some(column.name.clone())
+                } else {
+                    None
+                }
```
coriolinus: Could be

```rust
(column.column_type == ColumnType::Bytes).then(|| column.name.clone())
```

I don't expect the performance impact of deriving `PartialEq, Eq` for `ColumnType` to be severe.

crypto-attributes/src/entity_derive.rs
```diff
@@ -0,0 +1,567 @@
...
+            ..
+        } = self;
+
+        let string_id_conversion = matches!(id_type, ColumnType::String).then(|| {
+            quote! { let #id: String = id.try_into()?; }
+        });
+
+        let id_getter = match id_type {
+            ColumnType::String => quote! {self.#id.as_bytes() },
+            ColumnType::Bytes => quote! { &self.#id[..] },
```
coriolinus: It's equivalent, but IMO `self.#id.as_slice()` would be better. It introduces a nice parallel with the string case, and is easier to understand for people unfamiliar with the slice-everything idiom.

crypto-attributes/src/entity_derive.rs
```diff
@@ -0,0 +1,567 @@
...
+                }
+
+                async fn find_all(
+                    conn: &mut Self::ConnectionType,
+                    params: crate::entities::EntityFindParams,
+                ) -> crate::CryptoKeystoreResult<Vec<Self>> {
+                    let transaction = conn.transaction()?;
+                    let query: String = format!("SELECT rowid, {} FROM {} {}", #id_name, #collection_name, params.to_sql());
+
+                    let mut stmt = transaction.prepare_cached(&query)?;
```
coriolinus: Definitely a code smell to directly inject values into a SQL string and then execute it, though I suppose we can't really help it in the case of `params` if `to_sql` produces a `WHERE` clause. 

It's particularly odd here though because we have the option to bake `#id_name` and `#collection_name` into the query directly.

crypto-attributes/src/entity_derive.rs
```diff
@@ -0,0 +1,567 @@
...
+                    let mut rows = stmt.query_map([], |r| Ok((r.get(0)?, r.get(1)?)))?;
+                    let entities = rows.try_fold(Vec::new(), |mut acc, row_res| {
```
coriolinus: Why not write this as `rows.map(...).collect::<Result<Vec<_>, _>>()?`?

coriolinus (commented):
First pass review looks good! Lack of a formal approval doesn't mean anything, I just need to go back through after lunch and decide whether anything I wrote is actually a blocker. 

The structs you use for the abstract representation look reasonable to me. I'm not sure about the naming of `flatten`/`KSEFlattened`, but names are hard and nothing obviously better springs immediately to mind. 

It might be worth factoring the implementation into a few sub-modules. I spend a fair amount of time scrolling back and forth looking at various definitions; as a matter of personal preference, I find it simpler to tab between a handful of short files than to scroll through one long file.

crypto-attributes/src/entity_derive.rs
```diff
@@ -0,0 +1,567 @@
...
+        }
+    }
+}
+
+// Now identical to column, but
+// subject to change once more diverse entities are supported.
+struct IdColumn {
+    name: Ident,
+    column_type: ColumnType,
+}
```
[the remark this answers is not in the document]
SimonThormeyer: Yes, exactly. For instance, there will be `#[id(blob)]`.

coriolinus (approved):
Quick second pass looks good; I don't think that any of my notes are necessarily blocking. So, approved!

crypto-attributes/src/entity_derive.rs
```diff
@@ -0,0 +1,567 @@
...
+        let struct_name = input.ident.clone();
+
+        // #[entity(collection_name = "my_collection")]
+        let (collection_name, no_upsert) = Self::parse_collection_name_and_no_upsert(&input.attrs)?;
+        if collection_name.is_empty() {
+            return Err(syn::Error::new_spanned(
+                input,
+                format!(
+                    "Provide a collection in the `entity` attribute, e.g: `#[entity(collection_name = \"{}\")]`.",
+                    struct_name.to_string().to_lowercase()
```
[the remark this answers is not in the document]
SimonThormeyer: I would like not to have to provide an explicit `collection_name`, however, I think it's a bit risky because our tables aren't named consistently.

E.g., `e2ei_enrollment` (snake case, singular) vs `e2ei_intermediate_certs` (snake case, plural). 
This is why it's possibly better to keep the naming explicit.

Would it be worth doing the transformation for the error message?

crypto-attributes/src/entity_derive.rs
```diff
@@ -0,0 +1,567 @@
...
+
+enum ColumnType {
+    String,
+    Bytes,
+}
+
+impl ColumnType {
+    fn parse(ty: &Type) -> Result<Self, syn::Error> {
+        if let Type::Path(type_path) = ty {
+            let segment = type_path.path.segments.last().unwrap();
```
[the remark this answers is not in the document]
SimonThormeyer: Oh yeah, I can see how discarding the path can lead to problems.
And this is indeed much simpler, thanks!

PR #868, opened 2025-01-20
build: add a script to merge a PR according to our current workflow

coriolinus:
We have decided to require both signed commits and a linear history in github, which breaks the github "merge" button completely.

This script does the right thing to accomplish signed commits with a linear history, because it's kind of a pain to do manually each time.

# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/scripts/merge-pr.sh
+++ b/scripts/merge-pr.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+
+# We've disabled Github's merge button because it doesn't fast-forward properly.
+# Instead, we can use this script to merge an approved PR.
+#
+# Requires: `git`, `gh`, 'jq'
+#
+# Usage: $0 [branch-or-pr-number]
+#
+# If the current branch is `main` then `branch-or-pr-number` is required.
+
+set -e -o pipefail
+
+function bail() {
+    msg="$1"
+    if [ -n "$msg" ]; then
+        # we do want to emit to stderr here
+        #shellcheck disable=SC2210
+        echo >2 "$msg"
+    fi
+    exit 1
+}
+
+current_branch="$(git branch --show-current)"
+if [ -z "$1" ]; then
+    if [ "$current_branch" == "main" ]; then
+        bail "on main; must specify the PR number or branch name to merge"
+    else
+        branch="$current_branch"
+    fi
+else
+    # if it's a number, assume it's a PR number and get the branch name. otherwise just a branch.
+    if [[ "$1" =~ ^[0-9]+$ ]]; then
+        branch="$(gh pr view "$1" --json headRefName | jq -r ".headRefName")"
+    else
+        branch="$1"
+    fi
+fi
+
+# ensure the branch is approved
+status_json="$(gh pr view "$branch" --json reviewDecision,statusCheckRollup)"
+review_decision="$(jq -r .reviewDecision <<< "$status_json")"
+if [ "$review_decision" != "APPROVED" ]; then
+    bail "$branch has not yet been approved"
+fi
+
+# ensure all checks have completed successfully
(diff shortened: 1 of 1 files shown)


scripts/merge-pr.sh
```diff
@@ -0,0 +1,70 @@
...
+
+current_branch="$(git branch --show-current)"
+if [ -z "$1" ]; then
+    if [ "$current_branch" == "main" ]; then
+        bail "on main; must specify the PR number or branch name to merge"
+    else
+        branch="$current_branch"
+    fi
+else
+    # if it's a numbe, assume it's a PR number and get the branch name. otherwise just a branch.
```
istankovic, line 32: Typo: numbe -> number

scripts/merge-pr.sh
```diff
@@ -0,0 +1,70 @@
...
+# We've disabled Github's merge button because it doesn't fast-forward properly.
+# Instead, we can use this script to merge an approved PR.
+#
+# Requires: `git`, `gh`, 'jq'
+#
+# Usage: $0 [branch-or-pr-number]
+#
+# If the current branch is `main` then `branch-or-pr-number` is required.
+
+set -e
```
istankovic, line 12: I'd also add `pipefail` here.

scripts/merge-pr.sh
```diff
@@ -0,0 +1,70 @@
...
+# ensure all checks have completed successfully
+non_success="$(jq -r '.statusCheckRollup[] | select(.__typename == "CheckRun" and ((.status == "COMPLETED" and (.conclusion == "SUCCESS" or .conclusion == "SKIPPED")) | not)) | .name' <<< "$status_json")"
+if [ -n "$non_success" ]; then
+    bail "some CI checks are incomplete or unsuccessful:\n$non_success"
+fi
+
+# ensure that the branch is at the tip of `origin/main` for a linear history
+git fetch
+git checkout "$branch"
+if ! git rebase origin/main; then
```
istankovic, line 56: I think this is a prime use case for `git merge-base ---is-ancestor`. :smile:

scripts/merge-pr.sh
```diff
@@ -0,0 +1,70 @@
...
+    git rebase --abort
+    bail "$branch did not cleanly rebase onto origin/main; do so manually and try again"
+fi
+
+# if rebase moved the tip then force-push to ensure github is tracking the new history
+# this resets CI, but doesn't mess with the approvals. We can assume CI is OK, at this point.
```
istankovic, line 62: > We can assume CI is OK, at this point.

I'm not really a fan of this. This was already an issue before and we had PRs that failed on main after rebase, but not on the original branch.

istankovic (approved):
Looks good, with a few smaller things.

scripts/merge-pr.sh
coriolinus: Haha, I was looking for something like that, but google wasn't helpful. I think this behavior, which does the rebase as a side-effect, is actually better though.

There are 3 cases we care about:

1. The branch already originates from `origin/main`
2. The branch originates from some older commit than `origin/main` but applies cleanly.
3. The branch originates from some older commit than `origin/main` but does not apply cleanly.

`git merge-base --is-ancestor` will do a great job distinguishing case 1, but conflates cases 2 and 3. We do want to distinguish between those.

`git rebase` as written here conflates 1 and 2, but distinguishes case 3; I think that's the behavior we actually want.

scripts/merge-pr.sh
coriolinus: I know, and I did think about doing the CI checks after this point. The problem is just how long the CI takes. For a clean rebase, I think it's "safe enough" to assume that CI will still pass; the cases where a clean rebase introduces problems should be unusual.

PR #885, opened 2025-01-24
chore: improve logs when buffering messages

typfel:
# What's new in this PR

We have observed some unexpected behaviour when restoring buffered proposals, so let's improve the logs.

----

- The **PR Title**

### DIFF
--- a/crypto/src/mls/conversation/buffer_messages.rs
+++ b/crypto/src/mls/conversation/buffer_messages.rs
@@ -4,6 +4,7 @@
 //! Feel free to delete all of this when the issue is fixed on the DS side !
 
 use super::{Error, Result};
+use crate::obfuscate::Obfuscated;
 use crate::{
     context::CentralContext,
     group_store::GroupStoreValue,
@@ -71,7 +72,6 @@ impl MlsConversation {
         is_rejoin: bool,
     ) -> Result<Option<Vec<MlsBufferedConversationDecryptMessage>>> {
         // using the macro produces a clippy warning
-        info!("restore_pending_messages");
         let result = async move {
             let keystore = backend.keystore();
             let group_id = self.id().as_slice();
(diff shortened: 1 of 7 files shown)


crypto/src/mls/conversation/decrypt.rs
```diff
@@ -475,7 +477,12 @@ impl CentralContext {
...
         let decrypt_message = match decrypt_message {
-            Err(Error::BufferedFutureMessage) => self.handle_future_message(id, message).await?,
+            error @ Err(Error::BufferedFutureMessage { .. }) => {
+                self.handle_future_message(id, message).await?;
+                let Err(Error::BufferedFutureMessage { message_epoch }) = error;
+                info!("Buffered future message from epoch {message_epoch}");
+                return error;
+            }
             _ => decrypt_message?,
         };
```
SimonThormeyer, line 487: Not really nice, but should fix the compile error. 

```suggestion
 let decrypt_message = match decrypt_message {
            Err(Error::BufferedFutureMessage { message_epoch }) => {
                self.handle_future_message(id, message).await?;
                info!("Buffered future message from epoch {message_epoch}");
                return decrypt_message;
            }
            _ => decrypt_message?,
        };
```

SimonThormeyer, line 487:
```suggestion
        if let Err(Error::BufferedFutureMessage { message_epoch }) = decrypt_message {
            self.handle_future_message(id, message).await?;
            info!("Buffered future message from epoch {message_epoch}");
            return decrypt_message;
        }
        
        let decrypt_message = decrypt_message?;
```

crypto/src/mls/conversation/buffer_messages.rs
```diff
@@ -41,7 +35,7 @@ impl CentralContext {
             .save::<MlsPendingMessage>(pending_msg)
             .await
             .map_err(KeystoreError::wrap("saving pending mls message"))?;
-        Err(Error::BufferedFutureMessage)
+        Ok(())
```
coriolinus, line 38: This change looks significant. This is a bugfix, yes?

coriolinus (approved):
This all makes sense to me! 

It would be great to add a test which passes now that `handle_future_message` returns `Ok`, which wouldn't if it still incorrectly propagated the error. But that's deep in stretch goal territory; it's not essential. I doubt we're likely to accidentally revert this.

coriolinus (approved):
We can ignore the previous comment about adding a test; I understand the flow better now.

PR #904, opened 2025-02-04
feat: use entity derive macros for tables with hex ids [WPB-15434]

SimonThormeyer:
# What's new in this PR
Subtask of the above item, creating a separate PR to avoid large changeset

----

- The **PR Title**


keystore/src/error.rs
```diff
@@ -36,7 +36,7 @@ pub enum MissingKeyErrorKind {
     #[error("MLS CredentialBundle")]
     MlsCredential,
     #[error("MLS Persisted Group")]
-    MlsGroup,
+    PersistedMlsGroup,
```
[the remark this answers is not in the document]
SimonThormeyer: In this case I would rather leave it, as all error variants related to MLS start with `MLS`, but this one would have it in the middle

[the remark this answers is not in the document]
coriolinus: Ah, you're right, I hadn't spotted that pattern.

crypto-macros/src/entity_derive/parse.rs
```diff
@@ -137,9 +188,10 @@ impl ColumnType {
         match type_string.as_str() {
             "String" | "std::string::String" => Ok(Self::String),
             "Vec<u8>" | "std::vec::Vec<u8>" => Ok(Self::Bytes),
+            "Option<Vec<u8>>" | "Option<std::vec::Vec<u8>>" => Ok(Self::OptionalBytes),
```
coriolinus, line 191: Unfortunately there are four variants here to match on:

- `Option<Vec<u8>>`
- `Option<std::vec::Vec<u8>>`
- `std::option::Option<Vec<u8>>`
- `std::option::Option<std::vec::Vec<u8>>`

coriolinus (approved):
Looks good! I only manually looked over the derived implementations for `PersistedMlsGroup`, but that one seemed very reasonable.

crypto-macros/src/entity_derive/parse.rs
SimonThormeyer: And `core::option::Option<Vec<u8>>`, `core::option::Option<std::vec::Vec<u8>>`, I guess :)

coriolinus: Heh, right.

PR #907, opened 2025-02-05
feat: buffer commits until all required proposals have arrived [WPB-15810]

coriolinus:
After creating a test replicating [the 15810 scenario](https://wearezeta.atlassian.net/browse/WPB-15810). Turns out that one key issue was that clients who received a commit prior to a referenced proposal were just giving up. This PR fixes that issue by adding the capability to buffer one commit per conversation. When there is a buffered commit, it is re-attempted each time a proposal arrives until it succeeds.

Clients still need to watch out for desync! If the required proposal is simply never delivered, the client will lag behind until the epoch changes again, at which point they can prove desync. This fix solves everything under the hood, in the expected case that all messages are delivered, and reorderings don't delay any single message by more than a few seconds. 

----

- The **PR Title**

### DIFF
--- a/crypto/src/mls/conversation/buffer_messages.rs
+++ b/crypto/src/mls/conversation/buffer_messages.rs
@@ -418,4 +418,212 @@ mod tests {
         )
         .await
     }
+
+    /// Replicating [WPB-15810]
+    ///
+    /// [WPB-15810]: https://wearezeta.atlassian.net/browse/WPB-15810
+    #[apply(all_cred_cipher)]
+    async fn wpb_15810(case: TestCase) {
+        use openmls::{
+            group::GroupId,
+            prelude::{ExternalProposal, SenderExtensionIndex},
+        };
+
+        use crate::mls;
+
+        if case.is_pure_ciphertext() {
+            // The use case tested here requires inspecting your own commit.
+            // Openmls does not support this currently when protocol messages are encrypted.
+            return;
+        }
+        run_test_with_client_ids(
+            case.clone(),
+            ["external_0", "new_member", "member_27", "observer", "114", "115"],
+            move |[external_0, new_member, member_27, observer, member_114, member_115]| {
+                Box::pin(async move {
+                    // scenario start: everyone except "new_member" is in the conversation
+                    let conv_id = conversation_id();
+
+                    // set up external_0 as the backend / delivery service
+                    let signature_key = external_0.client_signature_key(&case).await.as_slice().to_vec();
+                    let mut config = case.cfg.clone();
+                    observer
+                        .context
+                        .set_raw_external_senders(&mut config, vec![signature_key])
+                        .await
+                        .unwrap();
+
+                    // create and initialize the conversation
+                    observer
+                        .context
+                        .new_conversation(&conv_id, case.credential_type, config)
+                        .await
+                        .unwrap();
+
+                    // everyone else except new_member joins (also except observer, who created it)
+                    observer
+                        .invite_all(&case, &conv_id, [&member_114, &member_115, &member_27])
+                        .await
+                        .unwrap();
+
+                    // Everyone should agree on the overall state here, to wit: the group consists of everyone
+                    // except "new_member", and "external_0", and no messages have been sent.
+                    // At this point only the observer is going to receive messages, because that shouldn't impact group state.
+
+                    // external 0 sends a proposal to remove 114
+                    let leaf_of_114 = observer.index_of(&conv_id, member_114.get_client_id().await).await;
+                    let sender_index = SenderExtensionIndex::new(0);
+                    let sc = case.signature_scheme();
+                    let ct = case.credential_type;
+                    let cb = external_0.find_most_recent_credential_bundle(sc, ct).await.unwrap();
+                    let group_id = GroupId::from_slice(&conv_id[..]);
+                    let epoch = observer.get_conversation_unchecked(&conv_id).await.group.epoch();
+                    let proposal_remove_114_1 = ExternalProposal::new_remove(
+                        leaf_of_114,
+                        group_id.clone(),
+                        epoch,
+                        &cb.signature_key,
+                        sender_index,
+                    )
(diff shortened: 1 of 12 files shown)


crypto/src/mls/conversation/buffer_messages.rs
```diff
@@ -418,4 +415,223 @@ mod tests {
...
+                    .unwrap();
+
+                    // now our observer receives these messages out of order
+                    println!("observer executing first proposal");
+                    observer
+                        .context
+                        .decrypt_message(&conv_id, &proposal_remove_114_1.to_bytes().unwrap())
+                        .await
+                        .unwrap();
+                    println!("observer executing second proposal");
```
SimonThormeyer, line 521: This looks like a leftover from debugging

crypto/src/mls/conversation/buffer_messages.rs
```diff
@@ -418,4 +415,223 @@ mod tests {
         )
         .await
     }
+
+    /// Replicating [WPB-15810]
+    ///
+    /// [WPB-15810]: https://wearezeta.atlassian.net/browse/WPB-15810
+    #[apply(all_cred_cipher)]
+    async fn wpb_15810(case: TestCase) {
```
SimonThormeyer, line 423: Is it useful to keep this test name or should we describe what it is actually testing in the name?

crypto/src/mls/conversation/buffer_messages.rs
```diff
@@ -418,4 +415,223 @@ mod tests {
...
+                    ));
+                    println!("executing commit adding new user");
+                    observer
+                        .context
+                        .decrypt_message(&conv_id, &new_member_join_commit.to_bytes().unwrap())
+                        .await
+                        .unwrap();
+
+                    // observer.context.commit_pending_proposals(&conv_id).await.unwrap();
+                    // let commit = observer.mls_transport.latest_commit().await;
```
SimonThormeyer, line 538: This also looks like a leftover from debugging

crypto/src/mls/conversation/buffer_messages.rs
```diff
@@ -418,4 +415,223 @@ mod tests {
...
+                        .decrypt_message(&conv_id, &proposal_remove_114_2.to_bytes().unwrap())
+                        .await
+                        .unwrap();
+
+                    // new_member
+                    //     .context
+                    //     .decrypt_message(&conv_id, new_member_join_commit.to_bytes().unwrap())
+                    //     .await
+                    //     .unwrap();
+                    // observer.try_talk_to(&conv_id, &new_member).await.unwrap();
```
SimonThormeyer, line 563: This as well

crypto/src/mls/conversation/decrypt.rs
```diff
@@ -196,6 +224,38 @@ impl MlsConversation {
...
+                        // presumably they are still coming in. So just return the normal
+                        // result from this proposal, below.
+                    } else {
+                        // either the commit applied successfully, in which case its return value
+                        // should override the return value from the proposal, or it raised some kind
+                        // of error, in which case the caller needs to know about that.
+                        return process_result
+                            .map_err(RecursiveError::mls_conversation("processing buffered commit"))
+                            .map_err(Into::into);
+                    }
```
SimonThormeyer, line 256: is that a better pattern than this?
```rust
if !matches!(process_result, Err(Error::BufferedCommit)) {
    // either the commit applied successfully, in which case its return value
    // should override the return value from the proposal, or it raised some kind
    // of error, in which case the caller needs to know about that. 
    return process_result
        .map_err(RecursiveError::mls_conversation("processing buffered commit"))
        .map_err(Into::into);
}
```

crypto/src/mls/conversation/decrypt.rs
```diff
@@ -318,29 +378,93 @@ impl MlsConversation {
         Ok(decrypted)
     }
 
+    /// Cache the bytes of a pending commit in the backend.
+    ///
+    /// By storing the raw commit bytes and doing deserialization/decryption from scratch, we preserve all
+    /// security guarantees. When we do restore, it's as though the commit had simply been received later.
```
SimonThormeyer, line 384: Nice comment!

SimonThormeyer (approved):
Approved with a few comments. Good work!!

SimonThormeyer:
This commit 033505fc1fbb4cbc6b04556e1701339702a307ea
looks like it does more than just improve the test legibility. Maybe factor out the other stuff and fixup the changes to the new test into the commit that introduced it?

crypto/src/mls/conversation/buffer_messages.rs
coriolinus: Within a test case I typically retain output like this. It gets suppressed during passing runs, and can give some additional context if something goes unexpectedly wrong.

SimonThormeyer: Didn't know that, sounds useful!

crypto/src/mls/conversation/buffer_messages.rs
coriolinus: I can rename it if you find it useful, but I don't mind seeing test cases named after issues. From there it's a straightforward link to long-form prose about what the issue was and how it was debugged/fixed. In my opinion that's better than a few concise words in the function name.

crypto/src/mls/conversation/buffer_messages.rs
coriolinus: True, I'll remove it.

crypto/src/mls/conversation/buffer_messages.rs
SimonThormeyer: We can keep it as is!

crypto/src/mls/conversation/decrypt.rs
coriolinus: I did it like this so that I could comment on each branch. I'd expect similar machine code in either case. But you're right, we should use the more normal idiom.

PR #913, opened 2025-02-10
wip

istankovic:


coriolinus (approved):
It's surprising that that's all it takes! But I guess all the real configuration is elsewhere.

PR #914, opened 2025-02-12
DO NOT MERGE: attempt 3 to backport commit buffering

coriolinus:
# The Process So Far

1. Started with the final commit in the list, the one which adds the test case for 15810. 
2. It doesn't compile due to changed / missing functions. 
3. Looked up the commits which added / changed the relevant functions.
4. For each such commit:
    1. cherry-pick that commit
    2. fix merge conflicts across multiple files
    3. Rebase to insert the commit earlier so that there is some kind of logical ordering.
5. GOTO 2

## Why I'm giving up on this approach

This was the basic approach I wanted to take last Friday and Monday. It's the obvious approach to this problem! But it has issues of its own:

- It's extremely time-consuming
- It ends up recursively adding large portions of mostly-unrelated code to the codebase
- It ends up accidentally backporting key 4.x features such as the mls transport provider to 3.x
- As we do work on the PR level, not the commit level, it also is extremely likely to accidentally miss some work which is logically necessary for an earlier feature, but not implemented in the particular commit-set that we've cherry-picked

This is up as a sort of proof of work to remind me not to try to do this again, because the first 2 failed attempts were not enough apparently. We should close this issue and delete this cursed branch as soon as we accomplish the goal of backporting commit buffering through other means.

### DIFF
--- a/crypto-ffi/bindings/js/CoreCrypto.ts
+++ b/crypto-ffi/bindings/js/CoreCrypto.ts
@@ -20,12 +20,13 @@ import initWasm, {
     ConversationConfiguration as ConversationConfigurationFfi,
     CoreCrypto as CoreCryptoFfi,
     CoreCryptoContext as CoreCryptoContextFfi,
-    CoreCryptoWasmCallbacks,
+    WasmMlsTransportResponse,
     CoreCryptoWasmLogger,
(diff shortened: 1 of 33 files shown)


coriolinus:
I'm confident enough in #915 that I am happy to consign this to the depths.

PR #922, opened 2025-02-20
chore: implement `ConversationGuard` [WPB-15577]

coriolinus:
# What's new in this PR

Start working on refactoring conversation methods so we don't have to duplicate the whole `Conversation` API on `CentralContext`.

----

- The **PR Title**

### DIFF
--- a/crypto/src/mls/conversation/conversation_guard.rs
+++ b/crypto/src/mls/conversation/conversation_guard.rs
@@ -1,0 +1,63 @@
+// this isssue is about creating this helper struct,
+// but we'll see the dead code warning until we actually use it in follow-up work
+#![expect(dead_code)]
+
+use async_lock::{RwLockReadGuard, RwLockWriteGuard};
+use mls_crypto_provider::MlsCryptoProvider;
+
+use crate::{context::CentralContext, group_store::GroupStoreValue, RecursiveError};
+
+use super::{commit::MlsCommitBundle, Error, MlsConversation, Result};
+
+/// A Conversation Guard wraps a `GroupStoreValue<MlsConversation>`.
+///
+/// By doing so, it permits mutable accesses to the conversation. This in turn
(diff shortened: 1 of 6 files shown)


crypto/src/mls/conversation/conversation_guard.rs
```diff
@@ -0,0 +1,63 @@
+// this isssue is about creating this helper struct,
+// but we'll see the dead code warning until we actually use it in follow-up work
```
istankovic, line 2: Just reading this comment, it's not clear what `this issue` refers to (I'm guessing the related Jira issue).
I'd prefer something like `Temporarily silence the compiler, until we use ConversationGuard in follow-up work`.

coriolinus: Yes, but given that the PR removing it is in progress and probably going to be ready for review today, could we proceed anyway?

crypto/src/mls/conversation/conversation_guard.rs
```diff
@@ -0,0 +1,63 @@
...
+/// By doing so, it permits mutable accesses to the conversation. This in turn
+/// means that we don't have to duplicate the entire `MlsConversation` API
+/// on `CentralContext`.
+pub(crate) struct ConversationGuard {
+    inner: GroupStoreValue<MlsConversation>,
+    central_context: CentralContext,
+}
+
+impl ConversationGuard {
+    pub(crate) fn new(inner: GroupStoreValue<MlsConversation>, central_context: CentralContext) -> Self {
```
istankovic, line 23: Why do we take ownership of the central context?

coriolinus: They're cheap to clone (wrapping an `Arc` internally), and retaining ownership here means that many function signatures get simpler down the line.

istankovic (approved):

PR #933, opened 2025-02-25
fix: android docs [WPB-16037]

typfel:
# What's new in this PR

Re-enable android docs via dokka.

----

- The **PR Title**


coriolinus (approved):
Looking at the output that CI generated ([here](https://github.com/wireapp/core-crypto/actions/runs/13532835680/artifacts/2652399645)), we see many instances of `<Error class: unknown class>) in places where end users probably want to know specifically what goes in the constructor.

image

That said, some docs are better than no docs, so I'm approving this.

typfel:
> Looking at the output that CI generated ([here](https://github.com/wireapp/core-crypto/actions/runs/13532835680/artifacts/2652399645)), we see many instances of `<Error class: unknown class>`) in places where end users probably want to know specifically what goes in the constructor.
> 
> image
> 
> That said, some docs are better than no docs, so I'm approving this.

Ah interesting this is referencing internal uniffi classes so the problem is likely these methods shouldn't be public.

coriolinus:
Huh. Those are the constructors though. How does a client in Kotlin get ahold of a `CoreCrypto` instance if not via the constructor?

istankovic:
@typfel why not fix https://wearezeta.atlassian.net/browse/WPB-16036 first?

typfel:
> @typfel why not fix WPB-16036 first?

Because it relies on this PR

istankovic:
> > @typfel why not fix WPB-16036 first?
> 
> Because it relies on this PR. In WPB-16036 i will use what I setup here like this:
[...]

Ah, I missed that you already moved to Dokka 2 in this PR, nice!

istankovic (approved):

PR #942, opened 2025-02-28
chore!: remove the cryptobox-migrate feature

istankovic:
All clients have moved away from cryptobox so it doesn't make sense to have this.

:notebook: This is a breaking change.


crypto/src/proteus.rs
coriolinus, line 1: Great job that we can get rid of this entirely!

coriolinus (dismissed):
I love getting rid of obsolete features!

crypto/src/proteus.rs
coriolinus: Oops, I misread the GH interface. We're not deleting the file entirely, but still, great to remove this much code from it.

crypto/Cargo.toml
```diff
@@ -15,21 +15,12 @@ crate-type = ["lib", "cdylib"]
...
     "dep:proteus-wasm",
     "dep:proteus-traits",
     "core-crypto-keystore/proteus-keystore",
 ]
-cryptobox-migrate = [
-    "proteus",
-    "proteus-wasm?/cryptobox-identity",
-    "dep:async-fs",
-    "dep:futures-lite",
-    "dep:rexie",
```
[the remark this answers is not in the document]
istankovic: Removal of `rexie` is not so straightforward as keystore v1 migration depends on it, and removing that requires making sure that all clients in production use CC >= 2.0. Plus, we will need to adjust the migration builder for the new minimum db version. That is all material for a separate PR.

PR #952, opened 2025-03-06
fix!: switch to a proper byte array type for database keys [WPB-10109]

istankovic:
This PR moves away from using strings to handle CoreCrypto database keys. Instead, a database key is now represented by a `DatabaseKey` type that is `[u8; 32]` under the hood. Wrappers enforce the correct type and key length.


crypto-ffi/src/generic/mod.rs
```diff
@@ -1650,3 +1651,21 @@ mod tests {
         });
     }
 }
+
+pub struct DatabaseKey(core_crypto_keystore::DatabaseKey);
+
+impl DatabaseKey {
+    // TODO: This is here only because we need to construct an instance in interop.
+    // Remove it once we drop the FFI client from interop.
+    pub fn new(key: core_crypto_keystore::DatabaseKey) -> DatabaseKey {
```
coriolinus, line 1660: could be `#[derive(derive_more::Constructor)]`

crypto-ffi/src/generic/mod.rs
```diff
@@ -1130,7 +1130,7 @@ impl CoreCrypto {
             .map_err(CoreCryptoError::generic())?;
         let configuration = MlsCentralConfiguration::try_new(
             path,
-            key,
+            key.0.clone(),
```
coriolinus, line 1133: Instead of explicit `.0`, maybe `#[derive(derive_more::Deref)]` on the local `DatabaseKey` so cloning produces the inner type?

Should produce a smaller diff when we can eliminate the local wrapper.

coriolinus (dismissed):
I'm concerned about the `Debug` impl; otherwise, this broadly looks good!

[edit] Actually I'm also concerned about adding the leading `x` to the `set_key` function.

keystore/src/connection/mod.rs
```diff
@@ -56,14 +61,73 @@ pub trait DatabaseConnectionRequirements: Sized + Send {}
...
 pub trait DatabaseConnectionRequirements: Sized {}
 
+const DB_KEY_LEN: usize = 32;
+
+/// The key used to encrypt the database.
+#[derive(Clone, Zeroize, ZeroizeOnDrop)]
+pub struct DatabaseKey([u8; DB_KEY_LEN]);
+
+impl DatabaseKey {
+    pub const LEN: usize = DB_KEY_LEN;
```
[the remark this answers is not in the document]
istankovic: Indeed, that actually works, nice! Thanks! :purple_heart:

keystore/src/connection/mod.rs
```diff
@@ -56,14 +61,73 @@ pub trait DatabaseConnectionRequirements: Sized + Send {}
...
+
+    pub fn generate() -> DatabaseKey {
+        DatabaseKey(rand::random::<[u8; DB_KEY_LEN]>())
+    }
+}
+
+impl fmt::Debug for DatabaseKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        f.write_str("DatabaseKey(0x")?;
```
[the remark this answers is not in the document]
istankovic: Good point. We already have `Obfuscate` in `crypto`, however we can't really use it here, so I'm thinking of just copying the approach (hash + truncation). It was introduced precisely for cases like this, but unfortunately it's not available outside `crypto`.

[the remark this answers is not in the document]
istankovic: Fixed.

keystore/src/connection/platform/generic/mod.rs
```diff
@@ -119,7 +121,17 @@ impl SqlCipherConnection {
         Ok(conn)
     }
 
-    fn init_with_key_in_memory(_path: &str, key: &str) -> CryptoKeystoreResult<Self> {
+    fn set_key(conn: &mut rusqlite::Connection, key: &DatabaseKey) -> CryptoKeystoreResult<()> {
+        // Make sqlite use raw key data, without key derivation. Also make sure to zeroize
+        // the string containing the key after the call.
+        let mut key = format!("x'{}'", hex::encode(key));
```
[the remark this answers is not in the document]
istankovic: Indeed, this is a change in the interpretation of the key. The migration function will take both the old key (a string) and the new key (a byte array), open the database with the old key and then switch to the new key (type).

keystore/src/connection/platform/generic/mod.rs
```diff
@@ -325,15 +338,15 @@ impl DatabaseConnectionRequirements for SqlCipherConnection {}
 impl<'a> DatabaseConnection<'a> for SqlCipherConnection {
     type Connection = MutexGuard<'a, rusqlite::Connection>;
 
-    async fn open(name: &str, key: &str) -> CryptoKeystoreResult<Self> {
+    async fn open(name: &str, key: &DatabaseKey) -> CryptoKeystoreResult<Self> {
         let name = name.to_string();
-        let key = key.to_string();
+        let key = key.clone();
         Ok(unblock(move || Self::init_with_key(&name, &key)).await?)
```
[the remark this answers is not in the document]
istankovic: > I wonder if Rustc is now smart enough that we could pass along `name` and `key` without cloning each of them? It's not that expensive, but it seems silly to design a nice API which uses references in appropriate places and then just internally clone them.

Pretty much my thoughts as I was writing this. Unfortunately, it's due to `unblock` from the `blocking` crate:
```
pub fn unblock<T, F>(f: F) -> Task<T>
where
    F: FnOnce() -> T + Send + 'static,
    T: Send + 'static,
{
    Executor::spawn(async move { f() })
}
```

crypto-ffi/src/generic/mod.rs
istankovic: Nice, didn't know that, thanks!

keystore/src/connection/platform/generic/mod.rs
coriolinus, line 127: Looking at [the docs](https://www.zetetic.net/sqlcipher/sqlcipher-api/#key), I think the hex key may need to be uppercase. 
```suggestion
        let mut key = format!("x'{}'", hex::encode_upper(key));
```

istankovic: Lowercase also works fine, I tested it.

keystore/src/connection/platform/wasm/migrations/mod.rs
```diff
@@ -38,11 +41,67 @@ const DB_VERSION_0: u32 = db_version_number(0);
...
+    /// The `from` argument represents the version the migration is performed from the function will
+    /// return the version number of the DB resulting from the migration.
+    ///
+    /// To add a new migration, add a new match arm below the latest one.
+    /// It must match on the version it migrates from, and call a function that performs the migration
+    /// workload, which returns the version it migrates to, which is the same value as TARGET_VERSION in
+    /// the function above at the time the migration is added.
+    ///
+    /// However, do not use the constant but hardcode the value into the function.
+    /// This way it will keep working once a new migration is added after it.
```
SimonThormeyer, line 88: Maybe this docstring should better live on the new, now relevant function. This one could refer to the new function.

istankovic: Done.

keystore/src/connection/platform/generic/mod.rs
```diff
@@ -132,6 +133,34 @@ impl SqlCipherConnection {
...
+
+        // ? iOS WAL journaling fix; see details here: https://github.com/sqlcipher/sqlcipher/issues/255
+        #[cfg(target_os = "ios")]
+        ios_wal_compat::handle_ios_wal_compat(&conn, path)?;
+
+        // Enable WAL journaling mode
+        conn.pragma_update(None, "journal_mode", "wal")?;
+
+        // Disable FOREIGN KEYs - The 2 step blob writing process invalidates foreign key checks unfortunately
+        conn.pragma_update(None, "foreign_keys", "OFF")?;
```
coriolinus, line 153: Can we turn the foreign keys back on afterwards?

keystore/src/connection/platform/generic/mod.rs
```diff
@@ -132,6 +133,34 @@ impl SqlCipherConnection {
...
+
+        Self::run_migrations(&mut conn)?;
+
+        // Rekey the database.
+        let mut key = format!("x'{}'", hex::encode(new_key));
+        let result = conn.pragma_update(None, "rekey", &key);
+        key.zeroize();
+        Ok(result?)
```
coriolinus, line 161: This line should be either just `result` or `result.map_err(Into::into)`. `Ok(result?)` is a silly pattern.

coriolinus (approved):
Great job!

keystore/src/connection/platform/generic/mod.rs
istankovic: I have no idea, I didn't really touch that (it was there before) and am not familiar with the details. @SimonThormeyer ?

keystore/src/connection/platform/generic/mod.rs
istankovic: Fixed.

keystore/src/connection/platform/generic/mod.rs
SimonThormeyer: Same situation for me, unfortunately.

crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/GeneralTest.kt
```diff
@@ -37,4 +37,20 @@ class DatabaseKeyTest {
         val exc = assertFailsWith<CoreCryptoException.Other> { wrapException { CoreCrypto(path, key) } }
         assertEquals("Invalid database key size, expected 32, got 48", exc.message)
     }
+
+    @Test
+    @OptIn(ExperimentalPathApi::class)
```
SimonThormeyer, line 42: Do we need this? What would be the non-experimental API for `Path`?

SimonThormeyer (approved):
Awesome work, great job!!

keystore/src/connection/mod.rs
```diff
@@ -40,14 +46,65 @@ pub trait DatabaseConnectionRequirements: Sized + Send {}
...
+    type Error = CryptoKeystoreError;
+
+    fn try_from(buf: &[u8]) -> Result<Self, Self::Error> {
+        if buf.len() != Self::LEN {
+            Err(CryptoKeystoreError::InvalidDbKeySize {
+                expected: Self::LEN,
+                actual: buf.len(),
+            })
+        } else {
+            Ok(Self(buf.try_into().unwrap()))
```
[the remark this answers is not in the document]
istankovic: Yes, if the length is correct, I see no way how that can fail.

keystore/src/connection/mod.rs
```diff
@@ -40,14 +46,65 @@ pub trait DatabaseConnectionRequirements: Sized + Send {}
...
+/// The key used to encrypt the database.
+#[derive(Clone, Zeroize, ZeroizeOnDrop, derive_more::From)]
+pub struct DatabaseKey([u8; Self::LEN]);
+
+impl DatabaseKey {
+    pub const LEN: usize = 32;
+
+    pub fn generate() -> DatabaseKey {
+        DatabaseKey(rand::random::<[u8; Self::LEN]>())
+    }
```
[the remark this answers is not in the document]
istankovic: > Should this be under `#[cfg(test)]`?

It could be, and if IIRC it was originally under `cfg(test)`, but I found it nicer to just have this convenience function so we don't need to copy the same code in e.g. interop, where testing code is not actually built with `test`.

crypto-ffi/bindings/js/test/CoreCrypto.test.ts
```diff
@@ -1036,3 +1036,79 @@ describe("database key", () => {
...
+    it("migrating key type to bytes works", async () => {
+        const stores = await import("./db-v10002003-dump.json");
+
+        const response = await fetch(
+            "https://cdn.jsdelivr.net/npm/idb@8/build/umd.js"
+        );
+        if (!response.ok)
+            throw new Error(`failed to fetch script: ${response.statusText}`);
+
+        await browser.executeScript(await response.text(), []);
```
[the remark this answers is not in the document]
istankovic: Sure, will add a comment, it's definitely missing. What we do here is download the `idb` module and inject it into the browser context. This is so that we can call `idb` API inside the browser. I struggled quite a lot with this, and actually am not too happy about the result, but I could not find a better way to do it.

crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/GeneralTest.kt
istankovic: > Do we need this? What would be the non-experimental API for `Path`?

I was using the experimental `Path` API, however, it turns out the non-experimental `File` API has the same thing, I just needed to convert:
```diff
diff --git a/crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/GeneralTest.kt b/crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/GeneralTest.kt
index 3e963734..721f2572 100644
--- a/crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/GeneralTest.kt
+++ b/crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/GeneralTest.kt
@@ -39,7 +39,6 @@ class DatabaseKeyTest {
     }
 
     @Test
-    @OptIn(ExperimentalPathApi::class)
     fun migrating_key_type_to_bytes_works() = runTest {
         // Skip this test on Android.
         if (System.getProperty("java.vm.name") == "Dalvik") {
@@ -56,6 +55,6 @@ class DatabaseKeyTest {
 
         CoreCrypto(path.absolutePathString(), newKey)
 
-        tmpdir.deleteRecursively()
+        tmpdir.toFile().deleteRecursively()
     }
 }
```

PR #953, opened 2025-03-06
chore: remove uniffi from crypto

istankovic:
core-crypto does/should not need uniffi.


coriolinus (approved):
Assuming green CI, this looks good.

coriolinus (approved):
Still looks good!

PR #965, opened 2025-03-11
chore: switch from pretty_env_logger to env_logger

istankovic:
pretty_env_logger uses humantime, which is unmaintained (see https://rustsec.org/advisories/RUSTSEC-2025-0014). Also there haven't been updates to pretty_env_logger in 2 years.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -260,7 +260,7 @@ checksum = "3b43422f69d8ff38f95f1b2bb76517c91589a924d1559a0e935d7c8ce0274c11"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.93",
+ "syn 2.0.100",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -86,7 +86,3 @@ opt-level = "s"
 # ! This will cause the FFI to stop working because UniFFI stores the Rust <-> cdylib mapping
 # ! in the `.strtab` section of the executable. Stripping this causes everything to stop functioning.
 strip = false
-# panic = "abort"
-
-[profile.dev.package.backtrace]
-opt-level = 3
```
SimonThormeyer, line 92: What does this do? Is it related to the switch to `env_logger`?

SimonThormeyer (approved):

Cargo.toml
coriolinus, line 92: Why change this in this PR?

istankovic: Ugh, that was automatically done by cargo, not sure why it did it, and I didn't notice. Will remove.
Could be that cargo detected this is a stale entry or something, but in any case that's for a different PR.

coriolinus (approved):

Cargo.toml
istankovic: No, see my answer to Peter below.

PR #967, opened 2025-03-11
refactor: rename `conversation_guard()` to `conversation()` [WPB-15582]

SimonThormeyer:
# What's new in this PR
See above and jira item

----

- The **PR Title**


crypto/src/mls/conversation/commit.rs
```diff
@@ -43,12 +43,15 @@ impl CentralContext {
...
-            .as_ref()
-            .map(|c| (c.id().clone(), c.group.epoch().as_u64()));
+        // Can't use map() because of .await
+        let conversation_id_and_epoch = match conversation.as_ref() {
+            None => None,
+            Some(conversation) => {
+                let inner = conversation.conversation().await;
+                Some((inner.id().clone(), inner.group().epoch().as_u64()))
+            }
+        };
```
coriolinus, line 53: This is only used within the `::Retry` response, right? And in that context, when it's `None`, it returns an error.

Why not just move this whole thing into that block?


```rust
let Some(conversation) = conversation.as_ref() else {
    return Err(Error::CannotRetryWithoutConversation);
};
let conversation = conversation.conversation().await;
let conversation_id = conversation.id();
let epoch_before_sending = conversation.group().epoch().as_u64();
```

coriolinus (approved):

crypto/src/mls/conversation/commit.rs
SimonThormeyer: In that case, we'll get the epoch _after_ sending, right? 
We need the epoch before we call `send_commit_bundle()`, because the epoch might change during the callback (see the if condition further below).

SimonThormeyer: But it would probably be better to move at least the id part down to that block, so that we have in the lines you commented on
```rust
let epoch_before_sending = match conversation.as_ref() {
    None => None,
    Some(conversation) => {
        let inner = conversation.conversation().await;
        Some(inner.group().epoch().as_u64())
    }
};
```

and below 
```rust
                    let Some(conversation) = conversation.as_mut() else {
                        return Err(Error::CannotRetryWithoutConversation);
                    };
                    let mut inner = conversation.conversation_mut().await;
                    if let Some(epoch_before_sending) = epoch_before_sending {
                        if epoch_before_sending == inner.group().epoch().as_u64() {
                            continue;
                        }
                    }
```

coriolinus: Getting `epoch_before_sending` in that second block might be a reasonable place to `.expect` it; it makes no sense for it to be `None` if `conversation` exists, and in the weird case of a mismatch, I'd rather panic than just not perform the check.

PR #968, opened 2025-03-11
feat: swift wrapper [WPB-16277]

typfel:
# What's new in this PR

Adds a Swift wrapper around the UniFFI generated code.

----

- The **PR Title**


istankovic:
What is the purpose of the `TestHost` directory and its contents?

crypto-ffi/Makefile.toml
```diff
@@ -72,14 +72,14 @@ args = [
...
-    "--out-dir", "./bindings/swift/WireCoreCrypto/WireCoreCrypto",
+    "--out-dir", "./bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi",
     "--library", "../target/release/deps/libcore_crypto_ffi.${LIBRARY_EXTENSION}"
 ]
 
 [tasks.bindings-swift]
 dependencies = ["_bindings-swift"]
 script = '''
-    mv ./bindings/swift/WireCoreCrypto/WireCoreCrypto/core_crypto_ffi.swift ./bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCrypto.swift
+    mv ./bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_crypto_ffi.swift ./bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/CoreCrypto.swift
```
istankovic, line 82: Can we get rid of this `mv` step entirely by tweaking swift settings in `uniffi.toml`? Maybe changing `cdylib_name` is enough?

crypto-ffi/bindings/swift/WireCoreCryptoUniffi/BuildSettings.xcconfig
```diff
@@ -0,0 +1,23 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+// Configuration settings file format documentation can be found at:
+// https://help.apple.com/xcode/#/dev745c5c974
+
+MARKETING_VERSION = 4.2.0
+CURRENT_PROJECT_VERSION = 1.0
```
istankovic, line 23: We have duplicate definitions of versions. Since we're always going to release the two packages together, it makes sense to have single definition. To that end, we could have a `BuildSettings.xcconfig` file in the parent dir, with these definitions, and then just `#include` them here.

Ideally, we would be able to tell (perhaps via `.pbxproj`?) to use the parent dir's config settings directly. In that case, we would not even need this file, nor the other neighboring file.

crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_cryptoFFI.modulemap
```diff
@@ -0,0 +1,4 @@
+module core_cryptoFFI {
+    header "core_cryptoFFI.h"
+    export *
+}
```
istankovic, line 4: Isn't this file auto-generated?

crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_crypto_ffiFFI.h
```diff
@@ -0,0 +1,1943 @@
+// This file was autogenerated by some hot garbage in the `uniffi` crate.
+// Trust me, you don't want to mess with it!
```
istankovic, line 2: If this is autogenerated, why include it in the repo?

istankovic:
Seems like many (most?) of the files under `WireCoreCryptoUniffi` are autogenerated -- I think we should not add them to git.

istankovic:
Do we have a mechanism that prevents using the low-level (`WireCoreCryptoUniffi`) part directly from client code?

crypto-ffi/Makefile.toml
typfel: I'll maybe we can just remove it now that this is an internal library.

crypto-ffi/bindings/swift/WireCoreCrypto/BuildSettings.xcconfig
```diff
@@ -19,4 +19,6 @@
 // Configuration settings file format documentation can be found at:
 // https://help.apple.com/xcode/#/dev745c5c974
 
-CURRENT_PROJECT_VERSION = 11.0.0
+MARKETING_VERSION=4.2.0
+CURRENT_PROJECT_VERSION=1.0
```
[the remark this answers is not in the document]
typfel: `CURRENT_PROJECT_VERSION` isn't allowed to be x.y.z version. It's generally just a number which goes up.

typfel:
> What is the purpose of the `TestHost` directory and its contents?

I need to run the tests inside a real app (test host) because core crypto uses keychain APIs.

crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_cryptoFFI.modulemap
typfel: No

crypto-ffi/bindings/swift/WireCoreCryptoUniffi/BuildSettings.xcconfig
typfel: Yeah I'll move the BuildSettings.xcconfig to root swift bindings folder and references it from both projects

crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_crypto_ffiFFI.h
typfel: Ah right forget to update the .gitignore

crypto-ffi/bindings/swift/WireCoreCrypto/TestHost/ContentView.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  ContentView.swift
+//  TestHost
+//
+//  Created by Jacob Persson on 10.03.25.
+//
```
coriolinus, line 6: Is this a useful comment given the existence of `git blame`?

typfel: Missed this one it's generated by XCode. I'll fix

crypto-ffi/bindings/swift/WireCoreCrypto/BuildSettings.xcconfig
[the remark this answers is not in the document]
istankovic: Alright, but then I suggest we make it an integer value `patch + minor * 100 + major * 10000`, to keep at least some relation to `MARKETING_VERSION`.

crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_cryptoFFI.modulemap
istankovic: Right, but that's because `uniffi.toml` has `generate_module_map = false`.
If we remove that line, the modulemap file will be generated. So my question now is, why not just turn on modulemap generation?

typfel: Actually my bad it was generated I'll delete it

crypto-ffi/bindings/swift/WireCoreCrypto/BuildSettings.xcconfig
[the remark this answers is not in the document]
typfel: I can I just remove the dots? 4.2.0 = 420 and spare me the hassle of parsing numbers and doing arithmetic

typfel:
> Do we have a mechanism that prevents using the low-level (`WireCoreCryptoUniffi`) part directly from client code?

Yes, it's using new functionality in Swift for `internal import`

https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md

crypto-ffi/bindings/swift/WireCoreCrypto/BuildSettings.xcconfig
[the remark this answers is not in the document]
istankovic: > I can I just remove the dots? 4.2.0 = 420 and spare me the hassle of parsing numbers and doing arithmetic

Yes, you can remove the dots, it can be an integer (https://developer.apple.com/documentation/xcode/build-settings-reference):
> This setting defines the current version of the project. The value must be a integer or floating point number, such as 57 or 365.8.

However, I wouldn't just concatenate the version like that because 420 may mean both `4.2.0` and `42.0`, which is why I suggested the formula above, making space for future releases. It's a common pattern in these scenarios.

.gitignore
```diff
@@ -7,11 +7,11 @@ crypto-ffi/bindings/*/out
...
-crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCrypto.swift
-crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/core_crypto.swift
-crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/core_cryptoFFI.h
-crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/core_crypto_ffiFFI.h
-crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/core_cryptoFFI.modulemap
+crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_crypto_ffi.swift
+crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_crypto.swift
+crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_cryptoFFI.h
+crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_crypto_ffiFFI.h
+crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_cryptoFFI.modulemap
```
istankovic, line 14: I think we should move this out of the root `.gitignore` and instead have one in `crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi` that ignores everything.

crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/WireCoreCryptoUniffi.h
```diff
@@ -0,0 +1,27 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+#import <Foundation/Foundation.h>
+
+//! Project version number for WireCoreCryptoUniffi.
+FOUNDATION_EXPORT double WireCoreCryptoUniffiVersionNumber;
+
+//! Project version string for WireCoreCryptoUniffi.
+FOUNDATION_EXPORT const unsigned char WireCoreCryptoUniffiVersionString[];
```
istankovic, line 25: Why do we need these two symbols? Is this being used by something outside this module?

typfel: This code is generated when creating a new ios framework.

istankovic:
> > Do we have a mechanism that prevents using the low-level (`WireCoreCryptoUniffi`) part directly from client code?
> 
> Yes, it's using new functionality in Swift for `internal import`
> 
> https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md

Oh, nice.

I wonder, how does that work with types? For example, even if we have the uniffi module imported via "internal import", does its `CoreCryptoLogger` become available? Reading the docs, I would guess no, but we do have that type as part of our public API, e.g the argument to `setLogger`.

istankovic:
> > What is the purpose of the `TestHost` directory and its contents?
> 
> I need to run the tests inside a real app (test host) because core crypto uses keychain APIs.

That is understandable, but could you please minimise the amount of code necessary to have a real app?
For instance, do we really need all of these?
```
./WireCoreCrypto/TestHost
./WireCoreCrypto/TestHost/Assets.xcassets
./WireCoreCrypto/TestHost/Assets.xcassets/AccentColor.colorset
./WireCoreCrypto/TestHost/Assets.xcassets/AccentColor.colorset/Contents.json
./WireCoreCrypto/TestHost/Assets.xcassets/AppIcon.appiconset
./WireCoreCrypto/TestHost/Assets.xcassets/AppIcon.appiconset/Contents.json
./WireCoreCrypto/TestHost/Assets.xcassets/Contents.json
./WireCoreCrypto/TestHost/ContentView.swift
./WireCoreCrypto/TestHost/Preview Content
./WireCoreCrypto/TestHost/Preview Content/Preview Assets.xcassets
./WireCoreCrypto/TestHost/Preview Content/Preview Assets.xcassets/Contents.json
./WireCoreCrypto/TestHost/TestHost.entitlements
./WireCoreCrypto/TestHost/TestHostApp.swift
```

We also have:
```swift
import SwiftUI
                                                                                                                                                                                                                                                           
struct ContentView: View {
    var body: some View {
        VStack {
            Image(systemName: "globe")
                .imageScale(.large)
                .foregroundStyle(.tint)
            Text("Hello, world!")
        }
        .padding()
    }
}
```
Would it not be possible to have an "empty" app, or a "background" app that's there only to give us access to keychain?

istankovic:
How can we run Swift wrapper tests from the command line?

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCryptoContext.swift
```diff
@@ -0,0 +1,999 @@
...
+        try await wrapError {
+            try await context.proteusFingerprintLocal(sessionId: sessionId)
+        }
+    }
+    
+    func proteusFingerprintPrekeybundle(prekey: Data) throws -> String {
+//        try await wrapError {
+            try context.proteusFingerprintPrekeybundle(prekey: prekey)
+//        }
+    }
```
istankovic, line 931: Why is this commented out?

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/Types.swift
```diff
@@ -0,0 +1,698 @@
...
+     */
+    case ciphertext = 2
+    
+    func lower() -> WireCoreCryptoUniffi.MlsWirePolicy {
+        switch self {
+        case .plaintext: .plaintext
+        case .ciphertext: .ciphertext
+        }
+    }
+}
```
istankovic, line 405: There are quite a few of cases like this. What is the reason we can't have a type alias to the internal type?
Missing docs?

The problem I see with this approach is that we're duplicating definitions (1 on the Rust side, 1 here).
With Wasm, some effort was done to deduplicate types, and the idea is to continue that way, i.e. have only a single type definition. So I would like us to find a better solution here.

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCryptoContext.swift
typfel: Mistake. It has to use a non-async error wrapper.

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/Types.swift
typfel: There's no other way if we don't want expose the `WireCoreCryptoUniffi` module. We have exactly the same problem and solution in the Kotlin bindings.

typfel:
> > > Do we have a mechanism that prevents using the low-level (`WireCoreCryptoUniffi`) part directly from client code?
> > 
> > 
> > Yes, it's using new functionality in Swift for `internal import`
> > https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md
> 
> Oh, nice.
> 
> I wonder, how does that work with types? For example, even if we have the uniffi module imported via "internal import", does its `CoreCryptoLogger` become available? Reading the docs, I would guess no, but we do have that type as part of our public API, e.g the argument to `setLogger`.

With an internal import no types and no classes from the internal uniffi module is allowed be part of the public API. All these these types needs to be wrapped

crypto-ffi/bindings/swift/WireCoreCrypto/BuildSettings.xcconfig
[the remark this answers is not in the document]
typfel: > However, I wouldn't just concatenate the version like that because 420 may mean both 4.2.0 and 42.0, which is why I suggested the formula above, making space for future releases. It's a common pattern in these scenarios.

Only if we leave out the patch component otherwise 42.0.0 = 4200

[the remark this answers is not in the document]
typfel: I'm not against what you're suggesting it's just that I think it gonna complicate the `update-versions.sh` script a lot.

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/Types.swift
```diff
@@ -0,0 +1,698 @@
...
+
+    // Default memberwise initializers are never public by default, so we
+    // declare one manually.
+    public init(
+        /**
+         * Whether this CRL modifies the old CRL (i.e. has a different revocated cert list)
+         */dirty: Bool,
+        /**
+         * Optional expiration timestamp
+         */expiration: UInt64?) {
```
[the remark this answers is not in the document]
typfel: Now I have

crypto-ffi/bindings/swift/WireCoreCrypto/BuildSettings.xcconfig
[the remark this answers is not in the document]
istankovic: > I'm not against what you're suggesting it's just that I think it gonna complicate the `update-versions.sh` script a lot.

I don't see how, really, for example this should do it:
```sh
IFS='.' read -r major minor patch <<< $new_version
CURRENT_PROJECT_VERSION=$((10000 * $major + 100 * $minor + $patch))
```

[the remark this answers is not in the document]
typfel: Nice, ok I'll use this

typfel:
> > > What is the purpose of the `TestHost` directory and its contents?
> > 
> > 
> > I need to run the tests inside a real app (test host) because core crypto uses keychain APIs.
> 
> That is understandable, but could you please minimise the amount of code necessary to have a real app? For instance, do we really need all of these?
> 
> ```
> ./WireCoreCrypto/TestHost
> ./WireCoreCrypto/TestHost/Assets.xcassets
> ./WireCoreCrypto/TestHost/Assets.xcassets/AccentColor.colorset
> ./WireCoreCrypto/TestHost/Assets.xcassets/AccentColor.colorset/Contents.json
> ./WireCoreCrypto/TestHost/Assets.xcassets/AppIcon.appiconset
> ./WireCoreCrypto/TestHost/Assets.xcassets/AppIcon.appiconset/Contents.json
> ./WireCoreCrypto/TestHost/Assets.xcassets/Contents.json
> ./WireCoreCrypto/TestHost/ContentView.swift
> ./WireCoreCrypto/TestHost/Preview Content
> ./WireCoreCrypto/TestHost/Preview Content/Preview Assets.xcassets
> ./WireCoreCrypto/TestHost/Preview Content/Preview Assets.xcassets/Contents.json
> ./WireCoreCrypto/TestHost/TestHost.entitlements
> ./WireCoreCrypto/TestHost/TestHostApp.swift
> ```
> 
> We also have:
> 
> ```swift
> import SwiftUI
>                                                                                                                                                                                                                                                            
> struct ContentView: View {
>     var body: some View {
>         VStack {
>             Image(systemName: "globe")
>                 .imageScale(.large)
>                 .foregroundStyle(.tint)
>             Text("Hello, world!")
>         }
>         .padding()
>     }
> }
> ```
> 
> Would it not be possible to have an "empty" app, or a "background" app that's there only to give us access to keychain?

So I deleted the `Asset` and `Preview` they aren't strictly necessary,.

istankovic:
While we are adding this, we should also add a check to CI that makes sure that the code is formatted according to swiftformat rules.

istankovic:
> > While we are adding this, we should also add a check to CI that makes sure that the code is formatted according to swiftformat rules.
> 
> Didn't we schedule that in separate ticket?

Ah yes. :+1:

istankovic:
The CI should make sure to run Swift tests.

SimonThormeyer (approved):
Good to go once the implementors of protocols have consistent names and once we know if we can avoid the type wrappers via re-exporting single items (don't know the correct terminology)

crypto-ffi/bindings/swift/BuildSettings.xcconfig
```diff
@@ -19,4 +19,7 @@
 // Configuration settings file format documentation can be found at:
 // https://help.apple.com/xcode/#/dev745c5c974
 
-CURRENT_PROJECT_VERSION = 11.0.0
+MARKETING_VERSION=4.2.0
+CURRENT_PROJECT_VERSION=1.0
```
istankovic, line 23: `CURRENT_PROJECT_VERSION` needs to be in sync with `MARKETING_VERSION`, per the formula in `update-versions.sh`.

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto.xcodeproj/project.pbxproj
```diff
@@ -242,110 +524,119 @@
 				CLANG_WARN_UNREACHABLE_CODE = YES;
 				CLANG_WARN__DUPLICATE_METHOD_MATCH = YES;
 				COPY_PHASE_STRIP = NO;
-				DEAD_CODE_STRIPPING = YES;
+				CURRENT_PROJECT_VERSION = 1;
```
istankovic, line 527: There's also this instance of `CURRENT_PROJECT_VERSION`, as well as some others -- we should make sure they are all in sync.

crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,379 @@
...
+		};
+		16CDD11C2A2636BB0073FB8E /* Release */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
+				CLANG_ENABLE_MODULES = YES;
+				CODE_SIGN_STYLE = Automatic;
+				DEAD_CODE_STRIPPING = YES;
+				DEFINES_MODULE = YES;
+				DYLIB_COMPATIBILITY_VERSION = 1;
+				DYLIB_CURRENT_VERSION = 1;
```
istankovic, line 321: How are these variables used?

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto.xcodeproj/project.pbxproj
typfel: A nice find. This shouldn't be here since it overrides what we define in in `BuildSettings.xcconfig`

crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi.xcodeproj/project.pbxproj
typfel: From what I can tell these are used to specify compatibllity when publishing dynamic libraries which only a thing on the Mac so this doesn't really apply to us. The values are here defined here because they have a default value in all new projects.

https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryDesignGuidelines.html#//apple_ref/doc/uid/TP40002013

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto.xcodeproj/project.pbxproj
istankovic: These are all reference to `CURRENT_PROJECT_VERSION`:
```
crypto-ffi/bindings/swift/BuildSettings.xcconfig:23:CURRENT_PROJECT_VERSION=40202
crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto.xcodeproj/project.pbxproj:555:                          CURRENT_PROJECT_VERSION = 1;
crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto.xcodeproj/project.pbxproj:573:                          CURRENT_PROJECT_VERSION = 1;
crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto.xcodeproj/project.pbxproj:593:                          CURRENT_PROJECT_VERSION = 1;
crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto.xcodeproj/project.pbxproj:622:                          CURRENT_PROJECT_VERSION = 1;
```

Shouldn't we also adjust the `update-versions.sh` script to take care of all of these?

typfel: No, we should only update the version in `BuildSettings.xcconfig` which takes precedence over the default values in the project.

crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi.xcodeproj/project.pbxproj
typfel: I know assign the value of  `CURRENT_PROJECT_VERSION` to `DYLIB_COMPATIBILITY_VERSION ` and `DYLIB_CURRENT_VERSION `.

istankovic:
Commit `c5112a3653f79ea7b707faab0c542ded1f952329`, which remove a lot of the code that is being added by `bdec05b42ac03a8de93e9b090717d97b8add16b7`, should be merged with commit `bdec05b42ac03a8de93e9b090717d97b8add16b7`. That way we avoid the intermediate state of adding ~3k lines.

typfel:
> Commit `c5112a3653f79ea7b707faab0c542ded1f952329`, which removes a lot of the code that is being added by `bdec05b42ac03a8de93e9b090717d97b8add16b7`, should be merged with commit `bdec05b42ac03a8de93e9b090717d97b8add16b7`. That way we avoid the intermediate state of adding ~3k lines.

I wanted to keep them in the history if we ever wanted recover the type wrappers

istankovic:
> > Commit `c5112a3653f79ea7b707faab0c542ded1f952329`, which removes a lot of the code that is being added by `bdec05b42ac03a8de93e9b090717d97b8add16b7`, should be merged with commit `bdec05b42ac03a8de93e9b090717d97b8add16b7`. That way we avoid the intermediate state of adding ~3k lines.
> 
> I wanted to keep them in the history if we ever wanted recover the type wrappers

In that case I would suggest to create a new branch that points to those commits.

.gitignore
```diff
@@ -7,11 +7,6 @@ crypto-ffi/bindings/*/out
 crypto-ffi/bindings/*/.build
 crypto-ffi/bindings/*/build
 crypto-ffi/bindings/*/.swiftpm
-crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_crypto_ffi.swift
-crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_crypto.swift
-crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_cryptoFFI.h
-crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_crypto_ffiFFI.h
-crypto-ffi/bindings/swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi/core_cryptoFFI.modulemap
```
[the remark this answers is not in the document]
typfel: In the first commit the path was different so it wouldn't be canceled out. Instead it would make the two commits  depend on each other which IMO is confusing.

istankovic (approved):

istankovic:
:fireworks:

PR #977, opened 2025-03-14
refactor: uniffy `generic` and `wasm` modules [WPB-16252]

coriolinus:
# What's new in this PR

Refactor in parts to a single module tree with only the differences which are absolutely required. Use conditional compilation to the greatest degree possible to make everything work properly.

----

- The **PR Title**


coriolinus:
After a week working on WASM, there are two major takeaways:

- duck typed objects are not and cannot be trait objects; you can't retain them past the lifetime of the function
- we do not need to change the public interface of the high-level wrapper at all, at least on wasm

crypto-ffi/src/metadata.rs
```diff
@@ -0,0 +1,113 @@
...
+        opt_level: core_crypto::BUILD_METADATA.opt_level.into(),
+        #[cfg_attr(target_family = "wasm", expect(clippy::useless_conversion))]
+        target_triple: core_crypto::BUILD_METADATA.target_triple.into(),
+        #[cfg_attr(target_family = "wasm", expect(clippy::useless_conversion))]
+        git_branch: core_crypto::BUILD_METADATA.git_branch.into(),
+        #[cfg_attr(target_family = "wasm", expect(clippy::useless_conversion))]
+        git_describe: core_crypto::BUILD_METADATA.git_describe.into(),
+        #[cfg_attr(target_family = "wasm", expect(clippy::useless_conversion))]
+        git_sha: core_crypto::BUILD_METADATA.git_sha.into(),
+        #[cfg_attr(target_family = "wasm", expect(clippy::useless_conversion))]
```
istankovic, line 97: Instead of adding this attribute to every member, we can add it to the whole block, i.e.
```rust
    #[cfg_attr(target_family = "wasm", expect(clippy::useless_conversion))]
    BuildMetadata {
        timestamp: core_crypto::BUILD_METADATA.timestamp.into(),
        ...
```

crypto-ffi/src/e2ei/enrollment.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /// See [core_crypto::e2e_identity::E2eiEnrollment::new_account_response]
+    pub async fn new_account_response(&self, account: Vec<u8>) -> CoreCryptoResult<()> {
+        self.write().await.new_account_response(account).map_err(Into::into)
+    }
+
+    /// See [core_crypto::e2e_identity::E2eiEnrollment::new_order_request]
+    #[allow(clippy::too_many_arguments)]
+    pub async fn new_order_request(&self, previous_nonce: String) -> CoreCryptoResult<Vec<u8>> {
```
SimonThormeyer, line 62: Do we still need this? If so, we should change it to `[expect()]`

crypto-ffi/src/e2ei/enrollment.rs
```diff
@@ -0,0 +1,221 @@
...
+        self.write()
+            .await
+            .new_authz_response(authz)
+            .map(Into::into)
+            .map_err(Into::into)
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    /// See [core_crypto::e2e_identity::E2eiEnrollment::create_dpop_token]
+    pub async fn create_dpop_token(&self, expiry_secs: u32, backend_nonce: String) -> CoreCryptoResult<String> {
```
SimonThormeyer, line 94: See above, if it's still needed, change to `#[expect(clippy::too_many_arguments)]`

crypto-ffi/src/core_crypto/command/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+pub trait CoreCryptoCommand: Send + Sync {
+    /// Will be called inside a transaction in CoreCrypto
+    async fn execute(&self, context: Arc<CoreCryptoContext>) -> CoreCryptoResult<()>;
+}
+
+#[cfg(not(target_family = "wasm"))]
+#[async_trait::async_trait]
+impl<F, Fut> CoreCryptoCommand for F
+where
+    F: Fn(Arc<CoreCryptoContext>) -> Fut + Send + Sync,
```
SimonThormeyer, line 22: Is it possible to use the `AsyncFn` trait here and spare the extra generic parameter for the future?

crypto-ffi/src/credential_type.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+impl From<CredentialType> for core_crypto::prelude::MlsCredentialType {
+    fn from(value: CredentialType) -> core_crypto::prelude::MlsCredentialType {
+        match value {
+            CredentialType::Basic => core_crypto::prelude::MlsCredentialType::Basic,
+            CredentialType::X509 => core_crypto::prelude::MlsCredentialType::X509,
+        }
+    }
+}
```
istankovic, line 32: I'm not sure how many instances of these bidirectional mappings we have across CC, but I have a feeling we saw a few. Not worth it for just this particular case, but if it turns out we have many, we could consider writing a macro like the one at https://github.com/Giftzwerg02/bidirectional_enum/blob/main/src/lib.rs.

crypto-ffi/src/core_crypto/command/mod.rs
coriolinus: The short answer is no. I don't remember what errors came up, but I had the same idea; it did not work out.

interop/src/clients/corecrypto/ffi.rs
```diff
@@ -3,7 +3,7 @@ use crate::{
     clients::{EmulatedClient, EmulatedClientProtocol, EmulatedClientType, EmulatedMlsClient},
 };
 use color_eyre::eyre::Result;
-use core_crypto_ffi::{ClientId, CoreCrypto, CustomConfiguration, CredentialType, context::TransactionHelper};
+use core_crypto_ffi::{ClientId, CoreCrypto, CredentialType, CustomConfiguration, context::TransactionHelper};
```
SimonThormeyer: Please squash 122664364d35a11d9918d6510429350cf9086d14 into the commit that introduced the need for this formatting change

crypto-ffi/src/decrypted_message.rs
```diff
@@ -0,0 +1,119 @@
...
+            sender_client_id: from.sender_client_id.map(ClientId),
+            has_epoch_changed: from.has_epoch_changed,
+            identity: from.identity.into(),
+            buffered_messages,
+            crl_new_distribution_points: from.crl_new_distribution_points.into(),
+        })
+    }
+}
+
+/// to avoid recursive structs
```
istankovic, line 80: Could you please expand this comment? I am not sure what it refers to.

istankovic:
A number of dependencies were update as part of `5951403a828a1c2971d2fec909ce9df97eca08ac`. Could you please remove dependency changes from the commit?

crypto-ffi/src/decrypted_message.rs
coriolinus: Just means that `BufferedDecryptedMessage` is identical to `DecryptedMessage` except for the addition of the `buffered_messages` field.

crypto-ffi/bindings/js/src/CoreCryptoInstance.ts
```diff
@@ -315,14 +317,29 @@ export class CoreCrypto {
...
+     * outstanding references that were detected.
+     *
+     * As always with this kind of thing, beware TOCTOU.
+     *
+     * @returns whether the CoreCrypto instance can currently close.
+     */
+    async canClose(): Promise<boolean> {
+        return await this.#cc.can_close();
+    }
+
```
SimonThormeyer, line 353: Can you explain the need for this new API? Is it just to enable a defensive handling of a potential error produced by `CoreCrypto.close()`?

istankovic:
Commit 
```
commit cf3f88d0f11fb4d677e3a0fdd215f90ae018b896
Refs: v5.0.0-123-gcf3f88d0
Author:     Peter Goodspeed-Niklaus <peter.goodspeed-niklaus@wire.com>
AuthorDate: Tue Apr 1 12:07:51 2025 +0200
Commit:     Peter Goodspeed-Niklaus <peter.goodspeed-niklaus@wire.com>
CommitDate: Wed Apr 2 21:49:26 2025 +0200

    fix(ffi/js): fix ffi incompatibilities in high-level ts wrapper

    "Fix" here is defined as "builds cleanly"; I have not yet attempted
    to make the tests actually work.
```
has too many, unrelated changes. Could you please split them up, and squash the pieces with their original changes, where it makes sense?
For instance,
```
+ #[cfg_attr(target_family = "wasm", wasm_bindgen(readonly))]
```
should have been retained while moving the original wasm code into the unified module.

crypto-ffi/src/error/wasm.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub(crate) fn generic<E>() -> impl FnOnce(E) -> Self
+    where
+        E: ToString,
+    {
+        |err| Self(InternalError::generic()(err))
+    }
+
+    pub(crate) fn ad_hoc(err: impl ToString) -> Self {
+        Self(InternalError::ad_hoc(err))
+    }
```
istankovic, line 65: I like the naming here! :smiley:

SimonThormeyer (approved):
Awesome work! 
Approved assuming my comments are going to resolve!

coriolinus:
> Could you please remove dependency changes from the commit?

Not really. What I _needed_ was to upgrade `wasm-bindgen` from `0.2.99` to `0.2.100`. That had a handful of changes. Then at some point during a rebase onto main, rustc got confused about what versions were selected and decided that it couldn't choose a compatible version for everything. After half an hour or so failing to find a manual fix, `cargo update` solved things. But I am very reluctant to dive back into that mess to manually search for a smaller upgrade set.

The key feature in wasm-bindgen v0.2.100 that I could not do without, by the way, was the ability to put its macro attributes into `cfg_attr` annotations. Not something that we could skip for this PR!

crypto-ffi/bindings/js/src/CoreCryptoInstance.ts
coriolinus: Yeah, I had to deprecate `isLocked` and remove its implementation because it wasn't async. (`isLocked` was susceptible to exactly the same TOCTOU issues as `canClose`, of course!) Unifying the implementations meant that we couldn't just check anymore if there was only one strong reference to an outer Arc. 

I'm fine dropping this API later, in a separate PR. But my motivating goal in this PR was to minimize the client-visible changes. In the case where I drop a client-visible API, that meant adding an easy migration path.

crypto-ffi/bindings/js/test/client_identity.test.ts
```diff
@@ -0,0 +1,44 @@
+import { browser, expect } from "@wdio/globals";
```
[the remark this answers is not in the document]
coriolinus: Huh. Can do, of course, but while the typescript book [agrees with you](https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md#filename), [the Google TS style guide does not](https://google.github.io/styleguide/tsguide.html#identifiers-imports).

I'll make the change.

crypto-ffi/src/metadata.rs
coriolinus: While it's possible to add it to the block, that unfortunately doesn't actually do what we want; clippy complains about an unfulfilled lint expectation in that case.

crypto-ffi/src/e2ei/enrollment.rs
coriolinus: Right, good catch.

crypto-ffi/bindings/js/test/client_identity.test.ts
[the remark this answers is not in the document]
SimonThormeyer: I didn't know this was controversial.

crypto-ffi/src/credential_type.rs
coriolinus: Mostly we don't actually want bidirectional types: in general, our types get used exclusively as parameters or as return values from CoreCrypto; only rarely are they both. And we only actually need bidi conversions when they are both. I don't think I'd write a macro for this.

crypto-ffi/src/e2ei/enrollment.rs
```diff
@@ -0,0 +1,219 @@
...
+            .await
+            .new_oidc_challenge_response(&provider, challenge)
+            .await
+            .map_err(Into::into)
+    }
+
+    /// See [core_crypto::prelude::E2eiEnrollment::get_refresh_token]
+    //
+    // Note: this function does not exist in WASM!
+    pub async fn get_refresh_token(&self) -> CoreCryptoResult<String> {
```
istankovic, line 212: Why doesn't this function exist in WASM?

crypto-ffi/bindings/js/test/client_identity.test.ts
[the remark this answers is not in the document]
coriolinus: It's not even standardized in this repo! in `bindings/js/src`, auto-generated js stuff is in snake_case files, and human-written stuff is in PascalCase files.

crypto-ffi/src/e2ei/enrollment.rs
istankovic: Obligatory :woozy_face:

istankovic:
This was herculean task, great job! :tada:

istankovic (approved):

PR #996, opened 2025-03-26
ci: fix macos runner rust setup and cache

SimonThormeyer:
# What's new in this PR

----

- The **PR Title**


coriolinus (approved):
Make issues to use the new composite action elsewhere, I guess?

SimonThormeyer:
> Make issues to use the new composite action elsewhere, I guess?

Not sure if that would improve the situation in other places in this case, we can keep using the setup toolchain action as-is elsewhere. We would replace a single action invocation by another single action invocation. It's just on our self-hosted runners where we need to run it without caching, then run the cleanup, and then the caching separately.

But I do have the feeling that there are other places where we could make good use of composite actions...

PR #997, opened 2025-03-26
chore: move to new eslint config [WPB-15105]

istankovic:
We're using a deprecated eslint config file. Move to the new version.


SimonThormeyer (changes requested):
Let's use `.ts` for the config file

SimonThormeyer (approved):

PR #1021, opened 2025-04-01
ci: fix cache key prefixes for interop test

SimonThormeyer:
The rust target based cache keys don't work in cases where we have multiple targets because cache keys can't contain commas. This adds a custom cache key parameter to our action for the only such case we currently have.

# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/.github/actions/setup-and-cache-rust/action.yml
+++ b/.github/actions/setup-and-cache-rust/action.yml
@@ -12,6 +12,10 @@ inputs:
     description: Additional components to install that don't come with the toolchain by default.
     required: false
     default: ''
+  cache-key-prefix:
+    description: A custom cache key prefix to be used instead of the rust target. Must be used if there are
+      multiple rust targets because keys cannot contain commas.
+    required: false
 runs:
   using: composite
   steps:
@@ -35,4 +39,4 @@ runs:
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
           target/
-        key: ${{ inputs.target }}-${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+        key: ${{ inputs.cache-key-prefix || format('{0}', inputs.target) }}-${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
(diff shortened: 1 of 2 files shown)


.github/actions/setup-and-cache-rust/action.yml
```diff
@@ -12,6 +12,9 @@ inputs:
     description: Additional components to install that don't come with the toolchain by default.
     required: false
     default: ''
+  cache-key-prefix:
+    description: A custom cache key prefix to be used instead of the rust target. Must be used if there are multiple rust targets because keys cannot contain commas.
```
istankovic, line 16: Nit: I think splitting this line would look nicer -- no reason to have ~160 character lines.

.github/workflows/bindings.yml
```diff
@@ -159,6 +159,7 @@ jobs:
         with:
           target: "wasm32-unknown-unknown,aarch64-apple-ios,aarch64-apple-ios-sim"
           rustflags: ''
+          cache-key-prefix: e2e-interop-test
```
istankovic, line 162: This means the interop test won't share cache with any other, right?

.github/actions/setup-and-cache-rust/action.yml
```diff
@@ -35,4 +38,4 @@ runs:
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
           target/
-        key: ${{ inputs.target }}-${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+        key: ${{ inputs.cache-key-prefix || format('{0}', inputs.target) }}-${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
istankovic, line 41: Thinking more about this, why does the cache key depend on the target(s)? Shouldn't it be enough to depend on `runner.os`?

SimonThormeyer: Unfortunately not, see f3596ad79a0c646b0bf2ff3d642d5689f31ccd48

istankovic (approved):

PR #1033, opened 2025-04-04
refactor: centralize mls client and rename to `Session` [WPB-15533]

coriolinus:
# What's new in this PR

- rename `mls::Client` to `Session`
- restructure module tree appropriately
- rename relevant helper structs
- rename relevant helper traits
- ensure all `Session` impls are in `session` tree

However, I intentionally decided not to go through and rename all locals of these types. That just felt like quite a lot of churn. I'm willing to be convinced otherwise though.

----

- The **PR Title**


crypto/src/mls/session/mod.rs
```diff
@@ -35,15 +35,18 @@ use std::sync::Arc;
...
+/// A MLS Session enables a user device to communicate via the MLS protocol.
 ///
-/// It can be the Android, iOS, web or desktop application which the authenticated user is using.
-/// A user has many client, a client has only one user.
-/// A client can belong to many MLS groups
+/// This closely maps to the `Client` term in [RFC 9720], but we avoid that term to avoid ambiguity;
+/// `Client` is very overloaded with distinct meanings.
 ///
-/// It is cheap to clone a `Client` because everything heavy is wrapped inside an [Arc].
+/// There is one `Session` per user per device. A session can contain many MLS groups/conversations.
```
istankovic, line 49: > There is one `Session` per user per device.

Hm, but isn't it possible to have, say, 3 different `Session` instances, having their own databases, that all run on the same device and use the same user?

crypto/src/mls/session/mod.rs
```diff
@@ -48,7 +48,7 @@ use tls_codec::{Deserialize, Serialize};
 #[derive(Clone, Debug)]
 pub struct Session {
     pub(crate) state: Arc<RwLock<Option<ClientInner>>>,
-    pub(crate) mls_backend: MlsCryptoProvider,
+    pub(crate) crypto_provider: MlsCryptoProvider,
```
istankovic, line 57: Thank you for this, that was very much needed! :purple_heart:

crypto/src/mls/session/mod.rs
coriolinus: I'm abstracting a little. "User" here is authenticated user id; if you have two usernames, you can have additional sessions. "Device" is in the MLS sense of an installation; you can have two Wire apps installed and each is a different device.

That complication didn't seem worth mentioning here though. Do you think it's worth going into that detail?

crypto/src/mls/conversation/merge.rs
```diff
@@ -107,7 +107,7 @@ mod tests {
                         .await
                         .commit_accepted(
                             &alice_central.context.client().await.unwrap(),
-                            &alice_central.client.mls_backend,
+                            &alice_central.client.crypto_provider,
```
istankovic: Nit: this hunk should have gone in the preceding commit.

crypto/src/mls/mod.rs
```diff
@@ -145,7 +145,7 @@ impl HasClientAndProvider for Session {
     }
 
     async fn mls_provider(&self) -> Result<MlsCryptoProvider> {
-        Ok(self.mls_backend.clone())
+        Ok(self.crypto_provider.clone())
```
istankovic: Nit: this hunk should have gone in the preceding commit.

istankovic (approved):

crypto/src/transaction_context/mod.rs
```diff
@@ -60,15 +60,15 @@ impl CoreCrypto {
 
 #[cfg_attr(target_family = "wasm", async_trait::async_trait(?Send))]
 #[cfg_attr(not(target_family = "wasm"), async_trait::async_trait)]
-impl HasClientAndProvider for TransactionContext {
-    async fn client(&self) -> crate::mls::Result<Session> {
+impl HasSessionAndCrypto for TransactionContext {
+    async fn session(&self) -> crate::mls::Result<Session> {
         self.mls_client()
```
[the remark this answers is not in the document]
coriolinus: Oh, good catch! I'll do that.

crypto/src/mls/session/mod.rs
istankovic: Probably not, but I find the comment confusing because I can imagine, in theory, an app that has one session that uses an on-disk database, and another session that uses in-memory database, with the exact same user and device IDs. The comment implies that would not be possible, but I'm not sure I understand why.

PR #1059, opened 2025-04-16
refactor: conclude centralization of `TransactionContext` impls [WPB-15535]

SimonThormeyer:
# What's new in this PR
See above

----

- The **PR Title**


crypto-ffi/src/error/core_crypto.rs
```diff
@@ -107,6 +107,13 @@ impl From<RecursiveError> for CoreCryptoError {
             core_crypto::mls::conversation::Error::BufferedCommit => MlsError::BufferedCommit.into(),
             core_crypto::mls::conversation::Error::MessageRejected { reason } => MlsError::MessageRejected { reason: reason.clone() }.into(),
             core_crypto::mls::conversation::Error::OrphanWelcome => MlsError::OrphanWelcome.into(),
+            e @ (
```
coriolinus, line 141: Do we need the `e` binding? It looks kind of like we could just do `ProteusError::from(innermost)_.into()` and have it work the same.

crypto-ffi/src/error/core_crypto.rs
```diff
@@ -67,25 +67,55 @@ impl From<RecursiveError> for CoreCryptoError {
...
+            ) => {
                 match ($val) {
                     $pattern $(if $guard)? => Some($out),
                     _ => None,
                 }
             };
+            // With cfg attribute
+            (
+                $val:expr,
+                #[cfg($meta:meta)]
```
coriolinus, line 83: I approve of the restraint here, not expanding an arbitrary `$( #[$meta:meta] )*`; YAGNI.

coriolinus (approved):
Nice job on the macros!

crypto-ffi/src/error/core_crypto.rs
SimonThormeyer: Hmm, to me it looks like we do need it:
`innermost` is of type `&dyn Error`, so we cannot call `ProteusError::from()` on `innermost`.
But we can call it on `e` in that match arm binding, which is of type  `&ProteusErrorKind`, and we have 
but we only have `From<&core_crypto::ProteusErrorKind>` implemented for  `ProteusError`.

PR #1083, opened 2025-04-24
chore: add Kotlin linter checks (WPB-15467)

istankovic:
And fix all the issues pointed out by ktlint.


.github/workflows/kotlin.yml
```diff
@@ -0,0 +1,24 @@
+name: kotlin
+
+concurrency:
```
typfel, line 3: So if this check fails does it cancel all other checks?

istankovic: No, it just means that if you push an update to the same ref, and there's already a run in progress for that ref, that run will be canceled. This is to avoid having multiple CI runs, where most of them would test the old state of the ref.

For example: 
```
push commit X to branch foo
<github starts a new run (R1) for brach foo, based on state X>
push commit Y to branch foo
<github sees the above and that run R1 is still active so it cancel R1 and starts a new run (R2) based on state Y
```
```

File: crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/GeneralTest.kt [hunk 2]
```diff
@@ -20,11 +22,11 @@ class GeneralTest {
...
     fun get_build_metadata() = runTest {
-        val build_metadata = buildMetadata()
+        val buildmetadata = buildMetadata()
 
-        assertThat(build_metadata).isNotNull()
-        assertThat(build_metadata.gitDescribe).isNotNull()
-        assertThat(build_metadata.gitDescribe).isNotEmpty()
+        assertThat(buildMetadata).isNotNull()
+        assertThat(buildMetadata.gitDescribe).isNotNull()
+        assertThat(buildMetadata.gitDescribe).isNotEmpty()
```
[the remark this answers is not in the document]
istankovic: Yeah, that doesn't work, Fixed. :+1:

coriolinus (approved):

PR #1086, opened 2025-04-25
chore: drop sysinfo

istankovic:
sysinfo is currently only used to get available RAM on non-wasm platforms, but we really don't need that, and the information is not so useful anyway. On the other hand, sysinfo is another dependency that brings with it a few windows dependencies, which we don't really have a need for.


coriolinus (approved):
No objections, given that we can always override this in the caller anyway. Pending green CI of course.

PR #1090, opened 2025-04-29
chore: update chrono

istankovic:
Also drop the clock feature which we're not really using, which allow us to drop a couple of dependencies.


coriolinus (approved):
If everything still works, then getting rid of the clock feature can only be a good thing.

PR #1092, opened 2025-04-29
feat: add ephemeral cc client [WPB-17097]

coriolinus:
# What's new in this PR

- [x] `struct HistorySecret`
- [x] `fn generate_history_secret`
- [ ] `fn history_client(history_secret) -> CoreCrypto`
- [ ] implementation in `crypto-ffi`
- [ ] implementation in wasm
- [ ] implementation in kotlin
- [ ] implementation in swift

----

- The **PR Title**


crypto/src/ephemeral.rs
```diff
@@ -0,0 +1,32 @@
...
+//! not be used to generate messages for sending. They should also not be instantiated to follow along with
+//! new messages as they are received, as that's pointless; the individual credentials suffice.
+
+use openmls::prelude::Credential;
+
+/// A `HistorySecret` encodes sufficient client state that it can be used to instantiate an
+/// ephemeral client.
+#[derive(serde::Serialize, serde::Deserialize)]
+pub struct HistorySecret {
+    client_id: String,
```
SimonThormeyer: I've seen a `ClientId` type in other places, can we use it here, too?

coriolinus: I didn't at first because it didn't implement `Serialize`, `Deserialize`, but on reflection you're right: better to just add those methods and use the proper type here.

crypto/src/ephemeral.rs
```diff
@@ -20,13 +20,74 @@
...
+    key_package: KeyPackage,
+}
+
+impl CoreCrypto {
+    /// Generate a new [`HistorySecret`].
+    ///
+    /// This is useful when it's this client's turn to generate a new history client.
+    ///
+    /// The generated secret is cryptographically unrelated to the current CoreCrypto client.
+    pub async fn generate_history_secret(&self, conversation_id: &ConversationId) -> Result<HistorySecret> {
```
SimonThormeyer, line 49: I see there is a `ConversationId` as a parameter; If we can, we should make this a method on the `ImmutableConversation` or `ConversationGuard`?

crypto/src/ephemeral.rs
```diff
@@ -0,0 +1,227 @@
...
+pub struct HistorySecret {
+    pub(crate) client_id: ClientId,
+    pub(crate) credential_bundle: CredentialBundle,
+    pub(crate) key_package: KeyPackageSecretEncapsulation,
+}
+
+/// Create a new [`CoreCrypto`] with an **uninitialized** mls session.
+///
+/// You must initialize the session yourself before using this!
+async fn in_memory_cc_with_ciphersuite(ciphersuite: impl Into<MlsCiphersuite>) -> Result<CoreCrypto> {
```
SimonThormeyer, line 50: Just wondering, is it in general a good pattern to have a generic parameter with a type like `impl Into<MlsCiphersuite>` instead of letting the caller make the `.into()` call?

coriolinus: It's situational. If you have a handful of different types you'd want to call it with, then accepting an `impl Into` parameter saves a little boilerplate. If you have hundreds of different types you'd want to call it with, then there might be monomorphization issues. 

Here, I think there are exactly two types we use; I think it's fine for the very slight simplification on the caller side.

crypto/src/ephemeral.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    use super::*;
+
+    /// Create a history secret, and restore it into a CoreCrypto instance
+    #[apply(all_cred_cipher)]
+    #[wasm_bindgen_test]
+    async fn can_create_ephemeral_client(case: TestContext) {
+        if case.credential_type != MlsCredentialType::Basic {
+            // history client will only ever have basic credentials, so not much point in testing
+            // how it interacts with an x509-only conversation
```
SimonThormeyer, line 183: But it does work?

SimonThormeyer (commented):
Great work! This looks good, but I'd like to discuss some things first.

crypto/src/ephemeral.rs
```diff
@@ -0,0 +1,227 @@
...
+        client_id,
+        credential_bundle,
+        key_package,
+    })
+}
+
+impl CoreCrypto {
+    /// Instantiate a history client.
+    ///
+    /// This client exposes the full interface of `CoreCrypto`, but it should only be used to decrypt messages.
```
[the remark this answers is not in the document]
coriolinus: Does this new type expose transactions? It should, for the batching properties. So now we have both `CoreCryptoEphemeralWrapper` and `TransactionContextEphemeralWrapper`, both of which exist only to reduce the number of methods in their implementations. We could go there I suppose, but it kind of feels like overkill.

Also, that would mean the current test case wouldn't work.

crypto/src/ephemeral.rs
SimonThormeyer: I think it's your test setup or the test function you're calling that might not be the best fit, but I think we should test this case. Check `heterogeneous_clients_can_send_messages()` for a setup where an `x509` session talks to a `Basic` session

SimonThormeyer (approved):
Great!

PR #1109, opened 2025-05-07
chore: remove ProposalBundle type from bindings (WPB-15373)

istankovic:
This type is not used anymore, since MlsTransport was added.


coriolinus:
Are you not removing it from Swift?

istankovic:
> Are you not removing it from Swift?

Interestingly, there's no `ProposalBundle` type in Swift bindings.

coriolinus (approved):

PR #939, opened 2025-02-28
chore: disable logging on CI which is very verbose

typfel:
# What's new in this PR

Disable logging on CI which is very verbose.

----
##### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #1120, opened 2025-05-09
chore: drop femme and reduce noise in CI interop test

istankovic:
This PR

- refactors the interop client setup a bit
- disables the iOS client on non-Mac platforms so we can run interop again on Windows and Linux
- drops femme in favor of env_logger

The last point also addresses the large output issue when running interop on CI. In that sense this PR obsoletes https://github.com/wireapp/core-crypto/pull/939.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1516,22 +1516,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e8c02a5121d4ea3eb16a80748c74f5549a5665e4c21333c6098f283870fbdea6"
 
 [[package]]
-name = "femme"
-version = "2.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc04871e5ae3aa2952d552dae6b291b3099723bf779a8054281c1366a54613ef"
-dependencies = [
- "cfg-if",
- "js-sys",
- "log",
- "serde",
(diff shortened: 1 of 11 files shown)


coriolinus (commented):
Looks good in principle, and can't complain about using more standard tools and processes. Of course CI needs to go green. But I was curious about two other things:

- why does the commit chain start with two dependency version-bumps?
- How specifically does this obsolete #939? Do we still need to adjust the environment to reduce the log output?

istankovic:
> Looks good in principle, and can't complain about using more standard tools and processes. Of course CI needs to go green. But I was curious about two other things:
> 
>     * why does the commit chain start with two dependency version-bumps?

That was because I was basing this PR on top of another one, which was merged meanwhile. Now this should be based right on top of latest `main`.

>     * How specifically does this obsolete [chore: disable logging on CI which is very verbose #939](https://github.com/wireapp/core-crypto/pull/939)? Do we still need to adjust the environment to reduce the log output?

femme seems to produce a lot of output by default (looks the default behaviour is quite different from that of `env_logger`), most of the noise was coming form the database crate. I remember @typfel mentioning that we'd still want to log _something_, but I'm not sure exactly what. It should be quite easy to configure this via setting `RUST_LOG` in CI appropriately, though.

coriolinus (approved):

PR #1129, opened 2025-05-13
fix(kotlin): prevent cancellations during transactions [WPB-17522]

vitorhugods:
# The issue

When doing a CC transaction from Kotlin, and the transaction is cancelled from the outside, it seems that the code never resumes, and future transactions are never successfully started.

### Example

The Android app sends pending proposals when the app is considered _online/in sync_.
When the app receives a push notification, the app syncs everything — becoming _in sync_ for a short time, and quickly disconnects after showing the user that there are new notifications.

When the app disconnects, the pending proposals sending is cancelled.
Because pending proposals have network calls within the transactions, they are quite likely to be cancelled midway through.

Future transactions are _never_ happening.

# The cause

Uniffi [doesn't support cancellations](https://mozilla.github.io/uniffi-rs/latest/futures.html#cancelling-async-code).

I couldn't wrap my head completely around the generated code to understand why the code doesn't resume, but it's clear that the transaction lock within CC is never released and future attempts require infinite patience.

# The fix

The quickest and easiest solution is to make transactions [NonCancellable](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-non-cancellable/).
I think this is a fair compromise.

For now, I fail to see the need for any cancellation within a transaction. The only reason Android apps are cancelling a transaction at the moment is because we try to cancel operations like

----

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] contains a reference JIRA issue number like `SQPIT-764`
  - [X] answers the question: _If merged, this PR will: ..._ ³


crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/MLSTest.kt
```diff
@@ -293,6 +299,48 @@ class MLSTest {
...
+
+        var didRun = false
+        val secondTransactionJob = launch {
+            coreCrypto.transaction {
+                didRun = true
+            }
+        }
+        secondTransactionJob.join()
+        assertTrue { didRun }
+    }
```
vitorhugods, line 342: Before the fix, `secondTransactionJob` wouldn't complete and the test would timeout.

coriolinus (approved):
Thanks for tracking this down and fixing it! You know the Kotlin much better than I do, so all I can say is that it looks reasonable to me.

PR #1142, opened 2025-05-16
chore: ci: run `cargo check` and clippy for all targets

istankovic:
Here targets refer to cargo targets: binaries, tests, benches etc.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -24,26 +24,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        tool:
-          - name: fmt
-            trailer: --all -- --check
-          - name: clippy --locked
-          - name: clippy --locked
-            trailer: --tests
-          - name: check --locked
-            trailer: --tests
-        target: ["", "--target wasm32-unknown-unknown"]
-        exclude:
-          - {tool: {name: fmt},
-             target: "--target wasm32-unknown-unknown"}
-
+        command:
+          - cargo fmt --all -- --check
+          - cargo check --locked --all-targets
+          - cargo check --locked --target wasm32-unknown-unknown
+          - cargo clippy --locked --all-targets
+          - cargo clippy --locked --target wasm32-unknown-unknown
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-and-cache-rust
         with:
           components: rustfmt, clippy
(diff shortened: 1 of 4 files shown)


.github/workflows/rust.yml
```diff
@@ -24,26 +24,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        tool:
-          - name: fmt
-            trailer: --all -- --check
-          - name: clippy --locked
-          - name: clippy --locked
-            trailer: --tests
```
coriolinus, line 32: Why are we no longer running clippy with the tests flag?

.github/workflows/rust.yml
```diff
@@ -24,26 +24,19 @@ jobs:
...
-          - name: check --locked
-            trailer: --tests
```
coriolinus, line 34: Likewise the check

coriolinus (commented):
I don't think that eliminating `black_box` calls from the benches is the right strategy. It feels like exactly the kind of hinting you want for a good benchmark. https://doc.rust-lang.org/beta/std/hint/fn.black_box.html

I'd much prefer annotating the black_box cases where clippy complains to say that we actually intended to pass a unit value there, rather than removing the black_box call.

.github/workflows/rust.yml
istankovic: Similar to above, `--all-targets` implies `--tests`.

Note that a `target` in this context does _not_ refer to what rustc commonly understands as `target`, i.e. the platform to compile for. In Cargo world, a target refers to various _build targets_, such a binary, benchmark, test...

istankovic:
> I don't think that eliminating `black_box` calls from the benches is the right strategy. It feels like exactly the kind of hinting you want for a good benchmark. https://doc.rust-lang.org/beta/std/hint/fn.black_box.html

For this particular case, I do not think that calls to `black_box` were done in any systematic or meaningful way.
(If someone knows more about the way it was done, I'd be happy to know!)

For instance, what should
```
black_box(());
```
actually do?

Not only that, but a thing like

```rust
black_box(foo(....)));
```
where `foo()` returns a unit value seems unnecessary, _especially_ when you consider the fact that arguments are evaluated eagerly. Honestly, without a proper reasoning why this actually help, this feels more like cargo culting.

coriolinus (approved):

PR #1150, opened 2025-05-20
Simon/test/use test conversation iteration 2 wpb 15788

SimonThormeyer:
# What's new in this PR

----

- The **PR Title**


crypto/src/test_utils/test_conversation.rs
```diff
@@ -55,6 +55,42 @@ impl<'a> TestConversation<'a> {
         member_count
     }
 
+    pub async fn are_members(&self, members_to_check: impl IntoIterator<Item = &'a SessionContext>) -> bool {
+        let member_ids = futures_util::future::join_all(self.members().map(|member| member.get_client_id())).await;
+        let ids_to_check =
+            futures_util::future::join_all(members_to_check.into_iter().map(|member| member.get_client_id())).await;
+
+        ids_to_check.into_iter().all(|id| member_ids.contains(&id))
```
coriolinus, line 63: It seems a little silly to take the iterator `members_to_check`, collect it into a vector `ids_to_check`, and then immediately turn that back into an iterator. It's a little painful to have to avoid nice convenience functions which only work synchronously, but even so, I'd still write it like this:

```rust
for member in members_to_check.into_iter() {
  let id = member.get_client_id().await;
  if !member_ids.contains(&id) {
    return false;
  }
}
true
```
```

SimonThormeyer:
Replaced by #1169

PR #1155, opened 2025-05-23
refactor!: remove 2-phase MLS init

istankovic:
It turns out clients don't actually use this API. And also there is no reason they should use it, since the backend doesn't require presence of a signature key before issuing a client ID.


typfel:
So this wasn't using a dedicated table in the database for storing the temporary credentials so need to clean anything up there?

istankovic:
> So this wasn't using a dedicated table in the database for storing the temporary credentials so need to clean anything up there?

As far as I can tell, no. It just used existing tables to store temporary IDs, but since nobody has been using the API, we don't need to clean that up either.

typfel (approved):

PR #1156, opened 2025-05-23
fix: allow registering epoch observer before calling mls_init

typfel:
# What's new in this PR

We are now storing the observer in the `CoreCrypto` struct instead of the `mls::Session`.

----

- The **PR Title**


crypto/src/test_utils/epoch_observer.rs
```diff
@@ -40,3 +41,9 @@ impl EpochObserver for TestEpochObserver {
         guard.observed_epochs.push((conversation_id, epoch));
     }
 }
+
+impl Debug for TestEpochObserver {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.write_str("TestEpochObserver")
+    }
+}
```
istankovic, line 49: We can probably use `derive_more` crate, which we're already using elsewhere, to derive such simple impls, like

```
#[derive(derive_more::Debug)]
#[debug("TestEpochObserver")]
struct TestEpochObserver(....)
```

crypto-ffi/src/core_crypto/epoch_observer.rs
```diff
@@ -140,24 +145,31 @@ impl EpochObserver {
...
     }
 }
 
+#[cfg(target_family = "wasm")]
+impl fmt::Debug for EpochObserver {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("EpochObserver")
+    }
+}
+
```
istankovic, line 166: Similar as in https://github.com/wireapp/core-crypto/pull/1156/files#r2104802935, we can use `derive_more` crate here to deriver simple impls.

crypto-ffi/src/core_crypto/epoch_observer.rs
```diff
@@ -55,12 +56,19 @@ pub trait EpochObserver: Send + Sync {
 #[cfg(not(target_family = "wasm"))]
 struct ObserverShim(Arc<dyn EpochObserver>);
 
+#[cfg(not(target_family = "wasm"))]
+impl fmt::Debug for ObserverShim {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("ObserverShim").field(&"Arc<dyn EpochObserver>").finish()
+    }
+}
```
istankovic, line 64: This should also be doable via `derive_more` crate, please see https://docs.rs/derive_more/latest/derive_more/derive.Debug.html.

crypto-ffi/src/core_crypto/epoch_observer.rs
```diff
@@ -53,14 +53,16 @@ pub trait EpochObserver: Send + Sync {
 /// actually reach its foreign target. However, the orphan rule prevents us from just tying the two traits
 /// together directly, so this is the straightforward way to accomplish that.
 #[cfg(not(target_family = "wasm"))]
+#[derive(Debug)]
```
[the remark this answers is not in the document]
typfel: Ok, fixed

crypto/src/mls/session/error.rs
```diff
@@ -31,8 +31,6 @@ pub enum Error {
     IdentityAlreadyPresent,
     #[error("The supplied credential does not match the id or signature schemes provided")]
     WrongCredential,
-    #[error("An EpochObserver has already been registered; reregistration is not possible")]
-    EpochObserverAlreadyExists,
```
coriolinus, line 35: Do we really want to relax this restriction? It's hard to envison non-buggy cases where the client registers an epoch observer more than once.

coriolinus (commented):
Looks like this PR has not yet removed `SessionInner::epoch_observer>`

crypto/src/mls/session/error.rs
typfel: Right, maybe you're right. I was aligning it with how the `mls_transport` callback.

I'll bring it back.

crypto/src/mls/session/epoch_observer.rs
```diff
@@ -2,14 +2,14 @@ use std::sync::Arc;
...
-use crate::{CoreCrypto, RecursiveError, mls::HasSessionAndCrypto as _, prelude::ConversationId};
+use crate::prelude::ConversationId;
 
 use super::{Error, Result, Session};
 
 /// An `EpochObserver` is notified whenever a conversation's epoch changes.
 #[cfg_attr(target_family = "wasm", async_trait(?Send))]
 #[cfg_attr(not(target_family = "wasm"), async_trait)]
-pub trait EpochObserver: Send + Sync {
+pub trait EpochObserver: std::fmt::Debug + Send + Sync {
```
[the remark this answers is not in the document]
typfel: No other reason than that's how it was done for the `MlsTransport` trait.

> Seems better to use the derive_more::Debug bound to stub the debug impl for this field in its container instead.

Ok so in practice this means remove the `Debug` trait here and instead use the `derive_more` macros on the epoch_observer field in the Session struct.

coriolinus (approved):
Sorry it took so much back and forth, but this looks great now!

PR #1161, opened 2025-05-27
feat!: crypto-ffi: bindings: remove getCredentialInUse

istankovic:
It's not used by clients.


coriolinus (approved):
Taking your word for it that it's unused, but assuming clients don't get bothered by it, then removing it is absolutely the right thing to do.

PR #1192, opened 2025-06-12
chore: remove proteus benchmarks

istankovic:
Proteus is going away, there's no need to have these benchmarks.


istankovic:
Dropped the second commit that removes the dependency on the old proteus branch, it's still too early for that because it's needed for the cryptobox migration test. Will move it to the PR that drops cryptobox migration.

PR #1200, opened 2025-06-16
chore: update CHANGELOG

istankovic:


coriolinus (approved):
I don't love that it'll get clobbered unless you take special care, but it's still better to have this than not.

PR #1227, opened 2025-07-01
feat: update Android NDK to 28.1 [WPB-18293]

istankovic:
NDK version 28 alows us to support Android environments with a 16k page
size. The NDK's tooling configuration ensures that built program
segments are aligned on a 16k boundary.


coriolinus (approved):
Haven't tested but agree it looks good.

PR #1204, opened 2025-06-17
chore: reduce duplication in Kotlin wrapper [WPB-16660]

coriolinus:
----
##### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [x] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #1244, opened 2025-07-08
refactor: drop `Ciphersuites` type

coriolinus:
# What's new in this PR

We'd added a rust-level `Ciphersuites` newtype to mirror what the Kotlin bindings were doing, but that ultimately seemed kind of pointless and tedious. So we're taking it out here. Later in #1204 we can remove it from the Kotlin wrapper.

----

- The **PR Title**


crypto-ffi/src/ciphersuite.rs
```diff
@@ -88,82 +86,22 @@ pub fn ciphersuite_default() -> Ciphersuite {
...
-impl Ciphersuites {
-    pub fn iter(&self) -> impl Iterator<Item = Ciphersuite> {
-        self.0.iter().copied()
-    }
-}
+pub(crate) type Ciphersuites = Vec<Ciphersuite>;
 
-#[cfg_attr(target_family = "wasm", wasm_bindgen)]
+/// Helper function to convert a list of integers into a list of ciphersuites
+#[cfg_attr(target_family = "wasm", wasm_bindgen(js_name = ciphersuitesFromU16s))]
```
istankovic, line 92: Do we need this function exposed? What is the intended usage? I would rather that clients are not aware of the internal representation of ciphersuite values and that they just use proper enums.

PR #1264, opened 2025-07-11
feat: add a way to update database key [WPB-18538]

istankovic:


istankovic:
> So we wrap `migrateDatabaseKeyTypeToBytes` but for `updateDatabaseKey` you rely on the re-exported function from `WireCoreCryptoUniffi`.

Right. I'm not sure what the reason was for wrapping the `migrateDatabaseKeyTypeToBytes`.

> But this achieves nothing so I guess we can stop wrapping `migrateDatabaseKeyTypeToBytes` as well?

We can, but unfortunately, that would be a breaking change since an argument's name would be changed (`path` in the wrapper, `name` in crypto-ffi). I think that's better left for later, as a separate PR.

crypto-ffi/bindings/js/test/wdio/database.test.ts
```diff
@@ -20,6 +20,63 @@ describe("database key", () => {
     });
 });
 
+describe("database key update", () => {
+    it("database key update works", async () => {
```
SimonThormeyer, line 24: NIt: AFAICT, the `describe()` string followed by the `it()` string should read like natural language, e.g., if you're grouping `database` tests inside a single `describe()`, 

```js
describe("database", () => {
    it(" key update works", async () => {
    // ....
    })
}) 

would read like _database key update works_.

Feel free to ignore though.
```

istankovic: Done.

istankovic: It would also make sense to move other tests in that file into the same group, but I will leave that for another PR.

PR #1269, opened 2025-07-14
chore: upgrade to use Dokka 2.0

typfel:
# What's new in this PR

We were already importing dokka 2.0 but we were still using the old v1 gradle tasks. This PR enables the v2 gradle tasks and updates the build config to use them.

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/Makefile.toml
+++ b/crypto-ffi/Makefile.toml
@@ -15,7 +15,7 @@ dependencies = [
 ]
 script = '''
   cd bindings
-  ./gradlew android:dokkaHtml
+  ./gradlew android:dokkaGeneratePublicationHtml
   cd -
   mkdir -p ../target/kotlin/doc
   cp -R bindings/android/build/dokka/html/ ../target/kotlin/doc
--- a/.github/workflows/bindings.yml
+++ b/.github/workflows/bindings.yml
@@ -160,7 +160,7 @@ jobs:
       - name: run dokka
         run: |
             cd crypto-ffi/bindings
-            ./gradlew android:dokkaHtml
+            ./gradlew android:dokkaGeneratePublicationHtml
             cd ..
             mkdir -p ../target/kotlin/doc
             cp -R bindings/android/build/dokka/html/ ../target/kotlin/doc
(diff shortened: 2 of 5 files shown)


crypto-ffi/Makefile.toml
```diff
@@ -15,7 +15,7 @@ dependencies = [
 ]
 script = '''
   cd bindings
-  ./gradlew android:dokkaHtml
+  ./gradlew android:dokkaGeneratePublicationHtml
   cd -
   mkdir -p ../target/kotlin/doc
```
SimonThormeyer, line 20: I guess we need the same diff in `bindings.yml`, too

typfel: Hmm shouldn't we be calling this script from bindings.yml?

crypto-ffi/bindings/jvm/build.gradle.kts
```diff
@@ -10,6 +10,18 @@ java {
...
+    pluginsConfiguration.html {
+        footerMessage.set("Copyright Wire GmbH")
+    }
+    dokkaSourceSets.main {
+        sourceLink {
+            remoteUrl("https://github.com/wireapp/core-crypto/tree/main/crypto-ffi/bindings/jvm")
+        }
+    }
+}
+
```
istankovic, line 24: Can we move this and the same block in `android` to toplevel `bindings/build.gradle.kts`?
They look identical, and that way we could define this in only one place.

typfel: So I explored this. First, no we can't just move the block to `bindings/build.gradle.kts` but what they suggest in the migration guide is to create convention plugin which applies the dokka configuration to all relevant modules.

https://kotlinlang.org/docs/dokka-migration.html#multi-module-projects-without-convention-plugins

Unfortunately when applying the dokka configuration in this way it breaks the setup and no documentation is generated at all :/

istankovic: Oh well, nothing is easy with gradle...

istankovic (approved):

crypto-ffi/bindings/jvm/build.gradle.kts
typfel: Oh wait, just tried some obvious which seems to work.

PR #1312, opened 2025-07-24
build: replace `Makefile.toml` and bun scripts with GNU `make` [WPB-17417] [WPB-17418]

coriolinus:
After some investigation, GNU `make` appears to be by far the most powerful tool we have to orchestrate our build procesess. And it's relatively ergonomic, for what it's doing! So this just does that migration.

----

- The **PR Title**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ WASM_PACK      := wasm-pack
 UNIFFI_BINDGEN := target/release/uniffi-bindgen
 
 # Default goal
-.DEFAULT_GOAL := all
+.DEFAULT_GOAL := local
 
 # Directory we store timestamps in
 STAMPS := .stamps
@@ -419,8 +419,9 @@ jvm-test: ## Run Kotlin tests on JVM (assuming you ran `make jvm` at some earlie
 # 9) Aggregate targets
 #-------------------------------------------------------------------------------
 
-.PHONY: wasm all
+.PHONY: wasm local all
 wasm: bindings-js  ## Alias for bindings-js
+local: bindings ts-fmt ## Generate and format all bindings
 all: android wasm jvm $(if $(filter Darwin,$(UNAME_S)),ios) docs ## Generate bindings for all platforms (android, iOS, wasm) and generate docs
 
 #-------------------------------------------------------------------------------


istankovic:
Running `make` on my machine results in a failure to build for `aarch64-apple-ios`, which is expected (not because Rust can't cross-compile, but because OpenSSL can't be cross-compiled on non-macOS systems). We should skip building Apple targets on non-Apple platforms.

Makefile
```diff
@@ -0,0 +1,484 @@
...
+#-------------------------------------------------------------------------------
+
+# Swift bindings
+bindings-swift.stamp: target/release/uniffi-bindgen target/release/libcore_crypto_ffi.$(LIBRARY_EXTENSION)
+	mkdir -p crypto-ffi/bindings/Swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi
+	$(UNIFFI_BINDGEN) generate \
+	  --language swift \
+	  --out-dir crypto-ffi/bindings/Swift/WireCoreCryptoUniffi/WireCoreCryptoUniffi \
+	  --library target/release/libcore_crypto_ffi.$(LIBRARY_EXTENSION)
+	touch $@
```
istankovic, line 102: We should prefix all `touch` commands with `@` to suppress command echoing, as it's not really interesting:
`@touch $@`.

Makefile
```diff
@@ -0,0 +1,484 @@
...
+#-------------------------------------------------------------------------------
+
+.PHONY: mobile wasm all
+mobile: android ios
+wasm: bindings-js
+all: mobile wasm docs
+
+# ────────────────────────────────────────────────────────────────────────
+# TypeScript / JS tasks (from crypto-ffi/bindings/js/package.json)
+# ────────────────────────────────────────────────────────────────────────
```
istankovic, line 414: We're not using the same character for making horizontal separators, please compare this to e.g. `9) Aggregate targets` above. I don't have a preference either way, but we should be consistent.

Makefile
```diff
@@ -0,0 +1,484 @@
...
+	  --build-tool-arguments "-project WireCoreCrypto.xcodeproj -scheme WireCoreCrypto" \
+	  -o ../../../target/swift/doc
+	touch $@
+
+.PHONY: docs-swift
+docs-swift: docs-swift.stamp
+
+# Group all docs
+.PHONY: docs
+docs: docs-rust-generic docs-rust-wasm docs-kotlin docs-ts docs-swift
```
istankovic, line 246: `docs-swift` should be skipped on non-Apple machines because the bindings and jazzy are not available.

Makefile
```diff
@@ -0,0 +1,484 @@
...
+#-------------------------------------------------------------------------------
+.PHONY: FORCE
+FORCE:
+
+#-------------------------------------------------------------------------------
+# 2) Build artifacts via Cargo (always run Cargo; Cargo itself detects up-to-date)
+#-------------------------------------------------------------------------------
+
+# one rule to build both the FFI library and the bindgen binary
+ target/release/uniffi-bindgen: FORCE
```
istankovic, line 73: Extra space at beginning of line.

Makefile
```diff
@@ -0,0 +1,484 @@
...
+# To enable optimizations, set the `WASM_RELEASE` variable to any non-empty value:
+#
+# i.e.
+#
+#   make wasm WASM_RELEASE=1
+ifeq ($(strip $(WASM_RELEASE)),)
+  WASM_BUILD_ARGS := --dev
+else
+  WASM_BUILD_ARGS :=
+endif
```
istankovic, line 153: This block can be writtent more compactly as
```Makefile
WASM_BUILD_ARGS := $(if $(WASM_RELEASE),,--dev)
```

Makefile
```diff
@@ -0,0 +1,484 @@
...
+		--locked \
+		--features uniffi/cli \
+		--package core-crypto-ffi \
+		--bin uniffi-bindgen
+
+target/release/libcore_crypto_ffi.$(LIBRARY_EXTENSION): FORCE
+	cargo build --release \
+		--locked \
+		--package core-crypto-ffi \
+		--lib
```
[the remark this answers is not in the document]
istankovic: I think Peter added this deliberately, and also the comment above suggests that this is the desired behaviour.
In general, I think it makes sense to delegate rebuild checks to cargo, as we don't want to get into that business. I asked the question here because I noticed some spurious rebuilds while I was testing, but it could be that there was something else going on. Let's leave this as-is for now, if I notice extra builds, I will make a note.

Makefile
```diff
@@ -0,0 +1,496 @@
...
+# one rule to build both the FFI library and the bindgen binary
+ target/release/uniffi-bindgen: FORCE
+	cargo build --release \
+		--locked \
+		--features uniffi/cli \
+		--package core-crypto-ffi \
+		--bin uniffi-bindgen
+
+target/release/libcore_crypto_ffi.$(LIBRARY_EXTENSION): FORCE
+	cargo build --release \
```
istankovic, line 87: I think it may make sense to generalize the `WASM_RELEASE` thing and just have a `RELEASE` variable, which would cause both wasm build args and cargo build args to be set correctly. Something like

```Makefile
ifeq ($(RELEASE),1)
  CARGO_BUILD_ARGS := --release
  WASM_BUILD_ARGS :=
else
  CARGO_BUILD_ARGS :=
  WASM_BUILD_ARGS := --dev
endif
```
That way we could also profit from faster local build times (e.g. I've been building for android locally and the difference is noticeable).

Makefile
```diff
@@ -0,0 +1,496 @@
...
+	  src/CoreCrypto.ts
+	@touch $@
+
+.PHONY: ts
+ts: $(DTS_OUT)
+
+# run WebDriver tests + bun’s built-in tests
+.PHONY: ts-test
+ts-test: ts
+	export CC_TEST_LOG_LEVEL=1 && \
```
istankovic, line 487: I would not set `CC_TEST_LOG_LEVEL=1` in the `Makefile`. That should be up to the user.

Makefile
```diff
@@ -0,0 +1,496 @@
...
+	cd crypto-ffi/bindings && \
+	./gradlew jvm:build -x lint -x lintRelease
+
+
+#-------------------------------------------------------------------------------
+# 9) Aggregate targets
+#-------------------------------------------------------------------------------
+
+.PHONY: mobile wasm all
+mobile: android ios
```
istankovic, line 418: We should also skip `ios` here if we're not on Darwin, just like in the other place.

Makefile
istankovic: > This leads to the question why do we want to build for release in CI if we're not publishing our package or runnings benches?

Because the CI is the last "gate" to `main` and the eventual release. In theory, one could postpone release builds to just before release, but then good luck to anyone who merged their PR some hundred commits ago.

Building in release mode on every PR ensures that potential breakages in release material are detected as soon as possible, before the PR gets merged.

istankovic:
`jvm-test` depends on `jvm`, which means that anytime we want to run jvm tests, `jvm` will be rebuilt. However, cargo's caching is not smart enough to not rebuild when nothing changed, so we always have to wait for the build when we do `make jvm-test`, which on my machine takes more than 1min.

Given that our old `Makefile.toml` didn't even have `-test` tasks, I think we should require the user to have manually invoked build rather than have test targets depend on build. (We should do this not just for `jvm-test` but for all `-test` targets.)

SimonThormeyer:
> `jvm-test` depends on `jvm`, which means that anytime we want to run jvm tests, `jvm` will be rebuilt. However, cargo's caching is not smart enough to not rebuild when nothing changed, so we always have to wait for the build when we do `make jvm-test`, which on my machine takes more than 1min.
> 
> Given that our old `Makefile.toml` didn't even have `-test` tasks, I think we should require the user to have manually invoked build rather than have test targets depend on build. (We should do this not just for `jvm-test` but for all `-test` targets.)

Does the same apply to `-docs` tasks?

Makefile
SimonThormeyer: While I don't necessarily disagree that we can do this, I would prefer if we could do this in another PR, as the PR says, we're here to migrate from `cargo make` to `GNU make`, not to optimize behavior (yet).

Makefile
```diff
@@ -0,0 +1,505 @@
...
+# run WebDriver benches
+.PHONY: ts-bench
+ts-bench: ts ## Run TypeScript wrapper benches in Chrome via wdio
+	cd $(JS_DIR) && \
+	bun x wdio run wdio.conf.ts --spec benches/**/*.bench.ts --log-level warn
+
+.PHONY: clean
+clean: ts-clean ## Run cargo clean and the ts-clean target, remove all stamps
+	cargo clean && \
+	rm -r .stamps
```
istankovic, line 505: Why not use `$(STAMPS)` here?

Makefile
```diff
@@ -0,0 +1,505 @@
...
+JS_DIR := crypto-ffi/bindings/js
+JS_SRC_DIR   := $(JS_DIR)/src
+PACKAGE_JSON := $(JS_DIR)/package.json
+
+GEN_DIR := $(JS_SRC_DIR)/autogenerated
+JS_GEN := $(GEN_DIR)/core-crypto-ffi.js
+WASM_GEN := $(GEN_DIR)/core-crypto-ffi_bg.wasm
+
+# find all .ts source files under src/ except in src/autogenerated
+TS_SRCS   := $(shell find $(JS_SRC_DIR) \( -path $(JS_SRC_DIR)/autogenerated -o -path $(JS_SRC_DIR)/corecrypto.d.ts -prune \) -o \( -type f -name '*.ts' \) -print)
```
istankovic, line 440: Can this be simpler? E.g.
```
TS_SRCS   := $(shell ls $(JS_SRC_DIR)/*.ts | grep -v corecrypto.d.ts)
```
?

crypto-ffi/bindings/js/package.json
```diff
@@ -50,11 +50,7 @@
     "access": "public"
   },
   "scripts": {
-    "clean": "rm -f src/corecrypto.js src/corecrypto.d.ts",
-    "build": "bun run clean && bun build --target browser --format esm --outfile src/corecrypto.js src/CoreCrypto.ts",
-    "postbuild": "bun x dts-bundle-generator --project tsconfig.json -o src/corecrypto.d.ts --no-check --export-referenced-types false src/CoreCrypto.ts",
-    "test": "bun x wdio run wdio.conf.ts --spec test/wdio/*.test.ts && bun test",
-    "bench": "bun x wdio run wdio.conf.ts --spec benches/**/*.bench.ts --log-level warn"
+    "clean": "rm -f src/corecrypto.js src/corecrypto.d.ts"
```
istankovic, line 53: I'm not sure what the intention was here, but it seems weird that we removed everything except `clean`.

.github/workflows/bindings.yml
```diff
@@ -282,10 +281,9 @@ jobs:
         run: |
           bun x @puppeteer/browsers install chromedriver@latest --path $PWD
           echo "CHROMEDRIVER_PATH=$(echo $PWD/chromedriver/*/*/chromedriver)" >> $GITHUB_ENV
-      - name: build wasm artifacts
+      - name: build ts artifacts
         run: |
-          cd crypto-ffi
-          cargo make wasm
+          make ts
```
istankovic, line 286: This is going to build WASM in dev mode, right? I think we should do a release mode build in CI.

.github/workflows/bindings.yml
```diff
@@ -332,16 +330,9 @@ jobs:
...
-      - name: build wasm
-        run: |
-          cd crypto-ffi
-          cargo make wasm
       - name: build ts
         run: |
-          cd crypto-ffi/bindings/js
-          bun install --frozen-lockfile
-          bun run build
+          make ts
```
istankovic, line 335: Similar comment as above, this is going to build wasm bindings in dev mode. We should generally set `WASM_RELEASE=1` in our CI jobs. Currently we're setting in only for publishing, but that means we're publishing something we didn't test in CI.

Makefile
```diff
@@ -0,0 +1,505 @@
...
+release-build:   target/release/libcore_crypto_ffi.$(LIBRARY_EXTENSION) ## Build core-crypto-ffi for the native arch (needed for uniffi)
+
+#-------------------------------------------------------------------------------
+# 3) Use stamp files for generators: only re-run when inputs change
+#-------------------------------------------------------------------------------
+
+# Swift bindings
+ifneq ($(UNAME_S),Darwin)
+$(STAMPS)/bindings-swift:
+	$(error Swift bindings generation is only supported on Darwin because OpenSSL can't be cross-compiled on non-Darwin systems; this is "$(UNAME_S)".)
```
istankovic, line 111: This should be a warning, not an error, because otherwise targets like `bindings` won't work at all on non-Apple platforms.

Makefile
```diff
@@ -0,0 +1,505 @@
...
+
+
+#-------------------------------------------------------------------------------
+# 9) Aggregate targets
+#-------------------------------------------------------------------------------
+
+.PHONY: mobile wasm all
+mobile: android $(if $(filter Darwin,$(UNAME_S)),ios) ## Generate bindings for android, and, if on macOS, also for iOS
+wasm: bindings-js  ## Alias for bindings-js
+all: mobile wasm docs ## Generate bindings for all platforms (mobile, wasm) and generate docs
```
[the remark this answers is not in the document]
coriolinus: Actually, do we even want `android` in `all`? For local development, I feel like the JVM target covers it, and it gives us a useful default for which we don't need to set up a relatively complicated dev environment.

Makefile
```diff
@@ -0,0 +1,504 @@
...
+
+.PHONY: bun-deps
+bun-deps: $(STAMPS)/bun-deps ## Install JS dependencies using bun
+
+# Full JS binding step
+bindings-js: wasm-build $(JS_OUT) ## Generate JavaScript bindings (append `WASM_RELEASE=1` to this command or any dependents for the release build)
+
+# All bindings
+.PHONY: bindings
+bindings: bindings-swift bindings-kotlin bindings-js ## Generate all bindings
```
coriolinus, line 200:
```suggestion
bindings: bindings-kotlin bindings-js $(if $(filter Darwin,$(UNAME_S)),bindings-swift) ## Generate all bindings
```

No need to generate a warning line about Swift every time we generate bindings locally.

Makefile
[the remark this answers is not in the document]
istankovic: I think `all` wouldn't be all without `android`. :-)
If we're building everything, then we should also include android and related artifacts.

Like you mentioned in another comment, perhaps there should be another default target.

There will be some pain points when this lands, for sure. For example, in my case, I was used to being able to just run `bun run build` in `crypto-ffi/bindings/js`, but that's no longer possible. It was nice being able to do such things, while being close to the source. With this, I would need to do things like `make -C` or resort to scripting.

Personally, I'm ok with iterating on the set of targets and adjusting the Makefile once we get some experience with the new setup.

coriolinus (commented):
Can't formally approve as I opened this PR, but I approve.

PR #1324, opened 2025-08-01
chore!: drop keystore v1

istankovic:
This PR drops support for keystore v1, which has been a full copy of our keystore entities on Wasm. This also drops support for migrating the db to v1. This would affect only the Wasm client that uses version of CoreCrypto < 2.0.


keystore/src/connection/platform/wasm/migrations/mod.rs
```diff
@@ -22,6 +21,7 @@ const fn db_version_number(counter: u32) -> u32 {
     VERSION_1_0_2 + counter
 }
 
+// Note that we no longer support migration to DB_VERSION_1.
```
SimonThormeyer, line 24: Shouldn't there be a superfluous constant below now?

PR #1325, opened 2025-08-01
Fix(android): support 16 kb

Garzas:
# What's new in this PR

This patch introduces two key changes to ensure that the Android native .so libraries produced by Core Crypto are compatible with devices and emulators using 16 KB memory pages (e.g. Pixel 9a, Pixel Fold, API 34+):

🔧 Changes introduced
🔼 Bumped JNA version from 5.14.0 to 5.17.0, which includes [support for 16 KB pages](https://github.com/java-native-access/jna/commit/0793e8be5b3342d54d02a0515fa94a62baadfa28) in libjnidispatch.so.

🛠 Added -Wl,-z,max-page-size=16384 linker flag for all Android Rust targets (armv7, armv8, x86_64) in Makefile.toml to ensure that libcore_crypto_ffi.so is built with proper segment alignment.

📉 Why this matters
Android 13+ on some devices (notably Pixel 9a) and emulator images now use 16 KB memory pages. If native libraries are built without this in mind, the loader will fail to map ELF segments, resulting in crashes like:

```
dlopen failed: segment 0 has unsupported page alignment
```

By explicitly setting max-page-size=16384, we ensure compatibility with:

- Pixel devices using 16K memory pages
- x86_64 emulator images (API 34+)
- Modern Android NDK requirements

✅ Verified
You can confirm proper alignment via:

```
readelf -lW libcore_crypto_ffi.so | grep LOAD
```
All LOAD entries should show:

```
Align 0x4000
```

(0x4000 = 16 KB)


crypto-ffi/bindings/gradle.properties
```diff
@@ -22,7 +22,7 @@ mavenCentralPublishing=true
 mavenCentralAutomaticPublishing=true
 signAllPublications=true
 GROUP=com.wire
-VERSION_NAME=8.0.1
+VERSION_NAME=8.0.2
```
SimonThormeyer, line 25: This change seems unrelated, would you mind factoring this out into its own commit?

crypto-ffi/Makefile.toml
```diff
@@ -217,7 +219,8 @@ args = [
     "--",
     "-C", "strip=symbols",
     "-l", "static=clang_rt.builtins-x86_64-android",
-    "-L", "${CLANG_RT_DIR}"
+    "-L", "${CLANG_RT_DIR}",
+    "-C", "link-arg=-Wl,-z,max-page-size=16384"
```
istankovic, line 223: Why is this needed? Isn't this supposed to be handled automatically by the correct tooling?

PR #1343, opened 2025-08-08
chore: update to Rust 1.89

istankovic:


rust-toolchain.toml
```diff
@@ -1,3 +1,11 @@
...
-components = [ "rust-analyzer" ]
+channel = "1.89"
+components = ["rust-analyzer"]
+targets = [
+  "aarch64-apple-ios",
+  "aarch64-apple-ios-sim",
+  "aarch64-linux-android",
+  "armv7-linux-androideabi",
+  "x86_64-linux-android",
+  "wasm32-unknown-unknown",
```
fewerner, line 10: Why do we have to set these targets now?

fewerner (approved):

rust-toolchain.toml
istankovic: This is so we don't need to manually tell rustup to get them every time we update the toolchain. (See message of the commit that adds these.)

If we don't have these, then any attempt to cross-compile for a target you don't have installed locally will fail.
You will get a nice error message saying you need to install the necessary target, and then you need to do e.g.
````
rustup target add armv7-linux-androideabi
```
and similar for all the targets you will use in the future.

So this is just for developer convenience, to not have to do that. Because we know the list of cross-compilation targets in advance, we can just specify them here and rustup will make sure to download them and install them, making future `cargo build`s just work, regardless of the target.

PR #1382, opened 2025-08-19
build: add `make android` heuristics [WPB-19444]

coriolinus:
# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ TOUCH_STAMP = @mkdir -p $(STAMPS) && touch $@
 #-------------------------------------------------------------------------------
 
 # Relevant crates for FFI builds
-CRATES := crypto keystore crypto-macros mls-provider
+CRATES := crypto crypto-ffi keystore crypto-macros mls-provider
 
 # Workspace-level Cargo config
 WORKSPACE_CARGO_FILES := Cargo.toml Cargo.lock


.github/workflows/build-android.yml
```diff
@@ -45,10 +45,39 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: ./.github/actions/setup-and-cache-rust
+      - run: make uniffi-bindgen
+      - uses: actions/upload-artifact@v4
+        with:
+          name: uniffi-bindgen
+          path: target/release/uniffi-bindgen
+          retention-days: 1
```
SimonThormeyer, line 59: To share this between platform-specific workflows, will we need to unify them?

Makefile
```diff
@@ -368,13 +370,10 @@ $(STAMPS)/android-x86: android-env
 	$(TOUCH_STAMP)
```
istankovic: Could we generate rules for all Android targets?

It seems redundant to have to write explicitly the target in cargo invocations and the target name (e.g. `target/x86_64-linux-android/...).

For example (heavily handwaving):
```Makefile
TARGETS := ...
$(foreach target,$(TARGETS),$(eval \
target/$(target)/...: android-env
   $(CARGO) rustc --target $(target) ...
))
```

SimonThormeyer (approved):
Great job!

Makefile
```diff
@@ -90,10 +90,23 @@ STAMPS := .stamps
...
 
 #-------------------------------------------------------------------------------
-# Dummy force target: always out-of-date, so Cargo build commands run every time
+# Rust file-based heuristics
 #-------------------------------------------------------------------------------
-.PHONY: FORCE
-FORCE:
+
+# Relevant crates for FFI builds
+CRATES := crypto keystore crypto-macros mls-provider
```
istankovic, line 97: What about `crypto-ffi`?

.github/workflows/build-android.yml
coriolinus: See also https://wearezeta.atlassian.net/browse/WPB-19624

Makefile
coriolinus: It may be technically possible, but I'd rather not, for 2 reasons:

1. We'd only be generalizing over 2 of the android targets; `android-x86` is straightforwardly different. IMO YAGNI applies here.
2. I'd rather stay away from `foreach/eval` syntax in this makefile if possible; its intrinsic complexity is high enough already without going there.

PR #1394, opened 2025-08-21
Simon/ci/reuse uniffi bindgen artifact

SimonThormeyer:
# What's new in this PR

----

- The **PR Title**


.github/actions/make/uniffi-bindgen/action.yml
```diff
@@ -0,0 +1,16 @@
+name: Fetch or make uniffi-bindgen binary
+
+runs:
+  using: composite
+  steps:
+    - name: uniffi bindgen artifact name
+      run: echo "ARTIFACT_NAME=uniffi-$(uname -s)-$(cargo info uniffi | grep version | head -1 | awk '{print $2}')" >> $GITHUB_ENV
```
coriolinus, line 7: I'd break up those sub invocations. Also, `cargo metadata` may be more stable over time than `cargo info`; it's intended for scripting purposes.

```sh
os="$(uname -s)"
uniffi_version="$(cargo metadata --format-version 1 | jq -r '.packages[] | select(.name == "uniffi") | .version'
echo "ARTIFACT_NAME=uniffi-${os}-${uniffi-version}" >> $GITHUB_ENV
```

.github/actions/make/action.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Attempt to download the artifact
+      uses: actions/cache/restore@v4
+      with:
+        key: ${{ inputs.key }}
+        path: "${{ inputs.target-path }}"
+      id: artifact-download
+
+    - name: Run make ${{ inputs.make-rule }} if needed
+      if: steps.artifact-download.outputs.cache-hit != 'true'
+      run: make ${{ inputs.make-rule }}
```
coriolinus, line 29: We assume release but are not making with `RELEASE=1`. How did this ever work?

But probably we should make in release mode anyway.

.github/workflows/build-android.yml
```diff
@@ -45,18 +45,13 @@ jobs:
           path: target/${{ matrix.target }}/release/libcore_crypto_ffi.*
           retention-days: 1
 
-  uniffi-bindgen:
```
coriolinus, line 48: What caused this rename?

coriolinus (commented):
I think we could use a precompiled `uniffi-bindgen` in many more places in the bindings in particular.

SimonThormeyer:
> I think we could use a precompiled `uniffi-bindgen` in many more places in the bindings in particular.

True! I only used it for android and ios, forgot to use it for JVM bindings.
I'm going to take a another look at the make file to find all necessary dependents.

.github/actions/make/uniffi-bindgen/action.yml
```diff
@@ -0,0 +1,19 @@
...
+        echo "ARTIFACT_NAME=uniffi-${os}-${uniffi_version}" >> $GITHUB_ENV
+      shell: bash
+    - uses: ./.github/actions/make
+      with:
+        key: ${{ env.ARTIFACT_NAME }}
+        make-rule: uniffi-bindgen
+        target-path: target/release/uniffi-bindgen
+
+    - run: chmod +x target/release/uniffi-bindgen
+      shell: bash
```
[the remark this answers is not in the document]
SimonThormeyer: This is required by github in composite actions

.github/actions/make/jvm-linux/action.yml
```diff
@@ -0,0 +1,11 @@
+name: Fetch or make the ffi lib for jvm linux
+
+runs:
+  using: composite
+
+  steps:
+      - uses: ./.github/actions/make
+        with:
+          key: jvm-linux-${{ github.sha }}
```
coriolinus, line 9: So, there are a ton of places where the key depends on `github.sha`. But the reusability there is pretty low.

I wonder how hard it would be to adjust our generic make action to compute a hash based on the inputs of the `make-rule`, and use that for the cache. Possibly that would be better for a separate PR after this. But it's easy to imagine a golden future in which, when you adjust a README or something, all the CI passes in seconds because it just reads the inputs from cached outputs. Or adjusting the wasm-tests reads the cache for ios and android. That kind of thing.

.github/actions/make/action.yml
```diff
@@ -0,0 +1,50 @@
...
+inputs:
+  key:
+    description: The artifact key
+    required: true
+  make-rule:
+    description: The GNU make rule which produces the artifact
+    required: true
+  target-path:
+    description: The relative path of the final artifact(s)
+    required: false
```
istankovic, line 12: I think `target-path` should be required to make it explicit for every job that uses this action what it actually stores.

coriolinus (approved):

.github/actions/make/jvm-linux/action.yml
```diff
@@ -0,0 +1,11 @@
...
+          make-rule: jvm-linux
+          target-path: target/x86_64-unknown-linux-gnu/release/*.so
```
istankovic, line 11: I would prefer to explicitly name the library, i.e. `libcore_crypto_ffi.so`.

.github/actions/make/android/lib/action.yml
```diff
@@ -0,0 +1,37 @@
...
+            ;;
+        esac
+        echo "ANDROID_TARGET=$target" >> $GITHUB_ENV
+
+    - uses: ./.github/actions/make
+      with:
+        key: lib-${{ inputs.task }}-${{ github.sha }}
+        make-rule: ${{ inputs.task }}
+        target-path: |
+          target/${{ env.ANDROID_TARGET }}/release/libcore_crypto_ffi.a
```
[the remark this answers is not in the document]
istankovic: Sorry for not being precise enough, I meant the `libcore_crypto_ffi.a` archive.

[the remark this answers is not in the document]
SimonThormeyer: Right. Removing that here would require
- updates to the `Makefile`
- updates to at least one `gradle` build file

which seems out of scope of this PR. I'll create a Jira item for removing these unneeded artifacts though!

PR #1405, opened 2025-08-25
build!: crypto-ffi: bump minimum Android SDK version to 35

istankovic:
The Android client already uses it.


coriolinus (dismissed):

Garzas (changes requested):
Current min sdk for android is 26 so we cannot change it, because all android devices with lower API will not support it

PR #1417, opened 2025-08-27
ci: use GitHub Actions artifacts instead of caches for build artifacts [WPB-17419]

SimonThormeyer:
# What's new in this PR
See title

----
##### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [x] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #1419, opened 2025-08-29
build: Makefile: don't use stamps for bindings-* rules

istankovic:
We know which files uniffi needs to generate so we should just refer to them.


fewerner (approved):
Approving, CI issue probably related to [PR1417](https://github.com/wireapp/core-crypto/pull/1417)

PR #1420, opened 2025-08-29
build: allow debug android builds

istankovic:
This makes gradle debug builds work properly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -298,14 +298,8 @@ $(ANDROID_X86): $(RUST_SOURCES) | android-env
 .PHONY: android-x86
 android-x86: $(ANDROID_X86) ## Build core-crypto-ffi for x86_64-linux-android
 
-.PHONY: ensure-release-mode
-ensure-release-mode:
-ifneq ($(RELEASE_MODE),release)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -19,10 +19,12 @@ SHELL := /usr/bin/env bash
 ifeq ($(RELEASE),)
   CARGO_BUILD_ARGS :=
   XCODE_CONFIG := Debug
+  GRADLE_BUILD_TYPE := Debug
```
coriolinus, line 22: Looks like `XCODE_CONFIG` and `GRADLE_BUILD_TYPE` are always equivalent. Could it be reasonable to unify them into some kind of `RELEASE_MODE_CAPITALIZED`?

coriolinus (commented):
Once CI goes green, this all looks reasonable.

Makefile
istankovic: Eh, I had weighed that before I pushed the PR, but eventually decided against it because I felt it was better to leave different build systems' options isolated. I don't have a strong opinion on it though.

SimonThormeyer (approved):
Approving this according to my "as long as it works its fine" gradle review policy

PR #1426, opened 2025-09-02
feat(decode): add options for listing members or identities

typfel:
# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/decode/Cargo.toml
+++ b/decode/Cargo.toml
@@ -10,6 +10,6 @@ workspace = true
 [dependencies]
 clap = { version = "4.5.47", features = ["derive"] }
 clap-stdin = "0.7.0"
+mls-rs = "0.50.1"
 proteus-wasm = { workspace = true, features = ["serde"] }
 base64 = { workspace = true }
-openmls = { workspace = true }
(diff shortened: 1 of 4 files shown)


decode/Cargo.toml
```diff
@@ -10,6 +10,7 @@ workspace = true
 [dependencies]
 clap = { version = "4.5.46", features = ["derive"] }
 clap-stdin = "0.6.0"
+mls-rs = "0.50.1"
 proteus-wasm = { workspace = true, features = ["serde"] }
 base64 = { workspace = true }
-openmls = { workspace = true }
+#openmls = { workspace = true }
```
istankovic, line 16: If we're not going to use openmls in decode, we should just remove this line, whoever needs it later can have a look at git history.

decode/src/main.rs
```diff
@@ -144,12 +160,83 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     }
 }
+
+trait GroupInfoExt {
+    fn print_members(&self);
+    fn print_basic_identities(&self);
+}
+
+impl GroupInfoExt for GroupInfo {
+    fn print_members(&self) {
```
istankovic, line 213: I'm not sure why we need a trait here, why not just have regular functions that take `&GroupInfo`, e.g.
```rust
fn print_members(group_info: &GroupInfo) { .. }
```
?

istankovic:
It would be nice to mention the new options in our changelog, to show that we're also able to add features, not just breaking changes. ;-)

PR #1428, opened 2025-09-02
ci: remove targets list from rust-toolchain.toml before rust setup [WPB-19717]

fewerner:
The setup-rust-toolchain action will install all targets specified in the rust-toolchain.toml even if we specify a target as an argument. This wastes CI runtime. Therefore we remove the targets list before running the action.

----

- The **PR Title**

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -19,11 +19,16 @@ jobs:
       matrix:
         include:
           - task: android-armv7
+            target: armv7-linux-androideabi
           - task: android-armv8
+            target: aarch64-linux-android
           - task: android-x86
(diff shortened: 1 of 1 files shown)


.github/actions/setup-and-cache-rust/action.yml
```diff
@@ -24,6 +24,14 @@ runs:
       run: |
         echo HOMEBREW_NO_AUTO_UPDATE=1 >> $GITHUB_ENV
         echo HOMEBREW_NO_INSTALL_UPGRADE=1 >> $GITHUB_ENV
+    - name: remove toolchain target list
```
SimonThormeyer, line 27: Maybe add a comment here why we're doing that

PR #1436, opened 2025-09-04
ci: orchestrate all workflows in a single `pipeline.yml`

SimonThormeyer:
# What's new in this PR

----

- The **PR Title**


coriolinus (approved):
I really like breaking things out into separate files like this; at the cost of a bit more tracing between files, we make it very easy to go directly to i.e. "the ci for interop", which is great.

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,233 @@
...
+      - ffi-library-linux
+      - build-android
+    uses: ./.github/workflows/package-android.yml
+
+  test-android:
+    needs: package-android
+    uses: ./.github/workflows/test-android.yml
+
+  #################
+  # JVM Linux
```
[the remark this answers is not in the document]
istankovic: AFAIU, we have been shipping both Linux and Darwin libraries in the `-uniffi-` package to allow also people on Macs to use CC on JVM. If we're no longer packaging the Darwin library, that would be a regression.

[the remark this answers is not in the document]
SimonThormeyer: Ugh, you're right. Now I know why this felt like a deja-vu. It's the second you're saying this and me finding in `publish-jvm.yml` that we're building jvm darwin only in the release workflow today.

.github/workflows/build-android.yml
```diff
@@ -13,7 +13,7 @@ env:
 
 jobs:
   build-android:
-    if: github.repository == 'wireapp/core-crypto'
+    if: github.event_name == 'pull_request'
```
istankovic, line 16: Why do we need this `if`?

.github/workflows/interop.yml
```diff
@@ -0,0 +1,74 @@
...
+
+on:
+  workflow_call:
+
+env:
+  RELEASE: 1
+
+jobs:
+  e2e-interop-test:
+    if: github.event_name == 'pull_request'
```
istankovic, line 13: Why do we need this `if`?

.github/workflows/build-android.yml
SimonThormeyer: we're not rebuilding android on pushes to main, because we've already built and tested it on the PR branch.

.github/workflows/test-ts.yml
```diff
@@ -0,0 +1,38 @@
...
+
+on:
+  workflow_call:
+
+env:
+  RELEASE: 1
+
+jobs:
+  test-ts:
+    if: github.event_name == 'pull_request'
```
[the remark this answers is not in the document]
SimonThormeyer: See above

.github/workflows/build-android.yml
istankovic: Aye, I was thinking we should probably reconsider, or at least audit, the whole approach but I guess that would be for another PR.

PR #1451, opened 2025-09-09
feat: salt obfuscated values

coriolinus:
Obfuscated values should change "regularly" for some definition of that term. Our goal is to compromise between debugging utility, for which pure determinism is nice, and customer privacy, for which we would simply never reveal any part of any sensitive value.

This solution means that the same value obfuscated on two devices will appear differently in the logs. It also means that the same value obfuscated between restarts of the client will appear differently.

# What's new in this PR

----

- The **PR Title**


fewerner (approved):
Looks good!

PR #1476, opened 2025-09-17
fix: use correct function name for key type migration

istankovic:
v9.0.0 had erroneously renamed `migrateDatabaseKeyTypeToBytes` to `migrateDbKeyTypeToBytes`.
This fixes it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+- v9.0.0 had erroneously renamed `migrateDatabaseKeyTypeToBytes` to `migrateDbKeyTypeToBytes`.
+  This has been fixed, and `migrateDatabaseKeyTypeToBytes` is usable again on all platforms.
+
+  Affected platforms: Android
+
+
 ## v9.0.0 - 2025-09-16
 
 ### Highlights


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+- v9.0.0 had erroneously renamed `migrateDatabaseKeyTypeToBytes` to `migrateDbKeyTypeToBytes`.
+  This has been fixed, and `migrateDatabaseKeyTypeToBytes` is usable again on all platforms.
+
+  Affected platforms: Android, iOS
```
SimonThormeyer, line 10: Is iOS really affected by this?
note that the function has been called `migrateDatabaseKeyTypeToBytes()` before https://github.com/wireapp/core-crypto/blob/89ddac21c72550a18a88443c6ff378b7a33185d9/crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCrypto.swift#L262-L267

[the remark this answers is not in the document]
istankovic: You're right, fixed, thanks.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
 ## Unreleased
 
+### Breaking Changes
+
+- v9.0.0 had erroneously renamed `migrateDatabaseKeyTypeToBytes` to `migrateDbKeyTypeToBytes`.
+  This has been fixed, and `migrateDatabaseKeyTypeToBytes` is usable again on all platforms.
+
+  Affected platforms: Android
+
+
```
SimonThormeyer, line 12: nit: below, we always had 1 blank line before a new heading, here it's 2

istankovic: Ah, sorry, saw this too late (now). I added the extra line semi-intentionally because it felt easier to read, distancing it a bit from the previous release. But I agree, we should be consistent. What do you think about separating releases in markdown with an additional newline like that?

PR #1480, opened 2025-09-18
ci: fix: `%-hash-deps` `make` rule fails appropiately

SimonThormeyer:
Previously, the rule wouldn't fail if a prerequisite file doesn't exist or the queried make variable hasn't been created.

# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/.github/workflows/publish-android.yml
+++ b/.github/workflows/publish-android.yml
@@ -44,6 +44,12 @@ jobs:
       - name: zip package and bindings
         run: |
           zip -r android.zip \
+          target/armv7-linux-androideabi/release/libcore_crypto_ffi.a \
+          target/armv7-linux-androideabi/release/libcore_crypto_ffi.so \
+          target/aarch64-linux-android/release/libcore_crypto_ffi.a \
+          target/aarch64-linux-android/release/libcore_crypto_ffi.so \
+          target/x86_64-linux-android/release/libcore_crypto_ffi.a \
+          target/x86_64-linux-android/release/libcore_crypto_ffi.so \
           crypto-ffi/bindings/android/build/outputs/aar/android-release.aar \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -108,8 +108,11 @@ RUST_SOURCES := $(WORKSPACE_CARGO_FILES) $(CRATE_MANIFESTS) $(RUST_RS_FILES)
 
 # Used by CI to calculate a hash of prerequisite files of a make rule
 %-hash-deps:
-	@deps="$($*-deps)"; \
-	hash=$$(sha256sum $$deps | sha256sum | awk '{print $$1}'); \
+	@if [ -z "$($*-deps)" ]; then \
+	  echo "ERROR: The \"$*-deps\" variable doesn't exist in the Makefile."; exit 1; \
+	fi
+	@@set -euo pipefail; \
+	hash=$$(ls $($*-deps) | xargs sha256sum | sha256sum | awk '{print $$1}'); \
```
coriolinus, line 115: I don't understand why we need `ls` here. Couldn't it be something like

```Makefile
hash=$$(sha256sum $($*-deps) | sha256sum | awk '{print $$1}')
```

?

SimonThormeyer: Yes, it can, thanks.

Makefile
```diff
@@ -260,6 +260,7 @@ ios-simulator-arm: $(IOS_SIMULATOR_ARM) ## Build core-crypto-ffi for aarch64-app
 $(STAMPS)/ios: ios-device ios-simulator-arm
 	$(TOUCH_STAMP)
 
+ios-deps := $(RUST_SOURCES)
```
coriolinus, line 263: Shouldn't it be:

```suggestion
ios-deps := $(ios-device-deps) $(ios-simulator-arm-deps)
```

Granted it comes out to the same thing, but I think this version will be more maintainable in the event something changes.

PR #1492, opened 2025-09-22
refactor: align swift and kotlin api [WPB-17438]

fewerner:
# What's new in this PR
This PR Aligns Swift and Kotlin APIs and aligns some ts parameters.
We rename our internal CoreCrypto class  to CoreCryptoFfi to avoid namespace conflicts.

----

- The **PR Title**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,29 @@
 
 ## Unreleased
 
+- Renamed `CoreCryptoClient` to `CoreCrypto` and moved `historyClient(historySecret: HistorySecret)` into `CoreCrypto` Companion functions
+  Affected platforms: jvm, android
+
+- Removed static methods from `CoreCrypto` that are globally available:
+
+  - removed `version()`
+  - removed `buildMetadata()`
+
+  Affected platforms: ios
+
+- Removed static methods from `CoreCrypto` that are globally available:
+
+  - removed `setMaxLogLevel(level: CoreCryptoLogLevel)` on ios, web
+  - removed `setLogger(logger: CoreCryptoLogger)` on ios, web
+
+  Affected platforms: ios, web
+
+- Removed `setLogger(logger: CoreCryptoLogger, level: CoreCryptoLogLevel)` and
+  renamed `setLoggerOnly(logger: CoreCryptoLogger)` to `setLogger(logger: CoreCryptoLogger)`.
+  To set the loglevel use `setMaxLogLevel(level: CoreCryptoLogLevel)`
+
+  Affected platforms: all
+
 ## v9.0.1 - 2025-09-18
 
 ### Breaking Changes


crypto-ffi/src/core_crypto/logger.rs
```diff
@@ -275,7 +265,7 @@ fn set_logger_only_inner(logger: Logger) {
 /// Initializes the logger
 #[cfg(not(target_family = "wasm"))]
 #[uniffi::export]
-pub fn set_logger_only(logger: Logger) {
+pub fn set_logger(logger: Logger) {
     set_logger_only_inner(logger);
```
coriolinus, line 269: Is `set_logger_only_inner` used anywhere else, or could we reasonably inline it here, now that the other `set_logger` call is gone?

crypto-ffi/bindings/jvm/src/main/kotlin/com/wire/crypto/CoreCrypto.kt
```diff
@@ -16,22 +16,22 @@ suspend operator fun CoreCrypto.Companion.invoke(
...
 class CoreCryptoClient(private val cc: CoreCrypto) {
-    companion object
+    companion object {
+        /**
+         * Instantiate a history client.
+         *
+         * This client exposes the full interface of `CoreCrypto`, but it should only be used to decrypt messages.
+         * Other use is a logic error.
+         */
+        suspend fun historyClient(historySecret: HistorySecret) = coreCryptoHistoryClient(historySecret).lift()
```
coriolinus, line 33: So to construct a history client, you need to call `CoreCryptoClient.historyClient`. This is the only public constructor defined on `CoreCryptoClient`. 

I'd prefer that it were different, but I suspect that if the FFI class gets renamed then my objection may become obsolete.

CHANGELOG.md
```diff
@@ -2,6 +2,29 @@
...
+
+  Affected platforms: ios
+
+- Moved `historyClient(historySecret: HistorySecret)` into CoreCryptoClient Companion functions
+
+  Affected platforms: jvm, android
+
+- Renamed CoreCrypto wrapper to CoreCryptoClient and CoreCryptoProtocol to CoreCryptoClientProtocol
+
+  Affected platforms: ios
```
coriolinus, line 24: I think at this point in the history you have not actually done this. Also we wanted not to do this.

crypto-ffi/bindings/jvm/src/main/kotlin/com/wire/crypto/CoreCrypto.kt
```diff
@@ -8,7 +8,7 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 
 /** Wrap a `CoreCrypto` instance in a `CoreCryptoClient` instance. Should largely be invisible to end-users. */
-fun CoreCrypto.lift() = CoreCryptoClient(this)
+fun CoreCryptoFfi.lift() = CoreCrypto(this)
 
 /** Opens an existing core crypto client or creates a new one if one doesn't exist at the `keystore` path */
 suspend operator fun CoreCrypto.Companion.invoke(
```
coriolinus, line 14: Now that we're renaming the uniffi class to `CoreCryptoFfi`, we don't need this external declaration of an invoke method on the companion object anymore. We can just declare a secondary constructor on `CoreCrypto` which takes these methods and produces a `CoreCrypto`.

coriolinus (commented):
Overall things look pretty good! But I think this is not quite fully baked yet.

crypto-ffi/bindings/js/src/CoreCryptoInstance.ts
```diff
@@ -149,7 +151,7 @@ function historySecretIntoFfi(secret: HistorySecret): HistorySecretFfi {
  * @param logger - the interface to be called when something is going to be logged
  **/
 export function setLogger(logger: CoreCryptoLogger): void {
-    CoreCryptoFfi.set_logger(loggerIntoFfi(logger));
+    set_logger(loggerIntoFfi(logger));
```
coriolinus, line 154: This doesn't recurse?

crypto-ffi/src/core_crypto/logger.rs
```diff
@@ -276,16 +276,16 @@ pub fn set_max_log_level(level: CoreCryptoLogLevel) {
...
 #[cfg(target_family = "wasm")]
 #[wasm_bindgen]
-impl CoreCryptoFfi {
-    /// Set the CC logger
-    pub fn set_logger(logger: CoreCryptoLogger) {
-        set_logger_only_inner(logger);
-    }
+pub fn set_logger(logger: CoreCryptoLogger) {
+    set_logger_inner(logger);
+}
```
coriolinus: So at this point, we have the wasm-only `fn set_logger`, and the non-wasm `fn set_logger`, which differ only in their annotations. So really it should be all one declaration now:

```rust
/// Set the CC logger
#[cfg_attr(target_family = "wasm", wasm_bindgen)]
#[cfg_attr(not(target_family = "wasm"), uniffi::export]
pub fn set_logger(logger: CoreCryptoLogger) {
  // contents of set_logger_inner go here
}
```

fewerner: I agree, but the logger type differs between uniffi and wasm.  `setLogger`uses Arc<CoreCryptoLogger>. For `set_max_log_level` we can do it like that.

PR #1505, opened 2025-09-29
ci: explicitly set make or gmake [WPB-20142]

fewerner:
login shell was needed so that the runners load bash configs. Instead of relying on hidden configs on our local runners we rather set the make executable in our workflow files.

# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/.github/actions/make/action.yml
+++ b/.github/actions/make/action.yml
@@ -17,11 +17,14 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Set MAKE
+      uses: ./.github/actions/make/set-make-executable
+
     - name: compute artifact key
       shell: bash -euo pipefail {0}
       id: compute-key
       run: |
-        files_hash=$(make "${{ inputs.make-rule }}"-hash-deps)
+        files_hash=$($MAKE "${{ inputs.make-rule }}"-hash-deps)
         echo "key=${{ inputs.key }}-$files_hash" >> "$GITHUB_OUTPUT"
 
     - name: branch name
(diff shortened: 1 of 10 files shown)


.github/actions/make/action.yml
```diff
@@ -89,18 +99,18 @@ runs:
             touch "$dir"
           fi
         done
-      shell: bash -leo pipefail {0}
+      shell: bash -euo pipefail {0}
```
SimonThormeyer, line 102: can we simplify this to just `bash`?

fewerner: default bash in gha is `bash --noprofile --norc -eo pipefail ` which misses -u.

coriolinus: Do we want `--noprofile --norc`, if we're not doing a login shell anyway?

fewerner: we did not use it before, I don't think it is relevant for us

.github/workflows/docs-swift.yml
```diff
@@ -18,6 +18,11 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
+      # This is only needed because of our composite actions calling gmake on ios
+      # to support our local runners without bash configs.
+      - name: Install gmake
+        run: brew install make
```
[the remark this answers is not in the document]
fewerner: Make comes preinstalled on all images and our local runners.  I don't think we will ever need an action for that. To install it here is just an easy workaround to give our local runners and the github action the same executable.

PR #1516, opened 2025-10-01
refactor: rm core-crypto prelude

coriolinus:
# What's new in this PR

Glob imports from a prelude are an antipattern which should be avoided
unless absolutely necessary. Simpler and clearer to reexport everything
which needs reexporting at the root level, then just import items directly.

This therefore drops those.

----

- The **PR Title**

### DIFF
--- a/keystore-dump/src/main.rs
+++ b/keystore-dump/src/main.rs
@@ -109,11 +109,11 @@ async fn main() -> anyhow::Result<()> {
         .collect::<postcard::Result<_>>()?;
     json_map.serialize_entry("mls_keypackages", &keypackages)?;
 
-    let e2ei_enrollments: Vec<core_crypto::prelude::E2eiEnrollment> = keystore
+    let e2ei_enrollments: Vec<core_crypto::E2eiEnrollment> = keystore
         .find_all::<E2eiEnrollment>(Default::default())
         .await?
         .into_iter()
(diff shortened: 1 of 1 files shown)


crypto/src/mls/conversation/commit.rs
```diff
@@ -52,17 +52,18 @@ mod tests {
     use itertools::Itertools;
     use openmls::prelude::SignaturePublicKey;
 
-    use crate::mls::conversation::Conversation as _;
-    use crate::mls::conversation::ConversationWithMls as _;
-    use crate::test_utils::*;
-    use crate::transaction_context::Error as TransactionError;
-
     use super::{Error, *};
```
fewerner, line 55: not part of your changes, but this looks redundant to me since `Error` is `pub` and will be imported by `*`. 
I genererally wonder if `use super::* `is something that we should avoid. For two reasons, 1. importing more than needed, 2. relative imports.

coriolinus: Agree, this particular line is pointless. (Though FWIW `use super::*` catches all items, not just `pub` items.)

In general `use super::*;` is practical and convenient for unit tests; this is a super common and useful pattern:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    // unit tests here
}
```

So we should not just unconditionally get rid of all of those instances. On the other hand, there are not many cases other than this where a glob import is idiomatic. 

Either way, I'd prefer to make that a separate PR.

PR #1519, opened 2025-10-02
style: run nightly fmt for all modules

fewerner:
# What's new in this PR

----

- The **PR Title**


.cargo/config.toml
```diff
@@ -24,24 +24,17 @@ CFLAGS = "-Wno-error=inline"
 [target.armv7-linux-androideabi]
 ar = "llvm-ar"
 linker = "armv7a-linux-androideabi24-clang"
-rustflags = [
-    "-C", "target_feature=+neon",
-    "-C", "link-args=-latomic"
-]
+rustflags = ["-C", "target_feature=+neon", "-C", "link-args=-latomic"]
 
 [target.aarch64-linux-android]
```
SimonThormeyer, line 29: Does the nightly formatter also format `.toml` files? Would that potentially be conflicting with the other toml fomatter we have?

fewerner: #1521 takes care of this

.github/workflows/rust.yml
```diff
@@ -32,6 +32,7 @@ jobs:
       - uses: actions/checkout@v5
       - uses: ./.github/actions/setup-and-cache-rust
         with:
+          toolchain: nightly-2025-08-24
```
SimonThormeyer, line 35: Why is it necessary to specify this here, shouldn't this be picked up from `rust-toolchain.toml`?

rust-toolchain.toml
```diff
@@ -1,6 +1,13 @@
 [toolchain]
-channel = "1.90"
-components = ["rust-analyzer"]
+channel = "nightly-2025-08-24"
```
coriolinus, line 2: I'm curious why you picked one from over a month ago.

.github/workflows/rust.yml
SimonThormeyer: Can we somehow avoid having to replicate the nightly version here and in `rust-toolchain.toml`?

coriolinus (approved):
I like it, but Simon had a good idea about an auto-updater script. Maybe consider adding that before merging.

SimonThormeyer (approved):
Thanks and good job!

PR #1520, opened 2025-10-02
ci: setup rust directly before calling `make` when building ios

SimonThormeyer:
Setting up rust earlier was causing problems.

# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/.github/workflows/docs-swift.yml
+++ b/.github/workflows/docs-swift.yml
@@ -23,6 +23,9 @@ jobs:
       - name: Install gmake
         run: brew install make
 
+      - name: Install gsed
+        run: brew install gnu-sed
+
       - name: download swift bindings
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-swift.yml
```diff
@@ -23,6 +23,9 @@ jobs:
       - name: Install gmake
         run: brew install make
 
+      - name: Install gsed
+        run: brew install gnu-sed
+
```
fewerner, line 28: I would install this when gsed is used in setup-and-cache-rust action

fewerner: But thinking more about it, this is the only github runner that needs the installation so nevermind

fewerner (approved):
ty

PR #1522, opened 2025-10-02
feat: implement new credential management api [WPB-19572] [WPB-15973]

coriolinus:
# What's new in this PR

- [x] Expose a `Credential` type as a first-class struct. 
- [ ] ~~`Credential` has `ciphersuite` not just `signature_scheme`~~ see [here](https://github.com/wireapp/core-crypto/blob/6dd9b2df4e05f22f2c024bc0fd8487042e472ac7/crypto/src/mls/session/identities.rs#L17-L23): if an identity uses signature schemes as the keys to an identity map, then it cannot have two identities with different ciphersuites which share the same signature scheme. And if that is impossible, then storing only the signature scheme is sufficient.
- [x] basic credential types can be instantiated from all bindings
- [x] tests of creating the credential type from all bindings
  - [x] ts
  - [x] kotlin
  - [x] ~~swift~~
- [x] Add a `CredentialRef` type from which we can get a credential from the db
- [x] Add `Session` methods:
  - [x] `add_credential` <- check the client id matches, stores in keystore
  - [x] `remove_credential` <- must first check that the credential is not used in any conversation and remove both the credential and all key packages that were generated from it
  - [x] `find_credentials` <- optional filters: credential_type, ciphersuite, signature scheme
  - [x] `get_credentials` <- all credentials without filtering
- [x] tests:
  - [x] `add_credential`
    - [x] ts
    - [x] kotlin
    - [x] ~~swift~~
  - [x] `remove_credential`
    - [x] ts
    - [x] kotlin
    - [x] ~~swift~~
  - [x] `find_credentials`
    - [x] ts
    - [x] kotlin
    - [x] ~~swift~~
  - [x] `get_credentials` 
    - [x] ts
    - [x] kotlin
    - [x] ~~swift~~
- [x] rm existing credential API
- [x] adjust existing tests which required old credential API
- [x] changelog

### DIFF
--- a/crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCrypto.swift
+++ b/crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCrypto.swift
@@ -253,7 +253,7 @@ final actor TransactionExecutor<Result>: WireCoreCryptoUniffi.CoreCryptoCommand 
 }
 
 extension Credential {
-    // Construct a new Credential from ciphersuite and client id
+    /// Construct a new Credential from ciphersuite and client id
     public static func basic(
         ciphersuite: Ciphersuite,
         clientId: ClientId


coriolinus:
> I note that Swift tests have been striked through and am wondering about the plan -- are those tests supposed to come in a different PR?

I struck them through fairly early on in the development of this PR on the (mistaken) basis that I thought that we didn't have swift tests. That's clearly not the case, but I can also say that it's a huge pain to actually work with swift tests when the only way to interact with them is via CI. It's not impossible to go back and add Swift tests for this, but I'd really prefer that someone with an apple computer be the one to do so.

istankovic:
> > I note that Swift tests have been striked through and am wondering about the plan -- are those tests supposed to come in a different PR?
> 
> I struck them through fairly early on in the development of this PR on the (mistaken) basis that I thought that we didn't have swift tests. That's clearly not the case, but I can also say that it's a huge pain to actually work with swift tests when the only way to interact with them is via CI. It's not impossible to go back and add Swift tests for this, but I'd really prefer that someone with an apple computer be the one to do so.

I fully sympathise -- this is also one of the reasons I wanted another Mac runner machine, so those of us without a Mac can at least remotely run something. Which, honestly, is not the same as running locally, but at least it's something.

@SimonThormeyer @fewerner would you guys mind helping with Swift tests?

My main concern is that we proceed with the rest of the API, without having covered newly added API on all platforms. Making sure that this API works on all platforms before going to the next API part would also avoid potentially costly scenarios where Swift bindings do not work for some reason or other. Ideally, once the new API epic is done, we will have full coverage on all supported platforms.

crypto/src/mls/credential/mod.rs
```diff
@@ -1,102 +1,232 @@
...
-        self.signature_key.public().hash(state);
-        self.credential().identity().hash(state);
-        match self.credential().mls_credential() {
-            openmls::prelude::MlsCredentialType::X509(cert) => {
+        self.earliest_validity.hash(state);
+        self.signature_key_pair.signature_scheme().hash(state);
+        self.signature_key_pair.public().hash(state);
+        // self.signature_key_pair.private().hash(state); // no two distinct public keys will share a private key so omitting this is ok
+        u16::from(self.credential_type()).hash(state);
+        self.mls_credential.identity().hash(state);
```
istankovic: Is there a reason why we need to hash `identity()`? Similarly below, where we hash certificates... 

I feel it should be sufficient to hash just the `(credential type, signature scheme, signature pubkey)` tuple.

crypto/src/mls/credential/mod.rs
```diff
@@ -1,102 +1,232 @@
...
+        // non-hashed fields here are apparently transient
```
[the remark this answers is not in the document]
coriolinus: It means whatever it means here: https://github.com/wireapp/openmls/blob/c9cde17076508968c9cbead5728454f0a1f60c4f/openmls/src/credentials/mod.rs#L149-L150

I'd have greatly preferred that `openmls::Credential` implement its own `Hash` with its own internal semantics as appropriate, but this impl was mostly inherited (though moved) from the status quo. In the existing implementation, it didn't include the identity as part of the hash, and I don't know enough about its semantics to update that implementation.

Not 100% sure what the hash implementation is used for right now. Not the keystore, that should operate only on the `StoredCredential` type, not this `Credential`. But we want sets of things often enough that hashing makes sense to me.

crypto/src/mls/credential/mod.rs
```diff
@@ -1,102 +1,232 @@
...
+            .then(self.credential_type().cmp(&other.credential_type()))
+            .then_with(|| self.mls_credential.identity().cmp(other.mls_credential.identity()))
+            .then((self.signature_scheme() as u16).cmp(&(other.signature_scheme() as u16)))
+            .then_with(|| self.signature_key_pair.public().cmp(other.signature_key_pair.public()))
     }
 }
-
-impl PartialOrd for CredentialBundle {
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+impl PartialOrd for Credential {
```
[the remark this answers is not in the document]
istankovic: Yeah, I'd prefer to drop both `PartialOrd` and `Ord` impls.

crypto/src/mls/credential/mod.rs
istankovic: It appears that this `Hash` impl is unused. I am struggling to find a use case where we would need the hash of a credential -- for lookups we would anyway use something like the above-mentioned tuple, and in maps credential would be the value, not the key. So I feel we should just remove this `Hash` impl.

crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/MLSTest.kt
```diff
@@ -475,4 +475,101 @@ class MLSTest : HasMockDeliveryService() {
...
+    fun can_add_basic_credential(): TestResult {
+        val scope = TestScope()
+        return scope.runTest {
+            val clientId = genClientId()
+            val credential = Credential.basic(CIPHERSUITE_DEFAULT, clientId)
+
+            val cc = initCc(this@MLSTest)
+            val ref = cc.transaction { ctx ->
+                ctx.mlsInitShort(clientId)
+                ctx.addCredential(credential)
```
SimonThormeyer: I know we're not doing this everywhere else, but for new instances, I think we should be more idiomatic and use
```
{ it } 
```
instead of `{ ctx -> ctx }` in callbacks with a single argument

SimonThormeyer: FYI, I tried setting up enforcing this rule via ktlint, but didn't succeed

crypto/src/mls/credential/keypairs.rs
```diff
@@ -0,0 +1,99 @@
...
+    debug_assert!(
+        {
+            let deserialized =
+                SignatureKeyPair::tls_deserialize_exact(&data).expect("keypair deserializes without error");
+            deserialized.signature_scheme() == keypair.signature_scheme()
+                && deserialized.public() == keypair.public()
+                && deserialized.private() == keypair.private()
+        },
+        "serialized keypair data must deserialize correctly"
+    );
```
istankovic: Why this `debug_assert!`? Feels like "testing the library"...

crypto/src/mls/credential/keypairs.rs
```diff
@@ -0,0 +1,99 @@
...
+/// Ensures the keypairs are sorted in order of their creation date.
+pub(super) async fn load_all(database: &Database) -> Result<Vec<StoredSignatureKeypair>> {
+    database
+        .find_all::<StoredSignatureKeypair>(EntityFindParams::default())
+        .await
+        .map_err(KeystoreError::wrap("finding all mls signature keypairs"))
+        .map_err(Into::into)
+}
+
+/// Generate a new keypair in-memory with the specificed signature scheme
```
istankovic: Typo: `specificed`

coriolinus: https://github.com/wireapp/core-crypto/pull/1522/commits/7c03bfd65df3eb86c91ef152b5d580975332e687

crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/MLSTest.kt
coriolinus: This might just be my own unfamiliarity with Kotlin, but `ctx ->` reads much more obviously as "this is a closure" to me than leaving it implicit and using `it`. I can change it if you feel strongly--I do not feel strongly about this--but intuitively I prefer this version.

Is there some style guide somewhere which definitively says that `it` is preferred?

SimonThormeyer: I don't feel strongly enough to require this to be changed right here and in this PR, I do think we should enforce this via `ktlint` and autofix

crypto/src/mls/credential/mod.rs
istankovic: I'm afraid I disagree. In the above impl you're violating the assumption that if two values are equal, their hashes must also be equal (see https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq). But that is not the main reason I disagree.

Just because we could impl `Hash` for a type does not mean we should. I'm now beginning to suspect that the authors of OpenMLS were on the right track here. In my view, the decision to impl hashing is very much related to type equality, its semantics and usage -- for example, credentials are not simple integers or strings; while it makes sense to look up a string given its hash, for a type representing a credential this is much less clear. My understanding of (new) credential semantics leads me to the feeling that we should not impl `Hash`. The fact that you disagree implies that we have different understanding and it would be good to clarify that (I don't think "cannot put it into a set so must impl Hash" is not a sufficient argument).

coriolinus:
> Why are some keystore structs renamed, and some others are left as they are?

Because I was trying to limit the scope of this PR which is already quite large. I prioritized focus over consistency in this case. 

> for consistent naming, I would expect the following structs to be named

I agree, but I think those are out of scope here. I only renamed the ones I did because it was critical to having sane names for doing this work.

> Generally speaking, if two different credentials have the same public key (even between different credential types), something has very likely gone wrong.

I tend to agree, but this is the kind of thing which led to the entities trait refactor ticket: right now the only obvious way way to perform that check is to load + deserialize every keypair from the keystore and check for matches on the public key, which I wasn't willing to do every time we add a credential.

crypto/src/mls/session/error.rs
```diff
@@ -11,12 +15,12 @@ pub enum Error {
...
     NoX509CertificateBundle,
-    #[error("Tried to insert an already existing CredentialBundle")]
-    CredentialBundleConflict,
+    #[error("credentials must be distinct in signature scheme, credential type, and earliest validity timestamp")]
+    CredentialConflict,
     #[error("A MLS operation was requested but MLS hasn't been initialized on this instance")]
     MlsNotInitialized,
-    #[error("A Credential of type {0:?} was not found locally which is very likely an implementation error")]
-    CredentialNotFound(crate::MlsCredentialType),
+    #[error("No Credential of type ({0:?}, {1:?}) was found in this session")]
```
istankovic: Nit: "Credential" should probably be lowercase

CHANGELOG.md
```diff
@@ -4,6 +4,55 @@
...
+  - Removes the credential from the current working set and also from the keystore.
+- Added a new method to transaction context: `getCredentials` which produces a `CredentialRef` for each credential known
+  by this client
+- Added a new method to transaction context: `findCredentials` which produces a `CredentialRef` for each credential
+  known by this client, efficiently filtering them by the specified criteria.
+
+#### Simplified Session initialization
+
+- Removed the `SessionConfig` and `ValidatedSessionConfig` types
+- A `Session` now needs only a `Database` for initialization
```
istankovic: These session-related changes were all internal, I think we don't need to mention them here.

CHANGELOG.md
```diff
@@ -4,6 +4,55 @@
...
+    - earliest validity
+- To remove a credential from the set MLS knows about, call `removeCredential` on a transaction context, handing it the
+  appropriate `CredentialRef`.
+  - Ensures the credential is not currently in use by any conversation.
+  - Removes all key packages generated from this credential.
+  - Removes the credential from the current working set and also from the keystore.
+- Added a new method to transaction context: `getCredentials` which produces a `CredentialRef` for each credential known
+  by this client
+- Added a new method to transaction context: `findCredentials` which produces a `CredentialRef` for each credential
+  known by this client, efficiently filtering them by the specified criteria.
```
istankovic: This is great stuff!
I think we will be adding a lot more over time and am wondering if it would be better perhaps to have a separate guide "Migration to CoreCrypto 10", which we could link from here.

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
```diff
@@ -78,10 +77,14 @@ final class WireCoreCryptoTests: XCTestCase {
         let clientId = ClientId(bytes: UUID().uuidString.data(using: .utf8)!)
         let ciphersuite = Ciphersuite.mls128Dhkemx25519Chacha20poly1305Sha256Ed25519
 
+        let credential = try credentialBasic(ciphersuite: ciphersuite, clientId: clientId)
```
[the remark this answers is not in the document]
coriolinus: e61cb7fc1

PR #2005, opened 2026-04-01
ci: publish benchmark results

typfel:
# What's new in this PR

Publish benchmark results into our Github Pages in the nightly benchmark run.

----

- The **PR Title**

### DIFF
--- a/.github/workflows/benchmark-jvm.yml
+++ b/.github/workflows/benchmark-jvm.yml
@@ -58,19 +58,13 @@ jobs:
       - name: Run benchmark
         run: make jvm-bench
 
-      # Publish the benchmark result on GitHub pages
-      - name: Publish benchmark result
-        uses: benchmark-action/github-action-benchmark@v1
+      - name: Upload benchmark result
+        uses: actions/upload-artifact@v7.0.0
         with:
-          # Name under which to group the results
-          name: "JVM Benchmarks"
-          # What benchmark tool the output.txt came from
-          tool: 'jmh'
-          # Where the output from the benchmark tool is stored
-          output-file-path: crypto-ffi/bindings/jvm/build/reports/jmh/results.json
-          # Workflow will fail when an alert happens
-          fail-on-alert: true
-          # Access token to deploy GitHub Pages branch
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          # Push and deploy GitHub pages branch automatically
-          auto-push: true
+          name: benchmark-jvm-results
+          path: crypto-ffi/bindings/jvm/build/reports/jmh/results.json
+
+  publish:
+    needs:
+      - benchmark
+    uses: ./.github/workflows/benchmark-jvm-publish.yml
--- a/.github/workflows/benchmark-jvm.yml
+++ b/.github/workflows/benchmark-jvm.yml
@@ -1,0 +1,76 @@
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}-benchmark-jvm"
+  cancel-in-progress: true
+
+on:
+  workflow_call:
+
+env:
+  RELEASE: 1
+
+permissions:
+  # deployments permission to deploy GitHub pages website
+  deployments: write
+  # contents permission to update benchmark contents in gh-pages branch
+  contents: write
+
+jobs:
+  benchmark:
+    name: Performance regression check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: set up jdk 17
+        uses: actions/setup-java@v5
+        with:
+          java-version: "17"
+          distribution: "adopt"
+      - name: gradle setup
+        uses: gradle/actions/setup-gradle@v6
+        with:
+          cache-provider: basic
+      - name: validate gradle wrapper
+        uses: gradle/actions/wrapper-validation@v5
+
+      - name: download linux library
+        uses: ./.github/actions/make/jvm
+        with:
+          gh-token: ${{ secrets.GITHUB_TOKEN }}
+          artifact-generation: ${{ vars.ARTIFACT_GENERATION }}
+
+      # The specific jvm action uses `uname -s` to determine which `jvm` make rule to call.
+      # Because this workflow runs on ubuntu, we're using the generic action
+      - name: download aarch64 apple darwin artifact
+        uses: ./.github/actions/make
+        with:
+          key: jvm-darwin
+          make-rule: jvm-darwin
+          target-path: target/aarch64-apple-darwin/release/libcore_crypto_ffi.dylib
+          gh-token: ${{ secrets.GITHUB_TOKEN }}
+          artifact-generation: ${{ vars.ARTIFACT_GENERATION }}
+
+      - name: download kotlin bindings
+        uses: ./.github/actions/make/bindings-kotlin-jvm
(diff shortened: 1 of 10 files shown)


.github/workflows/benchmark-jvm.yml
```diff
@@ -0,0 +1,103 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: set up jdk 17
+        uses: actions/setup-java@v5
+        with:
+          java-version: "17"
+          distribution: "adopt"
+      - name: gradle setup
+        uses: gradle/actions/setup-gradle@v5
```
istankovic: We've now moved to v6 of `setup-gradle`, with `basic` cache provider. Please see https://github.com/wireapp/core-crypto/pull/1979 for context.

.github/workflows/benchmarks_new.yml
```diff
@@ -0,0 +1,48 @@
+name: benchmarks (new)
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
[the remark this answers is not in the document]
typfel: This was during development, but doesn't it make sense to run this for every time we merge to main to have a consistent history.

.github/workflows/benchmarks_new.yml
```diff
@@ -0,0 +1,48 @@
...
+jobs:
+  benchmark:
+    name: Performance regression check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: "stable"
+      # Run benchmark with `go test -bench` and stores the output to a file
```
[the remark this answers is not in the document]
typfel: Cleaned up.

.github/workflows/benchmarks_new.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Run benchmark
+        run: make jvm-bench
+      # Download previous benchmark result from cache (if exists)
+      - name: Download previous benchmark data
+        uses: actions/cache@v4
+        with:
+          path: ./cache
+          key: ${{ runner.os }}-benchmark
```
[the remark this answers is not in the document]
typfel: Deleted this because we don't any cached results if we are just publishing results

.github/workflows/benchmark-jvm.yml
typfel: Updated

.github/workflows/benchmark-jvm.yml
```diff
@@ -0,0 +1,76 @@
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}-benchmark-jvm"
+  cancel-in-progress: true
+
+on:
+  workflow_call:
+
+env:
+  RELEASE: 1
```
coriolinus, line 9: I don't think this will affect anything.

.github/workflows/benchmark-ts-browser.yml
```diff
@@ -0,0 +1,59 @@
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}-benchmark-ts-browser"
+  cancel-in-progress: true
+
+on:
+  workflow_call:
+
+env:
+  RELEASE: 1
```
coriolinus, line 9: Same here.

.github/workflows/pipeline.yml
```diff
@@ -213,6 +213,12 @@ jobs:
           gh-token: ${{ secrets.GITHUB_TOKEN }}
           artifact-generation: ${{ vars.ARTIFACT_GENERATION }}
 
+  benchmark-ts-browser:
+    if: github.branch == 'main' && github.repository == 'wireapp/core-crypto'
+    needs:
+      - bindings-ts-browser
+    uses: ./.github/workflows/benchmark-ts-browser.yml
+
```
coriolinus, line 221: Do we _want_ to run benches on each PR? Right now they're all only running overnight, because they're just kind of time-consuming.

.github/workflows/pipeline.yml
```diff
@@ -114,6 +114,14 @@ jobs:
...
     uses: ./.github/workflows/test-jvm.yml
 
+  benchmark-jvm:
+    if: github.branch == 'main' && github.repository == 'wireapp/core-crypto'
+    needs:
+      - bindings-jvm
+      - build-jvm-linux
+      - build-jvm-darwin
+    uses: ./.github/workflows/benchmark-jvm.yml
+
```
coriolinus, line 124: Same here. Would it be worth running these in the nightly run instead?

coriolinus (approved):
I'll approve as-is because it's not wrong, but we'd talked about doing benching on nightly runs both for the unloaded/consistent machines and because they can be time-consuming. IMO we should move the triggers for these actions out of the pipeline and into those nightly runs.

.github/workflows/pipeline.yml
typfel: Sure we could but right now they run on a cloud ubuntu runner which is fairly cheap.

.github/workflows/benchmark-jvm.yml
typfel: AFAIK it affects if we look the debug or release version of the ffi library when building the kotlin bindings, and we are downloading release ffi library from the cache.

.github/workflows/pipeline.yml
istankovic: > Sure we could but right now they run on a cloud ubuntu runner which is fairly cheap.

Note that allocating a runner for this takes it away from the pool of runners available for all other workflows -- the runner pool is not infinite. Moreover, I expect there to be greater variance in bench results with github-hosted runners compared to our own.

So I'm with Peter here: ideally this should run overnight and use our own runner.

.github/workflows/benchmarks_new.yml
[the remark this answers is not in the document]
coriolinus: Nightly runs still give us a consistent history IMO. Our dev pace isn't so fast that having a 24h block will make it particularly hard to find where some regression came from.

I'm focused more on DX during dev than the potential for debugging though. Right now CI is pretty fast, and I just want it to stay like that.

.github/workflows/pipeline.yml
typfel: Updated to run the benchmark on our self hosted runner in the nightly.

.github/workflows/pipeline.yml
typfel: Done

crypto-ffi/bindings/js/packages/browser/benches/processMessage.bench.ts
```diff
@@ -124,5 +125,18 @@ describe("benchmark", () => {
 
         console.log(results.name);
         console.log(results.table);
+
+        if (!process.env["CI"]) return;
```
coriolinus, line 129: So if we run the benchmarks locally we don't get any postprocessing?

coriolinus (approved):
Things look good!

crypto-ffi/bindings/js/packages/browser/benches/processMessage.bench.ts
typfel: Right, you don't get a json file with the result. The results are also printed in the terminal

PR #2021, opened 2026-04-13
docs: improve ffi docs strings [WPB-21686]

coriolinus:
# What's new in this PR

Docs-only changes.

Got Claude to make changes according to 4 rules:

1. No bare "see X"
2. No non-trivial markdown or Rust internal links
3. First line is a single-sentence summary
4. Use proper grammar and punctuation except for trivial labels and descriptors.

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/src/client_id.rs
+++ b/crypto-ffi/src/client_id.rs
@@ -1,8 +1,8 @@
-/// A Client identifier
+/// A unique identifier for an MLS client.
 ///
-/// A unique identifier for clients. A client is an identifier for each App a user is using, such as desktop,
-/// mobile, etc. Users can have multiple clients.
-/// More information [here](https://messaginglayersecurity.rocks/mls-architecture/draft-ietf-mls-architecture.html#name-group-members-and-clients)
+/// Each app instance a user is running, such as desktop or mobile, is a separate client
+/// with its own client id. A single user may therefore have multiple clients.
+/// More information: <https://messaginglayersecurity.rocks/mls-architecture/draft-ietf-mls-architecture.html#name-group-members-and-clients>
 #[derive(Debug, Clone, Eq, Hash, PartialEq, derive_more::From, uniffi::Object)]
 pub struct ClientId(pub(crate) core_crypto::ClientId);
 
--- a/crypto-ffi/src/core_crypto/conversation.rs
+++ b/crypto-ffi/src/core_crypto/conversation.rs
@@ -31,7 +31,7 @@ impl AsRef<ConversationIdRef> for ConversationId {
 
 #[uniffi::export]
 impl CoreCryptoFfi {
-    /// See [core_crypto::mls::conversation::Conversation::epoch]
+    /// Returns the current MLS epoch of the given conversation.
     pub async fn conversation_epoch(&self, conversation_id: &ConversationId) -> CoreCryptoResult<u64> {
         let conversation = self
             .inner
@@ -45,7 +45,7 @@ impl CoreCryptoFfi {
         Ok(conversation.epoch().await)
     }
 
-    /// See [core_crypto::mls::conversation::Conversation::ciphersuite]
+    /// Returns the ciphersuite in use for the given conversation.
     pub async fn conversation_ciphersuite(&self, conversation_id: &ConversationId) -> CoreCryptoResult<Ciphersuite> {
         let cs = self
             .inner
@@ -77,7 +77,7 @@ impl CoreCryptoFfi {
             .map_err(Into::into)
     }
 
-    /// See [core_crypto::Session::conversation_exists]
+    /// Returns true if a conversation with the given id exists in the local state.
     pub async fn conversation_exists(&self, conversation_id: &ConversationId) -> CoreCryptoResult<bool> {
         self.inner
             .mls_session()
@@ -88,7 +88,7 @@ impl CoreCryptoFfi {
             .map_err(Into::into)
     }
 
-    /// See [core_crypto::mls::conversation::Conversation::get_client_ids]
+    /// Returns the client ids of all members of the given conversation.
     pub async fn get_client_ids(&self, conversation_id: &ConversationId) -> CoreCryptoResult<Vec<Arc<ClientId>>> {
         let conversation = self
             .inner
@@ -106,7 +106,7 @@ impl CoreCryptoFfi {
             .collect())
     }
 
-    /// See [core_crypto::mls::conversation::Conversation::get_external_sender]
+    /// Returns the serialized public key of the external sender for the given conversation.
     pub async fn get_external_sender(&self, conversation_id: &ConversationId) -> CoreCryptoResult<Vec<u8>> {
         let conversation = self
             .inner
@@ -120,7 +120,11 @@ impl CoreCryptoFfi {
         conversation.get_external_sender().await.map_err(Into::into)
     }
 
-    /// See [core_crypto::mls::conversation::Conversation::export_secret_key]
+    /// Derives and exports a secret of `key_length` bytes for the given conversation.
+    ///
+    /// The secret is derived from the MLS key schedule's exporter mechanism (RFC 9420 §8.5),
+    /// which produces output bound to the current group state and epoch. The exported value
+    /// changes whenever the epoch advances.
     pub async fn export_secret_key(
         &self,
         conversation_id: &ConversationId,
@@ -139,7 +143,7 @@ impl CoreCryptoFfi {
             .map_err(Into::into)
     }
 
-    /// See [core_crypto::mls::conversation::Conversation::is_history_sharing_enabled]
+    /// Returns true if history sharing is currently enabled for the given conversation.
     pub async fn is_history_sharing_enabled(&self, conversation_id: &ConversationId) -> CoreCryptoResult<bool> {
         let conversation = self
             .inner
--- a/crypto-ffi/src/core_crypto_context/e2ei.rs
+++ b/crypto-ffi/src/core_crypto_context/e2ei.rs
@@ -9,7 +9,7 @@ use crate::{
 
 #[uniffi::export]
 impl CoreCryptoContext {
-    /// See [core_crypto::mls::conversation::Conversation::e2ei_conversation_state]
+    /// Returns the end-to-end identity verification state of the given conversation.
     pub async fn e2ei_conversation_state(
         &self,
         conversation_id: &ConversationId,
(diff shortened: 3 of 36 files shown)


crypto-ffi/src/credential_type.rs
```diff
@@ -1,11 +1,11 @@
-/// Type of Credential
+/// The type of credential used to authenticate an MLS client's identity.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default, uniffi::Enum)]
 #[repr(u8)]
 pub enum CredentialType {
-    /// Basic credential i.e. a KeyPair
+    /// A basic credential backed by a raw public key, without any certificate infrastructure.
```
istankovic: Nit: I'd rather use "raw key pair" (preferred) or "raw private key" instead of "raw public key", which gives the impression that the credential contains _only_ the public part.

crypto-ffi/src/signature_scheme.rs
```diff
@@ -6,23 +6,23 @@
...
+/// Signature schemes supported by MLS, as defined in RFC 9420.
 #[allow(non_camel_case_types)]
 #[allow(clippy::upper_case_acronyms)]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, derive_more::TryFrom)]
 #[try_from(repr)]
 #[repr(u16)]
 #[derive(uniffi::Enum)]
 pub enum SignatureScheme {
-    /// ECDSA_SECP256R1_SHA256
+    /// ECDSA with secp256r1 (P-256) and SHA-256; signature is DER-encoded per RFC 8446.
```
istankovic: I wouldn't mention signature encoding at all here. What is the value of that?
Also, the comments points to RFC 8446, "The Transport Layer Security (TLS) Protocol Version 1.3", which I'm failing to see it being related to what we're doing.

coriolinus: The signature scheme defines the encoding and it varies according to which signature scheme is chosen. The encoding is a relevant part of the scheme. Specifically, the reference to RFC 8446 came from [here](https://www.rfc-editor.org/rfc/rfc9420.html#section-5.1.1-6). 

I figured that just duplicating the name of the enum's variant was pointless. So I wanted to go into a little bit more detail about what each of these things implied.

crypto-ffi/src/credential_type.rs
coriolinus: Sure, fc8826bc8

crypto-ffi/src/signature_scheme.rs
istankovic: I would say that signature encoding is probably the least interesting thing about a signature scheme.  :-)

This documentation is going to end up in our public API, right? I do not see how this helps clients. They should neither care nor have a way to affect the signatures and their encoding. I think we should just drop this part from the comments.

crypto-ffi/src/credential_ref.rs
```diff
@@ -2,29 +2,28 @@ use core_crypto::CredentialRef as CryptoCredentialRef;
 
 use crate::{Ciphersuite, ClientId, CredentialType, SignatureScheme};
 
-/// A reference to a credential which has been persisted in CC.
+/// A compact reference to a credential that has been persisted in CoreCrypto.
```
istankovic, line 5: I like this phrasing: "compact reference"! 👍

crypto-ffi/src/client_id.rs
```diff
@@ -1,8 +1,8 @@
-/// A Client identifier
+/// A unique identifier for an MLS client.
 ///
-/// A unique identifier for clients. A client is an identifier for each App a user is using, such as desktop,
-/// mobile, etc. Users can have multiple clients.
-/// More information [here](https://messaginglayersecurity.rocks/mls-architecture/draft-ietf-mls-architecture.html#name-group-members-and-clients)
+/// Each app instance a user is running, such as desktop or mobile, is a separate client
```
istankovic, line 3: If I run two Wire desktop clients at the same time (= 2 Wire client processes), are those two really different clients, meaning they have different client IDs? I would expect they would try to use the same db (not sure how that ends -- do you know perhaps?).

coriolinus: The first answer is that CC doesn't try to use one database or another; it just uses what you pass it. 

If you run two instances with different databases, they are different clients; that much is clear. If you try to run two instances with the same database, I don't know precisely what will happen. My expectation is that it'll be effectively the same client, but you'll probably run into weird errors, because that's not a supported configuration.

Do you feel like this should be explained here?

istankovic: Honestly, I'm not sure. I also think this is Good Enough(tm) as-is, so feel free to ignore this comment and what follows.

I'm not even sure how I would explain it without writing a bunch of text.
Perhaps a better way to convey what client IDs are may be from the backend perspective? That is,  the Wire backend manages clients and assigns client IDs. Apps are merely consumers of the db and by sharing the db, they share the client ID, keys etc. (though, in practice that shouldn't really be the case, so maybe we should not even mention it...). Maybe just trying to avoid mentioning "apps" would be enough of an improvement....

crypto-ffi/src/signature_scheme.rs
coriolinus: https://github.com/wireapp/core-crypto/pull/2021/commits/ba70b33df7c577d4bf4034a48b725cf742144b70

crypto-ffi/src/identity/wire.rs
```diff
@@ -30,17 +31,19 @@ impl From<core_crypto::WireIdentity> for WireIdentity {
...
-/// This does not represent the states where a device is not using MLS or is not using end-to-end identity
+/// This does not represent states where a device is not using MLS or end-to-end identity.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, uniffi::Enum)]
 #[repr(u8)]
 pub enum DeviceStatus {
-    /// All is fine
+    /// The device credential is valid.
     Valid = 1,
-    /// The Credential's certificate is expired
+    /// The device credential's certificate has expired.
```
istankovic, line 42: Nit: while technically true, device status is credential-type-agnostic so I'd prefer more something like 
```
    /// The device credential has expired.
```
to not imply the underlying credential type.

crypto-ffi/src/identity/wire.rs
```diff
@@ -30,17 +31,19 @@ impl From<core_crypto::WireIdentity> for WireIdentity {
...
     Expired = 2,
-    /// The Credential's certificate is revoked (not implemented yet)
+    /// The device credential's certificate has been revoked.
```
istankovic, line 44: Nit: similarly to above:
```
    /// The device credential has been revoked.
```

crypto-ffi/src/pki_env.rs
```diff
@@ -103,8 +105,9 @@ impl From<uniffi::UnexpectedUniFFICallbackError> for PkiEnvironmentHooksError {
     }
 }
 
-/// The PKI Environment Hooks used for external calls during e2e enrollment flow.
-/// When communicating with the Identity Provider (IDP)  and Wire server,
+/// Callbacks for external calls made by CoreCrypto during the end-to-end identity enrollment flow.
```
istankovic: Since we're here and this is talking about the new acquisition process, let's use the opportunity to eliminate things like "end-to-end identity" and "enrollment":

```
/// Callbacks for external calls made by CoreCrypto during X509 credential acquisition.
```

crypto-ffi/src/pki_env.rs
```diff
@@ -217,13 +221,13 @@ impl pki_env::hooks::PkiEnvironmentHooks for PkiEnvironmentHooksShim {
     }
 }
 
-/// A PkiEnvironment
+/// The PKI environment used for certificate management during end-to-end identity enrollment.
```
istankovic: How about
```
/// The PKI environment used for certificate management, X509 credential acquisition and
/// credential verification.
```
?

crypto-ffi/src/core_crypto_context/mls.rs
```diff
@@ -117,7 +117,11 @@ impl CoreCryptoContext {
         Ok(client_ids)
     }
 
-    /// See [core_crypto::mls::conversation::Conversation::export_secret_key]
+    /// Derives and exports a secret of `key_length` bytes for the given conversation.
+    ///
+    /// The secret is derived from the MLS key schedule's exporter mechanism (RFC 9420 §8.5),
+    /// which produces output bound to the current group state and epoch. The exported value
+    /// changes whenever the epoch advances, so callers should re-export after each commit.
```
istankovic: Nit: I'd omit the second part of the last sentence ("so callers should re-export after each commit"), because that really depends on the caller's use case and security requirements. I think it's enough to mention that the exported value changes as the epoch advances.

crypto-ffi/src/identity/wire.rs
coriolinus: I don't think basic credentials can expire though? This status _at all_ implies X509.

crypto-ffi/src/identity/wire.rs
coriolinus: But it's the same thing: we don't have a mechanism in place to revoke a basic credential, I'm pretty sure.

crypto-ffi/src/core_crypto_context/e2ei.rs
```diff
@@ -62,7 +62,7 @@ impl CoreCryptoContext {
         Ok(user_ids)
     }
 
-    /// See [core_crypto::Session::e2ei_is_pki_env_setup]
+    /// Returns true if the PKI environment has been set up and its provider is configured.
```
istankovic: Nit: I would drop the "and its provider is configured" part as it is an internal thing (and it's likely to go away soon, with any luck).

crypto-ffi/src/core_crypto/conversation.rs
```diff
@@ -120,7 +120,11 @@ impl CoreCryptoFfi {
         conversation.get_external_sender().await.map_err(Into::into)
     }
 
-    /// See [core_crypto::mls::conversation::Conversation::export_secret_key]
+    /// Derives and exports a secret of `key_length` bytes for the given conversation.
+    ///
+    /// The secret is derived from the MLS key schedule's exporter mechanism (RFC 9420 §8.5),
+    /// which produces output bound to the current group state and epoch. The exported value
+    /// changes whenever the epoch advances, so callers should re-export after each commit.
```
istankovic: Nit: similar comment as in the other place documenting `export_secret_key`.

istankovic (approved):
Some nits, but nothing blocking. Thanks for doing this!

PR #2024, opened 2026-04-14
build: improve `Makefile` by better splitting wasm from ts builds [WPB-23808]

coriolinus:
# What's new in this PR

----

- The **PR Title**


make/ts.mk
```diff
@@ -54,28 +56,29 @@ ts-clean: ## Cleanup old TypeScript build outputs
 	&& rm -rf $(WASM_TARGET_DIR) \
 	&& rm -rf $(JS_DIR)/rust_modules
 
+ubrn-deps := $(RUST_SOURCES)
+$(WASM_FFI_LIB) $(BROWSER_TS_IMPL) &: $(ubrn-deps)
```
SimonThormeyer: what does the `&` do again and why don't we have it in other places?

coriolinus: `&` here says "all this stuff got generated by the same invocation; you just need this step once, not N times". 

We don't need it wherever there's a single target. But IMO it should be our default. (It should be `make`'s default, but you can't change the defaults in a tool older than me.)

SimonThormeyer (commented):
Not sure if CI failure implies that there are still some changes missing or not; feel free to ping me again once CI is green

SimonThormeyer (approved):
Looks good, pending green CI.

PR #2028, opened 2026-04-15
feat!: drop mark_conversation_as_child_of

istankovic:
No client is actually using this.


SimonThormeyer (approved):
Thanks for doing this!

PR #2029, opened 2026-04-15
chore: upgrade dependencies to make `cargo deny` happier

coriolinus:
# What's new in this PR

Note that we can't upgrade `proteus-wasm` from `v2.1.1.` to `v2.1.3` because `v2.1.2` updates its rand-core to v0.9.x which we are still currently blocked from upgrading to.

----

- The **PR Title**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4958,9 +4958,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.10"
+version = "0.103.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


istankovic (approved):

istankovic:
I note that we're using both password-hash@0.5.0 and password-hash@0.6.1 -- would it be possible to use just one?

coriolinus:
> I note that we're using both password-hash@0.5.0 and password-hash@0.6.1 -- would it be possible to use just one?

0.5.0 is an indirect dependency of `argon2`, which is at its most recent version.

PR #2044, opened 2026-04-20
ci: exercise benchmarks on PRs [WPB-24082]

typfel:
# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/.github/workflows/benchmark-jvm.yml
+++ b/.github/workflows/benchmark-jvm.yml
@@ -57,7 +57,7 @@ permissions:
 jobs:
   benchmark:
     name: Performance regression check
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, benchmark]
     env:
       BENCHMARK_MESSAGE_COUNTS: ${{ inputs.benchmark_message_counts }}
       BENCHMARK_MESSAGE_SIZES: ${{ inputs.benchmark_message_sizes }}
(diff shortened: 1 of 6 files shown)


SimonThormeyer:
@typfel could you please rebase this so CI starts running?

typfel:
> @typfel could you please rebase this so CI starts running?

Sure, but it's not ready yet was just testing things on CI. I'll move it back to draft.

fewerner (commented):
I like it a lot. GJ!

I only have one point. Not being able to set the ciphersuite but every other parameter feels incomplete. I do think that this is a use case especially for profiling.
For PR runs it would be enough to run a single ciphersuite to save us some time imo.

If you agree with me, I would like you to add the overriding of the ciphersuite param. It might be better UX if we can provide the number of the ciphersuite in cli.

typfel:
> I like it a lot. GJ!
> 
> I only have one point. Not being able to set the ciphersuite but every other parameter feels incomplete. I do think that this is a use case especially for profiling. For PR runs it would be enough to run a single ciphersuite to save us some time imo.
> 
> If you agree with me, I would like you to add the overriding of the ciphersuite param. It might be better UX if we can provide the number of the ciphersuite in cli.

Yeah makes sense I'll also add the option override the ciphersuites

fewerner (approved):
LGTM! Thank you!

PR #2053, opened 2026-04-22
refactor!: remove mls transport response [WPB-24734]

SimonThormeyer:
# What's new in this PR
We don't need the `Retry` mechanism anymore. Removing it from the
response enum effectively turns this enum into a `Result`. We're already
returning a result from the function. In case of failures (which
includes rejection by the DS) implementors of this trait can just
construct and throw an error of the correct type.

----

- The **PR Title**


crypto/src/mls/conversation/conversation_guard/commit.rs
```diff
@@ -52,46 +38,14 @@ impl ConversationGuard {
...
-                    let Some(commit_to_retry) = self.commit_pending_proposals_inner().await? else {
-                        // The intended operation was already done in one of the merged commits.
-                        return Ok(TransportedCommitPolicy::None);
-                    };
-                    commit = commit_to_retry;
-                }
-            }
-        }
+        transport
+            .send_commit_bundle(commit.clone())
```
coriolinus: Why clone here?

coriolinus (approved):
Great simplification!

crypto/src/mls/conversation/conversation_guard/commit.rs
SimonThormeyer: Good catch! It's not necessary anymore

PR #2074, opened 2026-04-24
Jacob/feat/acquisition flow bindings

typfel:
# What's new in this PR

Expose the acquisition flow in the bindings

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/bindings/js/packages/browser/test/e2ei.test.ts
+++ b/crypto-ffi/bindings/js/packages/browser/test/e2ei.test.ts
@@ -77,6 +77,46 @@ describe("PKI environment", () => {
 });
 
 describe("end to end identity", () => {
+    it("should instantiate an x509 credential acquisition object", async () => {
+        const acquisitionCreated = await browser.execute(async () => {
+            const key = new Uint8Array(32);
+            window.crypto.getRandomValues(key);
+            const database = await window.ccModule.Database.open(
+                crypto.randomUUID(),
+                new window.ccModule.DatabaseKey(key.buffer)
+            );
+            const pkiEnvironment = await window.ccModule.PkiEnvironment.create(
+                window.pkiEnvironmentHooks,
+                database
+            );
+
+            const clientId = window.helpers.newClientId(
+                "LcksJb74Tm6N12cDjFy7lQ:8e6424430d3b28be@world.com"
+            );
+            const config =
+                window.ccModule.X509CredentialAcquisitionConfiguration.new({
+                    acmeUrl: "acme.example.com",
+                    idpUrl: "https://idp.example.com",
+                    ciphersuite: window.defaultCipherSuite,
+                    displayName: "Alice Smith",
+                    clientId,
+                    handle: "alice_wire",
+                    domain: "world.com",
+                    team: undefined,
+                    validityPeriodSecs: BigInt(3600),
+                });
+
+            const acquisition = new window.ccModule.X509CredentialAcquisition(
+                pkiEnvironment,
+                config
+            );
+
+            return acquisition !== undefined;
+        });
+
+        await expect(acquisitionCreated).toBe(true);
+    });
+
     it("should not be enabled on conversation with basic credential", async () => {
         const conversationState = await browser.execute(async () => {
             const cc = await window.helpers.ccInit();
--- a/crypto-ffi/bindings/js/packages/browser/test/e2ei.test.ts
+++ b/crypto-ffi/bindings/js/packages/browser/test/e2ei.test.ts
@@ -79,12 +79,7 @@ describe("PKI environment", () => {
 describe("end to end identity", () => {
     it("should instantiate an x509 credential acquisition object", async () => {
         const acquisitionCreated = await browser.execute(async () => {
-            const key = new Uint8Array(32);
-            window.crypto.getRandomValues(key);
-            const database = await window.ccModule.Database.open(
-                crypto.randomUUID(),
-                new window.ccModule.DatabaseKey(key.buffer)
-            );
+            const database = await window.helpers.newDatabase();
             const pkiEnvironment = await window.ccModule.PkiEnvironment.create(
                 window.pkiEnvironmentHooks,
                 database
--- a/crypto-ffi/src/e2ei/mod.rs
+++ b/crypto-ffi/src/e2ei/mod.rs
@@ -1,3 +1,12 @@
+use std::sync::Arc;
+
+use async_lock::Mutex;
+use jwt_simple::prelude::{ES256KeyPair, ES384KeyPair, ES512KeyPair, Ed25519KeyPair};
+use wire_e2e_identity::{HashAlgorithm, JwsAlgorithm};
+use x509_cert::der::Encode as _;
+
+use crate::{Ciphersuite as FfiCiphersuite, ClientId, CoreCryptoError, CoreCryptoResult, Credential, PkiEnvironment};
+
 /// The end-to-end identity verification state of a conversation.
 ///
 /// Note: this does not check pending state (pending commit, pending proposals), so it does not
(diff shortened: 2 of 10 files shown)


crypto-ffi/src/e2ei/mod.rs
```diff
@@ -24,3 +31,184 @@ impl From<core_crypto::E2eiConversationState> for E2eiConversationState {
...
+pub enum E2eiJwsAlgorithm {
+    /// Ed25519 / EdDSA
+    Ed25519,
+    /// ECDSA P-256 / ES256
+    P256,
+    /// ECDSA P-384 / ES384
+    P384,
+    /// ECDSA P-521 / ES512
+    P521,
+}
```
istankovic: This is a bit of a mess right now -- not your PR, but the general situation with `JwsAlgorithm` and similar types.

The simple thing first: I would avoid saying "E2EI" or "end-to-end identity" in new API, that's a thing from the past and we would like to move away from that terminology. Instead, we should use credential-type names, e.g. "X509Credential" or similar, that make it much more explicit what the API is about.

That said, here I would perhaps just use "JwsAlgorithm", though even that is not ideal as the enum values are just curve names, not proper JWS algorithm names (I thought about using `Curve` but I don't think that is really sufficient in this context). It's really a mess and to resolve it properly we would need to rework the way we handle `JwsAlgorithm`, config and related types (for that we have a item already, I think created by Felix). 😞

crypto-ffi/src/e2ei/mod.rs
```diff
@@ -24,3 +31,184 @@ impl From<core_crypto::E2eiConversationState> for E2eiConversationState {
...
+            E2eiJwsAlgorithm::P256 => Self::P256,
+            E2eiJwsAlgorithm::P384 => Self::P384,
+            E2eiJwsAlgorithm::P521 => Self::P521,
+        }
+    }
+}
+
+/// Hash algorithm used for E2EI certificate acquisition.
+#[derive(Debug, Copy, Clone, uniffi::Enum)]
+pub enum E2eiHashAlgorithm {
```
istankovic: Similarly, here, I would not use the `E2ei` prefix. Just `HashAlgorithm` should suffice and it may perhaps come useful in the future in other places.

crypto-ffi/src/e2ei/mod.rs
```diff
@@ -24,3 +31,184 @@ impl From<core_crypto::E2eiConversationState> for E2eiConversationState {
...
+            client_id,
+            handle: self.handle,
+            domain: self.domain,
+            team: self.team,
+            validity_period: std::time::Duration::from_secs(self.validity_period_secs),
+        })
+    }
+}
+
+/// Result of a completed E2EI X509 credential acquisition flow.
```
istankovic: Here we also mention "E2EI", which should be avoided.

crypto-ffi/src/e2ei/mod.rs
```diff
@@ -24,3 +31,184 @@ impl From<core_crypto::E2eiConversationState> for E2eiConversationState {
...
+#[derive(Debug, Clone, uniffi::Record)]
+pub struct X509CredentialAcquisitionResult {
+    /// PEM-encoded signing keypair for the new credential.
+    pub signing_key_pem: String,
+    /// PEM-encoded certificate chain, starting with the end-entity certificate.
+    pub certificate_chain_pem: Vec<String>,
+}
+
+/// Initial state of the E2EI X509 credential acquisition flow.
```
istankovic: Similarly here.

crypto-ffi/src/e2ei/mod.rs
```diff
@@ -24,3 +31,184 @@ impl From<core_crypto::E2eiConversationState> for E2eiConversationState {
...
+}
+
+/// Result of a completed E2EI X509 credential acquisition flow.
+#[derive(Debug, Clone, uniffi::Record)]
+pub struct X509CredentialAcquisitionResult {
+    /// PEM-encoded signing keypair for the new credential.
+    pub signing_key_pem: String,
+    /// PEM-encoded certificate chain, starting with the end-entity certificate.
+    pub certificate_chain_pem: Vec<String>,
+}
```
istankovic: We should not be exposing this. From the perspective of a CoreCrypto user, the acquisiton flow results in a valid credential, not the (signing key, certificate chain) pair. That is what is going under the hood, as an implementation detail, but the caller of our API should never see the signing key, only the (opaque) credential, that can then be added to the client, used in a conversation etc.

crypto-ffi/src/e2ei/mod.rs
```diff
@@ -24,3 +31,184 @@ impl From<core_crypto::E2eiConversationState> for E2eiConversationState {
...
+
+/// Initial state of the E2EI X509 credential acquisition flow.
+#[derive(uniffi::Object)]
+pub struct X509CredentialAcquisition {
+    inner: Mutex<Option<wire_e2e_identity::X509CredentialAcquisition>>,
```
istankovic: Why is `Option` needed here?

crypto-ffi/src/e2ei/mod.rs
```diff
@@ -24,3 +31,184 @@ impl From<core_crypto::E2eiConversationState> for E2eiConversationState {
...
+        Ok(Arc::new(X509CredentialAcquisitionDpopChallengeCompleted {
+            inner: Mutex::new(Some(inner)),
+        }))
+    }
+}
+
+#[uniffi::export]
+impl X509CredentialAcquisitionDpopChallengeCompleted {
+    /// Complete the OIDC challenge and return the acquired signing key and certificate chain.
+    pub async fn complete_oidc_challenge(&self) -> CoreCryptoResult<X509CredentialAcquisitionResult> {
```
istankovic: Heh, it is not intended that clients know about the underlying process details, such as which challenges we have and to call them. Please see the RFC CC2 and in particular how `.finalize()` is supposed to be used. The state we're in should be tracked internally.

crypto-ffi/src/e2ei/mod.rs
typfel: Alright, renamed it to just JwsAlgorithm and removed any references to e2ei.

crypto-ffi/src/e2ei/mod.rs
typfel: Gone

crypto-ffi/src/e2ei/mod.rs
typfel: Gone

crypto-ffi/src/e2ei/mod.rs
typfel: Gone

crypto-ffi/src/e2ei/mod.rs
typfel: Right of course! Wasn't thinking initially and just exposed everything as is, but why isn't the result type a credential in acquisition flow?

crypto-ffi/src/e2ei/mod.rs
typfel: The intention was to remove the `X509CredentialAcquisition` once the flow completes so that you can't call finalize twice. But I guess the idea was to be able to resume from an intermediate state after deserializing the acquisition in the web case so this doesn't work.

crypto-ffi/src/e2ei/mod.rs
istankovic: The problem is that we can't do that right now without introducing a circular dependency because the credential type is currently defined in the `crypto` crate. Ideally, we would have the credential type in the `e2e-identity` crate (I'd like to rename that crate too, but that's not critical for CC10 as it is an impl detail), however moving it would probably not be a small change. In a way, the situation is similar to the one with the keystore trait definitions. Right now for `crypto-ffi` this should not be a problem because it depends on both `e2e-identity` and `crypto` crates.

crypto-ffi/src/e2ei/mod.rs
typfel: Replaced the option with an enum which would need to be extended when add support for serializing / deserializing.

crypto-ffi/src/e2ei/mod.rs
typfel: Cleaned this up

crypto-ffi/bindings/js/packages/browser/test/e2ei.test.ts
```diff
@@ -91,6 +91,48 @@ describe("PKI environment", () => {
...
+            const pkiEnvironment = await window.ccModule.PkiEnvironment.create(
+                window.pkiEnvironmentHooks,
+                database
+            );
+
+            const clientId = new window.ccModule.ClientId(
+                new TextEncoder().encode(
+                    "LcksJb74Tm6N12cDjFy7lQ:8e6424430d3b28be@world.com"
+                ).buffer
+            );
```
fewerner: Use `window.helpers.newClientId("LcksJb74Tm6N12cDjFy7lQ:8e6424430d3b28be@world.com")`

crypto-ffi/bindings/js/packages/browser/test/e2ei.test.ts
```diff
@@ -91,6 +91,48 @@ describe("PKI environment", () => {
...
 
 describe("end to end identity", () => {
+    it("should instantiate an x509 credential acquisition object", async () => {
+        const acquisitionCreated = await browser.execute(async () => {
+            const key = new Uint8Array(32);
+            window.crypto.getRandomValues(key);
+            const database = await window.ccModule.Database.open(
+                crypto.randomUUID(),
+                new window.ccModule.DatabaseKey(key.buffer)
+            );
```
fewerner: We do this logic over and over in our tests. Do you want to add a `newDatabase` helper to `Helpers`?
You could also replace the logic in `ccInit`

fewerner (commented):
We just merged the ts-browser util changes. You can now access `window.helpers` to get util from within the browser context.

crypto-ffi/bindings/js/packages/browser/test/e2ei.test.ts
typfel: Done

crypto-ffi/bindings/js/packages/browser/test/e2ei.test.ts
typfel: Done

crypto-ffi/src/e2ei/mod.rs
```diff
@@ -24,3 +33,173 @@ impl From<core_crypto::E2eiConversationState> for E2eiConversationState {
...
+    /// ACME server hostname.
+    pub acme_url: String,
+    /// OIDC identity provider URL.
+    pub idp_url: String,
+    /// Ciphersuite of the acquired credential.
+    pub ciphersuite: FfiCiphersuite,
+    /// User-visible display name.
+    pub display_name: String,
+    /// Wire client id for the device acquiring the credential.
+    pub client_id: Arc<ClientId>,
```
istankovic, line 65: This `Arc` here looks wrong -- is it to satisfy some uniffi constraint?

istankovic:
I think this looks good overall, but I would just like to clarify the situation around that `Arc` holding the client ID.

crypto-ffi/src/e2ei/mod.rs
typfel: Seems so, I get an uniffi error if remove the arc and all ClientIds are wrapped in Arcs in the ffi

```
The trait bound `client_id::ClientId: uniffi::Lift<UniFfiTag>` is not satisfied [E0277]
unsatisfied trait bound
Help: the trait `uniffi::Lift<UniFfiTag>` is not implemented for `client_id::ClientId`
Help: the following other types implement trait `uniffi::Lift<UT>`:
```

typfel: So clientIds are uniffi objects which must always be wrapped in Arcs.

istankovic: To expand on this: the problem is that we want proper types on the client side, i.e. external to Rust. If we were to use custom types and manually lower/lift instances, we would lose type safety on the client side, which is a big downside. So we need to have a proper type on the client side, which means we have to use `Arc`s.

istankovic (approved):
Looks good, thanks!

PR #2079, opened 2026-04-27
test(kotlin): wait for epoch observer callback

SimonThormeyer:
The Kotlin epoch observer wrapper reports events by launching the user callback into the provided CoroutineScope. That means the native epoch change notification can return, and the transaction can complete, before the launched callback has run.

epochObserverEvent_shouldAllowReadingData asserted on the observer's mutable list immediately after updateKeyingMaterial. In CI this occasionally raced with the launched callback and observed zero events even though the epoch change was correctly scheduled.

Use a CompletableDeferred that is completed only after the observer callback has read the conversation epoch and recorded the event. The test now waits for that callback completion before checking the event count and epoch values, preserving the behavior under test without depending on scheduler timing.

# What's new in this PR

----

- The **PR Title**


coriolinus (approved):
This Kotlin is beyond my comfort zone in that language, but it seems plausible.

PR #2080, opened 2026-04-27
chore(ffi): hand-write a `Display` impl for `SecretKey`

coriolinus:
This makes it slightly harder to accidentally reveal the secret value.

This also never does anything with the heap, which keeps it nicely lean / efficient.

Note that this does _not_ use `core_crypto_macros::Debug` because the macro has no good way to forward the `#[sensitive]` attribute, and if nothing is sensitive, then there's no point in using the CC version of the macro.

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/src/core_crypto_context/mls.rs
+++ b/crypto-ffi/src/core_crypto_context/mls.rs
@@ -1,4 +1,4 @@
-use std::{sync::Arc, time::Duration};
+use std::{fmt, sync::Arc, time::Duration};
 
 use core_crypto::{
     Ciphersuite as CryptoCiphersuite, CredentialFindFilters, MlsConversationConfiguration,
@@ -21,7 +21,16 @@ bytes_wrapper!(
     #[uniffi::export(Eq, Hash, Display)]
     SecretKey infallibly wraps core_crypto::mls::conversation::SecretKey; copy_bytes
 );
-impl_display_via_hex!(SecretKey);
+impl fmt::Display for SecretKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let hash = obfuscate::compute_hash(&self.0);
+        let mut buffer = [0; 20];
+        hex::encode_to_slice(hash, &mut buffer).expect("hash has constant length 10 therefore encodes to 20 bytes");
+        // SAFETY: hex crate always produces valid utf-8
+        let data = unsafe { str::from_utf8_unchecked(&buffer) };
+        f.write_str(data)
+    }
+}
 
 bytes_wrapper!(
     /// The raw public key of an external sender.


crypto-ffi/src/core_crypto_context/mls.rs
```diff
@@ -21,7 +21,29 @@ bytes_wrapper!(
...
+            hex::encode_to_slice(&self.0[..EXCERPT_SIZE], &mut head).expect("valid buffer length");
+            hex::encode_to_slice(&self.0[(self.0.len() - EXCERPT_SIZE)..], &mut tail).expect("valid buffer length");
+            // SAFETY: hex crate always produces valid utf8 data
+            let (head, tail) = unsafe {
+                let head = str::from_utf8_unchecked(&head);
+                let tail = str::from_utf8_unchecked(&tail);
+                (head, tail)
+            };
+            write!(f, "[{head}..{tail}]")
+        }
```
istankovic: Eh, can't say I'm a fan of this.
Why not just dump `obfuscate::compute_hash(self.0)`, possibly marked with something?

coriolinus: Mainly because I thought `obfuscate` was not already a dependency at the FFI layer. And writing a little heap-free elider is fun.

You're right though. Update incoming.

coriolinus: https://github.com/wireapp/core-crypto/pull/2080/commits/c8bafd8df1eff098b58690e590ac8f87e7e3bbcd

crypto-ffi/src/core_crypto_context/mls.rs
```diff
@@ -21,7 +21,16 @@ bytes_wrapper!(
...
 );
-impl_display_via_hex!(SecretKey);
+impl fmt::Display for SecretKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let hash = obfuscate::compute_hash(&self.0);
+        let mut buffer = [0; 20];
+        hex::encode_to_slice(hash, &mut buffer).expect("hash has constant length 10 therefore encodes to 20 bytes");
+        // SAFETY: hex crate always produces valid utf-8
+        let data = unsafe { str::from_utf8_unchecked(&buffer) };
+        f.write_str(data)
```
istankovic, line 31: How about
```rust
        let data = str::from_utf8(&buffer).expect("hex crate always produces valid utf-8");
        f.write_str(data)
```
?

As this is not a performance sensitive place, I don't think the extra UTF-8 check is going to hurt
and we should be keeping the number of `unsafe` blocks as low as possible.

coriolinus: Why leave performance on the table when we know for a fact that this will always succeed? I don't think that minimizing unsafe blocks is the best heuristic to apply, especially when safety is trivial to prove.

istankovic: I understand your point, however I just feel that it is better to have fewer assumptions, especially around external crates, rather than more. Generally speaking, the failure mode of a failed assumption with `unsafe` is UB, which is probably the worst that could happen. So I prefer to minimise the set of assumptions, especially when it does not impact performance in any meaningful way, like in this particular case.

That said, I'm not going to block this.

istankovic (approved):
I do think we can do without `unsafe` here, but I'm leaving it up to you.

PR #2088, opened 2026-04-28
fix(conversation): don't eagerly emit epoch change events [WPB-25113]

SimonThormeyer:
# What's new in this PR
Instead, use the new transaction context API to queue them. From that queue, they'll emitted only if the transaction is actually committed.

----

- The **PR Title**

### DIFF
--- a/crypto/src/mls/conversation/conversation_guard/decrypt/buffer_messages.rs
+++ b/crypto/src/mls/conversation/conversation_guard/decrypt/buffer_messages.rs
@@ -226,18 +226,6 @@ mod tests {
                     assert!(msg.app_msg.is_none());
                 }
             }
-            let observed_epochs = observer
-                .observed_epochs()
-                .await
-                .into_iter()
-                .map(|(_conversation_id, epoch)| epoch)
-                .collect::<Vec<_>>();
-            dbg!(&observed_epochs);
-            assert_eq!(
-                observed_epochs.len(),
-                2,
-                "there was 1 buffered commit changing the epoch plus the outer commit changing the epoch"
-            );
 
             assert_eq!(conversation.member_count().await, 4);
             assert!(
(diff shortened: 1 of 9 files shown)


crypto/src/transaction_context/mod.rs
```diff
@@ -51,6 +51,7 @@ enum TransactionContextInner {
         database: Database,
         mls_session: Arc<RwLock<Option<Session<Database>>>>,
         mls_groups: Arc<RwLock<GroupStore<MlsConversation>>>,
+        pending_epoch_changes: Arc<RwLock<Vec<(ConversationId, u64)>>>,
```
coriolinus: This isn't wrong, but a simple `Mutex` might be lighter-weight than a `RwLock` here. I don't imagine that there's going to be a ton of contention for this value.

crypto/src/transaction_context/mod.rs
```diff
@@ -226,16 +240,34 @@ impl TransactionContext {
...
-        let commit_result = keystore
+        let commit_result = database
             .commit_transaction()
             .await
             .map_err(KeystoreError::wrap("commiting transaction"))
             .map_err(Into::into);
 
+        if let Some(session) = mls_session.read_arc().await.clone()
+            && commit_result.is_ok()
+        {
```
coriolinus, line 259: Nice! This is the heart of it, and it's easier than I'd expected.

coriolinus (approved):

crypto/src/mls/conversation/conversation_guard/decrypt/buffer_messages.rs
```diff
@@ -248,6 +236,17 @@ mod tests {
...
+
+            bob.transaction.finish().await.unwrap();
+
+            let observed_epochs = observer
+                .observed_epochs()
+                .await
+                .into_iter()
+                .map(|(_conversation_id, epoch)| epoch)
+                .collect::<Vec<_>>();
+            dbg!(&observed_epochs);
```
istankovic, line 248: This `dbg!` was probably not meant to be part of the final PR...

PR #2100, opened 2026-05-04
docs: CC book in mdbook and deployment [WPB-24894]

coriolinus:
# What's new in this PR

----

- The **PR Title**


cc-book/book.toml
```diff
@@ -0,0 +1,6 @@
+[book]
+title = "CoreCrypto"
```
SimonThormeyer: Not "CoreCrypto Book"?

cc-book/src/concepts/mls/client.md
```diff
@@ -0,0 +1,44 @@
+# The Client
+
+In MLS terminology a _client_ is a single device acting on behalf of a user — each phone, desktop, or web session that participates in MLS groups is a distinct client with its own identity and key material.
+CoreCrypto maps this concept to the `CoreCrypto` object itself.
+One `CoreCrypto` instance, backed by one `Database`, represents one MLS client.
+
+> **Note**
+>
+> The Rust library internally calls this type `Session` to avoid ambiguity — "client" means something different at the Wire application layer (where it refers to a registered device in the Wire backend sense) and at the MLS protocol layer.
+> The name `Session` does not appear in platform bindings; from Swift, Kotlin, and TypeScript, the type is always `CoreCrypto`.
```
SimonThormeyer: Wouldn't it simplify things if we didn't introduce the term `Client` here, and just called it `Session` here, too?

cc-book/src/concepts/mls/client.md
```diff
@@ -0,0 +1,44 @@
...
+## Lifecycle
+
+A `CoreCrypto` instance is typically created once at application startup and kept alive for the life of the process.
+The initialization sequence is:
+
+```typescript
+const db = await Database.open(path, key);       // open the keystore
+const cc = CoreCrypto.new(db);                   // construct CoreCrypto
+await cc.transaction(async (ctx) => {
+    await ctx.mlsInit(clientId, transport);      // initialize the MLS session
```
SimonThormeyer, line 40: E.g., here, we also refer to it as session.

cc-book/src/concepts/mls/client.md
```diff
@@ -35,6 +36,23 @@ await cc.transaction(async (ctx) => {
...
+
+```kotlin
+val db = Database.open(path, key)  // open the keystore
+val cc = CoreCrypto(db)            // construct CoreCrypto
+cc.transaction { ctx ->
+    ctx.mlsInit(clientId, transport)  // initialize the MLS session
+}
+```
+<!-- langtabs-end -->
+
```
SimonThormeyer, line 61: Amazing that we get langtabs!

fewerner (commented):
I like it so far.
I think it would be great if the migration guide wasn't split into separate pages for readers, not sure if there is a preprocessor for achieving this. 

I would also include external links to the language specific APIs in the book. Can we maybe include the API docs as assets?

SimonThormeyer (approved):
Very curious how this is going to look!
Just one comment unrelated to this PR, but would like to discuss it anyway

cc-book/src/concepts/mls/client.md
coriolinus: My thought here was to at least introduce both terms to help people bridge the gap between RFC 9420, the CC internals, and the CC externals. 9420 calls it a client, internals call it a session, externals calls it `CoreCrypto` (after `mls_init`, anyway). I tried to make that explicit in case people go diving.

istankovic: I actually like the way it's phrased right now. And honestly, I've never been a big fan of `Session` and `CoreCrypto`. Hopefully at some point we will have a chance to name the internal type just `Client`. :-)

coriolinus:
> I think it would be great if the migration guide wasn't split into separate pages for readers, not sure if there is a preprocessor for achieving this.

We do have that option. We'd need to use something like [`mdbook-tera`](https://github.com/avitex/mdbook-tera) and then build the book N times to fill in the gaps, but then we could have "CC Book: Typescript Edition" etc. 

I intentionally chose not to do that, and instead just add language-specific sub-chapters to the migration section, on these grounds:

1. I found there to be a lot of complexity in the conditionals; it was hard for me to ensure that this was a complete migration, because I had to check multiple source files each of which _might_ link to other source files, in order to render the complete text.
2. I expect it to be more helpful for readers to have a single version of the CC book to link to.
3. This isn't actually a change; the existing book was also split into language-specific sub-chapters, as seen here:

{04E6C7BD-2875-46AF-A219-15870C067863}

> I would also include external links to the language specific APIs in the book. Can we maybe include the API docs as assets?

The external links to language specific APIs are [here, in the introduction](https://github.com/wireapp/core-crypto/blob/4dc21877c24cf0399d7658fb73dd29693d42d262/cc-book/src/introduction.md?plain=1#L7). I intentionally did not include them as assets because it took over 10 realtime minutes to rebuild all the docs to get the screenshot for the old verison in the last section.

PR #2107, opened 2026-05-05
refactor: externalize `Arc`s and make `CoreCrypto: !Clone` [WPB-25220]

coriolinus:
# What's new in this PR

Note: we'd do something similar for `Database`, externalizing its internal `Arc`s, but that ends up with a problem. For context, each `CoreCrypto` instance + several other places need to store an `Arc<Database>`, not just a `Database`, because we have to copy the database around into a bunch of places. Unfortunately, the [Proteus `PreKeyStore` trait](https://github.com/wireapp/proteus/blob/7a307e819e16ae9046128fa85b24917992b329fd/crates/proteus-traits/src/lib.rs#L80-L88) takes mutable references, which would mean that all those places need to store some kind of `Arc<RwLock<Database>>`, and at that point the cost-benefit analysis kind of stops being worth it. 

----

- The **PR Title**


SimonThormeyer (approved):
Very nice!

PR #2116, opened 2026-05-07
chore: x509 refactor, part 11: The Valley Of Horrors

istankovic:
This comprises several changes:
- PKI env now stores trust anchors, intermediate certs and CRLs to the db
- cert and credential validation is done via the outer PKI env, instead of using the inner directly
- introduction of `AuthenticationService` that is just a wrapper around a PKI env; the purpose
  of this type is linking the MLS parts of the `crypto` crate and PKI env/`e2e-identity`
- time of interest is now computed internally during cert validation
- removal of `PkiEnvironmentProvider`
- misc simplifications

:warning: x509 tests in the `crypto` crate are disabled because they are broken with this PR. The idea is to fix them as part of a later PR.


e2e-identity/src/pki_env/mod.rs
```diff
@@ -161,68 +161,58 @@ impl PkiEnvironment {
...
-                            }
-                            PathValidationStatus::InvalidNotAfterDate => {
-                                // ? Expired credentials are A-OK. They still degrade conversations though.
-                                // return CredentialAuthenticationStatus::Expired;
-                            }
-                            _ => return CredentialAuthenticationStatus::Invalid,
-                        }
-                    } else {
-                        return CredentialAuthenticationStatus::Unknown;
+    pub async fn validate_credential<'a>(
```
coriolinus: I hope that later on in this PR you go through and add docs to all these public functions. The knowledge that passing a basic credential is a programmer error which will panic is critical, and exactly like what we'd want to see documented in inline docs and not just the commit message.

e2e-identity/src/pki_env/mod.rs
```diff
@@ -161,68 +161,58 @@ impl PkiEnvironment {
...
+        &'a mut self,
+        credential: CredentialRef<'a>,
+    ) -> CredentialAuthenticationStatus {
+        let certificates = if let CredentialRef::X509 { certificates } = credential {
+            certificates
+        } else {
+            unreachable!("this function can only be called with an X509 credential");
```
coriolinus: This is not unreachable at all. You might `panic!` here, but `unreachable!` is wrong.

coriolinus: But why not just return unknown status per the previous behavior of the function, here?

e2e-identity/src/pki_env/mod.rs
```diff
@@ -161,68 +161,58 @@ impl PkiEnvironment {
...
+        };
+
+        let pki_env = &mut self.rjt_pki_env;
+        pki_env.refresh_time_of_interest();
+
+        use x509_cert::der::Decode as _;
```
coriolinus: Could we move imports like this to either the top of the function implementation or the top of the file? It's a convention which makes it easier to mentally group them.

e2e-identity/src/pki_env/mod.rs
```diff
@@ -161,68 +161,58 @@ impl PkiEnvironment {
...
+                    _ => return CredentialAuthenticationStatus::Invalid,
                 }
-
-                CredentialAuthenticationStatus::Valid
+            } else {
+                return CredentialAuthenticationStatus::Unknown;
             }
         }
+
+        CredentialAuthenticationStatus::Valid
```
coriolinus: It is possible to inline all this logic into a single match statement, which makes the whole thing shorter and IMO easier to understand:

```rust
match self.rjt_pki_env.lock().await.validate_cert_and_revocation(&cert) {
    Err(RustyX509CheckError::CertValError(CertvalError::PathValidation(
        PathValidationStatus::CertificateRevoked
        | PathValidationStatus::CertificateRevokedEndEntity
        | PathValidationStatus::CertificateRevokedIntermediateCa,
    ))) => {
        // ? Revoked credentials are A-OK. They still degrade conversations though.
        CredentialAuthenticationStatus::Valid
    }
    Err(RustyX509CheckError::CertValError(CertvalError::PathValidation(
        PathValidationStatus::InvalidNotAfterDate,
    ))) => {
        // ? Expired credentials are A-OK. They still degrade conversations though.
        CredentialAuthenticationStatus::Valid
    }
    Err(RustyX509CheckError::CertValError(CertvalError::PathValidation(_))) => {
        CredentialAuthenticationStatus::Invalid
    }
    Err(_) => CredentialAuthenticationStatus::Unknown,
    Ok(_) => CredentialAuthenticationStatus::Valid,
}
```

e2e-identity/src/pki_env/mod.rs
```diff
@@ -131,6 +131,10 @@ impl PkiEnvironment {
         Ok(())
     }
 
+    pub async fn set_pki_environment_provider(&mut self, provider: RjtPkiEnvironment) {
```
coriolinus: `#[cfg(test)]`?

Anyway if this is temporary, please document that inline, not just in the commit message.

crypto/src/mls_provider/mod.rs
```diff
@@ -68,6 +70,36 @@ impl std::ops::DerefMut for EntropySeed {
...
+    }
+}
+
+#[cfg_attr(target_os = "unknown", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_os = "unknown"), async_trait::async_trait)]
+impl openmls_traits::authentication_service::AuthenticationServiceDelegate for AuthenticationService {
+    async fn validate_credential<'a>(&'a self, credential: CredentialRef<'a>) -> CredentialAuthenticationStatus {
+        match credential {
+            // We assume that Basic credentials are always valid
+            CredentialRef::Basic { identity: _ } => CredentialAuthenticationStatus::Valid,
```
coriolinus: If we make this assumption, why not move it into the `PkiEnvironment::validate_credential` implementation and avoid a match statement at this level?

coriolinus: Nit: it's simpler to write `CredentialRef::Basic { .. }` than to manually unbind each of its fields. Likewise for `CredentialRef::X509 { .. }`.

e2e-identity/src/pki_env/mod.rs
istankovic: > This is not unreachable at all. You might panic! here, but unreachable! is wrong.

Right, `panic!` is a much better fit. 👍 

> But why not just return unknown status per the previous behavior of the function, here?

Because ending up in this branch is really a programmer error. Dispatching based on the credential type happens in `crypto` and here one would ideally be able to specify something like a pattern in the function type, e.g. 
```rust
pub async fn validate_credential<'a>(&'a self, credential: CredentialRef::X509<'a>) -> CredentialAuthenticationStatus {
```
but unfortunately Rust does not support that.

e2e-identity/src/pki_env/mod.rs
```diff
@@ -165,6 +165,50 @@ impl PkiEnvironment {
...
+        let (ski, aki) = RjtPkiEnvironment::extract_ski_aki_from_cert(&cert)?;
+        let ski_aki_pair = format!("{ski}:{}", aki.unwrap_or_default());
+        let cert_der = RjtPkiEnvironment::encode_cert_to_der(&cert)?;
+        let intermediate_cert = E2eiIntermediateCert {
+            content: cert_der,
+            ski_aki_pair,
+        };
+
+        self.database.new_transaction().await?;
+        self.database.save(intermediate_cert).await?;
```
coriolinus: So this is where all the hassle is coming from. But what happens if we don't bother adding a second concurrent transaction, but just fail here if no database transaction is already open? Just let `.save` throw the `MutatingOperationWithoutTransaction` and let callers deal with it.

e2e-identity/src/pki_env/mod.rs
```diff
@@ -165,6 +165,50 @@ impl PkiEnvironment {
...
+        // Get CRL distribution points and CRLs
+        let dps: Vec<String> = extract_crl_uris(&cert)?.iter().flatten().cloned().collect();
+        let crls = self.fetch_crls(dps.iter().map(AsRef::as_ref)).await?;
+
+        // Save all CRLs to the database
+        for (distribution_point, crl) in &crls {
+            self.save_crl(distribution_point, crl).await?;
+        }
+
+        self.database.commit_transaction().await?;
```
coriolinus, line 221: Likewise I'd remove this.

e2e-identity/src/pki_env/mod.rs
istankovic: Absolutely. :+1: 
I tried to leave the function intact as much as possible in order to make it easier to review the move.

e2e-identity/src/acquisition/checks.rs
```diff
@@ -46,12 +46,13 @@ pub(crate) async fn verify_cert_chain(
 fn verify_leaf_certificate(
     config: &X509CredentialConfiguration,
     pki_env: &RjtPkiEnvironment,
+    outer_pki_env: &PkiEnvironment,
```
coriolinus, line 46: It feels very weird that we'd need both here. Why not make `PkiEnvironment::rjt_pki_env` `pub(crate)` and then just pass `pki_env: &PkiEnvironment`?

e2e-identity/src/pki_env/mod.rs
```diff
@@ -219,13 +225,16 @@ impl PkiEnvironment {
...
+            .await
+            .add_certificate_source(Box::new(cert_source));
 
         Ok(())
     }
 
-    pub fn validate_cert(&self, cert: &x509_cert::Certificate) -> RustyX509CheckResult<()> {
-        self.rjt_pki_env.validate_cert_and_revocation(cert)
+    pub async fn validate_cert(&self, cert: &x509_cert::Certificate) -> RustyX509CheckResult<()> {
+        self.rjt_pki_env.lock().await.validate_cert_and_revocation(cert)
```
coriolinus, line 246: If you use [`lock_blocking`](https://docs.rs/async-lock/latest/async_lock/struct.Mutex.html#method.lock_blocking) here then this doesn't have to be async at all, which means we can drop 9330edeb9f506984074442d85ffa922b3b26f439 entirely.

crypto/src/mls/conversation/mod.rs
```diff
@@ -224,16 +223,20 @@ pub trait Conversation<'a>: ConversationWithMls<'a> {
...
-            .map(|(_, c)| {
-                c.extract_identity(conversation.ciphersuite(), guard.as_ref().map(|v| &**v))
-                    .map_err(RecursiveError::mls_credential("extracting identity"))
-            })
-            .collect::<Result<Vec<_>, _>>()
-            .map_err(Into::into)
+
+        let mut identities = vec![];
+        for (id, credential) in conversation.members_with_key() {
+            if device_ids.iter().all(|client_id| client_id.borrow() != id) {
```
coriolinus: I'd expect the `any` check to have better short-circuit behavior than this `all` implementation.

e2e-identity/src/pki_env/mod.rs
istankovic: Much nicer, thanks! :+1:

coriolinus (commented):
- commit 27973334fc354576ebf60cd52bc713bf22cb145d has a message which appears entirely unrelated to its contents

e2e-identity/src/pki_env/mod.rs
coriolinus: Likewise c4ee6b115584d93eb42c295073b4efa932c2f28f

coriolinus:
And as mentioned in chat, it's not obvious why https://github.com/wireapp/core-crypto/pull/2116/changes/4d60315c3dd31c48afadd15a6d84f6d7530787cc is necessary. If we can remove that, then we can remove several downstream changes.

e2e-identity/src/pki_env/mod.rs
istankovic: Originally I had used `lock_blocking`. Unfortunately, it's not available on wasm.

crypto/src/mls_provider/mod.rs
istankovic: Because the intent is to have `e2e-identity` be x509-credential-only. Nothing in the `e2e-identity` crate should be responsible for basic credentials.

Long-term, I would like to move basic credentials to their own separate crate and completely decouple credential creation and validation from the `crypto` crate.

I would also like to rename the `e2e-identity` crate to something more appropriate, e.g. `x509-credential`. The only reason I haven't done this yet is that there have been more important issues.

istankovic:
> commit https://github.com/wireapp/core-crypto/commit/27973334fc354576ebf60cd52bc713bf22cb145d has a message which appears entirely unrelated to its contents

I admit it's not quite obvious, but if you look at the types being imported, you will see that the `PkiEnvironment` type no longer comes from the `revocation` module (the inner PKI env) but the `pki_env` module (the outer PKI env). :sweat_smile:

e2e-identity/src/acquisition/checks.rs
istankovic: Heh, fair, but I feel that, given that the major theme of the PR was to remove the equivalent of `rjt_pki_env` on the our PKI env (former `mls_pki_env_provider`), it would be a step back to reintroduce it, even if were just `pub(crate)`. I would rather fix this as part of a later refactoring.

crypto/src/mls_provider/mod.rs
istankovic: Good idea, thanks! Fixed.

e2e-identity/src/pki_env/mod.rs
istankovic: Note that this is removed in `0b26c7d293119deeaab986f111efbed3e288793b`.

e2e-identity/src/pki_env/mod.rs
istankovic: We cannot do that because that would require a client. And our public API should not require a client in order to configure the PKI environment.

e2e-identity/src/pki_env/mod.rs
istankovic: See the other comment, https://github.com/wireapp/core-crypto/pull/2116#discussion_r3209638163.

crypto/src/mls/conversation/mod.rs
istankovic: Why?

istankovic: `any` and `all` can be considered dual, with the inverted condition the short-circuiting behaviour is the same. But you have a point that using `all` here instead of `any` is less natural and harder to read so I'm going to switch to `any`.

e2e-identity/src/pki_env/mod.rs
```diff
@@ -161,111 +161,118 @@ impl PkiEnvironment {
...
+    }
 
+    pub async fn validate_credential<'a>(&'a self, credential: CredentialRef<'a>) -> CredentialAuthenticationStatus {
+        let certificates = if let CredentialRef::X509 { certificates } = credential {
+            certificates
+        } else {
+            panic!("this function can only be called with an X509 credential");
+        };
```
coriolinus, line 257: I still think that returning `CredentialAuthenticationStatus::Unknown` would be more appropriate than panicing here, even if you're adamant that this crate won't handle basic credentials. But I'm willing to accept it.

What I won't accept is a panicing public function with zero documentation about what causes it to panic. That's just asking for trouble.

coriolinus: Also it's slightly simpler like this:

```suggestion
    pub async fn validate_credential<'a>(&'a self, credential: CredentialRef<'a>) -> CredentialAuthenticationStatus {
        let CredentialRef::X509 { certificates } = credential else {
            panic!("this function can only be called with an X509 credential");
        };
```

e2e-identity/src/pki_env/mod.rs
coriolinus: Normally I'd suggest a structural change, define the `CredentialRef` enum such that each branch contains an independent struct which can be named. Unfortunately this one comes directly from OpenMLS. Oh well.

e2e-identity/src/pki_env/mod.rs
coriolinus: It does not require a client, it just requires that the public face of `PkiEnv` defines its own way to open and resolve a database transaction. Honestly I feel that that's a better approach than defining two different flavors of transaction.

istankovic: > It does not require a client, it just requires that the public face of PkiEnv defines its own way to open and resolve a database transaction.

But why should clients care about that? The transaction is something that happens internally, I see no reason to expose that to clients.

e2e-identity/src/pki_env/mod.rs
istankovic: Oh that is definitely nicer, thanks! :+1: 

Regarding 

> What I won't accept is a panicing public function with zero documentation about what causes it to panic. That's just asking for trouble.

note that I haven't forgotten about documentation, I  do plan to add docs after I rebase on top of Felix' PR and make sure everything else works fine. It's coming, but I left it for the end of the PR, which is also why I didn't resolve the one comment at the top where you asked for docs.

e2e-identity/src/pki_env/mod.rs
coriolinus: We agreed to use this but with `database.try_new_immediate_transaction().await?`.

e2e-identity/src/pki_env/mod.rs
istankovic: Done in 154f0d305b33529590b42dc7302130e0b6855833.

e2e-identity/src/pki_env/mod.rs
istankovic: We're now using the nicer pattern. Documentation added in 154f0d305b33529590b42dc7302130e0b6855833.

e2e-identity/src/pki_env/mod.rs
```diff
@@ -153,17 +153,6 @@ impl PkiEnvironment {
         &self.database
     }
 
-    pub async fn trust_anchor(&self) -> Result<Certificate> {
-        let trust_anchor = self
-            .database
-            .get_unique::<E2eiAcmeCA>()
```
coriolinus, line 156: Do we even need a unique `E2eiAcmeCA` entity anymore? That can be a follow-up PR, but if we're not using it here, then what's the point of storing it?

e2e-identity/src/pki_env/mod.rs
coriolinus, line 82: If trust anchors derive from the `rjt_pki_env` value now instead of the database, then what is this loading?

e2e-identity/src/pki_env/mod.rs
```diff
@@ -97,175 +103,182 @@ async fn restore_pki_env(data_provider: &impl FetchFromDatabase) -> Result<RjtPk
...
+        // Save cert's DER representation to the database
+        let cert_data = E2eiAcmeCA {
+            content: cert.to_der()?,
+        };
 
-    pub async fn add_trust_anchor(&mut self, name: &str, cert: Certificate) -> Result<()> {
-        let mut guard = self.mls_pki_env_provider.0.write().await;
-        let pki_env = guard.as_mut().expect("inner PKI environment must be set");
+        self.database.try_new_immediate_transaction().await?;
+        self.database.save(cert_data).await?;
```
coriolinus, line 176: `E2eiAcmeCA` implements `UniqueEntity` via `UniqueEntityImplementationHelper`. This `save` call is going to silently replace any earlier trust anchors:

https://github.com/wireapp/core-crypto/blob/eba0fdd61d9dd6f3399de5040033fd2f301af047/keystore/src/traits/unique_entity.rs#L297-L304

coriolinus (changes requested):
I really want to get this PR into place, but the trust anchor stuff is underbaked. Could you just split the last commits into the start of part 12 so I can approve this?

e2e-identity/src/pki_env/mod.rs
istankovic: We're still storing only one trust anchor, even though the API supports multiple ones. Note the `restore_pki_env(data_provider`, which is used to load that one from the db at PKI env creation.

For this to be completely finished, we'd need a db migration and said we're going to postpone additional migrations until after db unification (this is also why WPB-24480 was postponed).

e2e-identity/src/pki_env/mod.rs
istankovic: It's loading the trust anchor from the db, during PKI env construction.

e2e-identity/src/pki_env/mod.rs
istankovic: Indeed! I agree, it's underbaked, but I'm not sure how to proceed, see https://github.com/wireapp/core-crypto/pull/2116#discussion_r3232445292.

coriolinus (approved):

e2e-identity/src/pki_env/mod.rs
istankovic: Decided to document the limitation and remove it post-CC10.

PR #2121, opened 2026-05-11
feat: add separate pki transaction

fewerner:
# What's new in this PR

This PR adds a separate pki transaction to the database to allow crypto and e2ei to use independent transactions.

----

- The **PR Title**


keystore/src/connection/mod.rs
```diff
@@ -69,11 +69,13 @@ pub trait DatabaseConnection<'a>: DatabaseConnectionRequirements {
...
 
+const INDEPENDENT_TRANSACTIONS_COUNT: usize = 2;
+
 #[derive(Debug, Clone)]
 pub struct Database {
     pub(crate) conn: Arc<Mutex<Option<KeystoreDatabaseConnection>>>,
-    pub(crate) transaction: Arc<Mutex<Option<KeystoreTransaction>>>,
-    transaction_semaphore: Arc<Semaphore>,
+    pub(crate) transaction: [Arc<Mutex<Option<KeystoreTransaction>>>; INDEPENDENT_TRANSACTIONS_COUNT],
+    transaction_semaphore: [Arc<Semaphore>; INDEPENDENT_TRANSACTIONS_COUNT],
```
SimonThormeyer: I'd name these fields `transactions` and `transaction_semaphores` (plural).

keystore/src/connection/mod.rs
SimonThormeyer, line 222: Maybe I'm missing something, but couldn't this whole function just be much simpler, e.g.,
```rust
async fn take(&self) -> CryptoKeystoreResult<KeystoreDatabaseConnection> {
        for semaphore in &self.transaction_semaphore {
            semaphore.acquire_arc().await;
        }

        let mut guard = self.conn.lock().await;
        guard.take().ok_or(CryptoKeystoreError::Closed)
    }
```
?

fewerner: Wouldn't in your example the semaphores be dropped before taking the connection, when leaving the for loop scope?

keystore/src/connection/mod.rs
```diff
@@ -458,11 +458,14 @@ impl FetchFromDatabase for Database {
         E: Entity<ConnectionType = KeystoreDatabaseConnection> + Clone + Send + Sync,
     {
         // If a transaction is in progress...
-        if let Some(transaction) = self.transaction[SLOT].lock().await.as_ref()
-            //... and it has information about this entity, ...
-            && let Some(cached_record) = transaction.get(id).await
-        {
-            return Ok(cached_record.map(Arc::unwrap_or_clone));
+        for slot in 0..self.transaction.len() {
+            let guard = self.transaction[slot].lock().await;
```
SimonThormeyer: Instead of indexing into the array, you can do

```
for transaction in &self.transaction {
...
```

keystore/src/connection/mod.rs
```diff
@@ -476,14 +479,16 @@ impl FetchFromDatabase for Database {
         E::PrimaryKey: Borrow<E::BorrowedPrimaryKey>,
         for<'a> &'a E::BorrowedPrimaryKey: KeyType,
     {
-        // If a transaction is in progress...
-        if let Some(transaction) = self.transaction.lock().await.as_ref()
+        for slot in 0..self.transaction.len() {
+            let guard = self.transaction[slot].lock().await;
```
SimonThormeyer: Same here

keystore/src/connection/mod.rs
```diff
@@ -493,7 +498,16 @@ impl FetchFromDatabase for Database {
     where
         E: Entity<ConnectionType = KeystoreDatabaseConnection> + Clone + Send + Sync,
     {
-        if self.transaction.lock().await.is_some() {
+        let mut has_any_transaction = false;
+
+        for slot in &self.transaction {
```
SimonThormeyer: nit: I would expect a variable named `slot` to be a number of 0 or 1 in this context, but it's not, I guess? Maybe `transaction` would be the better name here?

keystore/src/connection/mod.rs
```diff
@@ -511,11 +525,21 @@ impl FetchFromDatabase for Database {
         let mut conn = self.conn().await?;
         let persisted_records = E::load_all(&mut conn).await?;
 
-        let transaction_guard = self.transaction.lock().await;
-        let Some(transaction) = transaction_guard.as_ref() else {
-            return Ok(persisted_records);
-        };
-        transaction.find_all(persisted_records).await
+        for slot in 0..self.transaction.len() {
+            let guard = self.transaction[slot].lock().await;
```
SimonThormeyer: No need to index into the array here, either

keystore/src/connection/mod.rs
```diff
@@ -526,11 +550,20 @@ impl FetchFromDatabase for Database {
...
         let persisted_records = E::find_all_matching(&mut conn, search_key).await?;
 
-        let transaction_guard = self.transaction.lock().await;
-        let Some(transaction) = transaction_guard.as_ref() else {
-            return Ok(persisted_records);
-        };
-
-        transaction.search(persisted_records, search_key).await
+        for slot in 0..self.transaction.len() {
+            let guard = self.transaction[slot].lock().await;
```
SimonThormeyer: No need to index into the array here, either

keystore/src/connection/mod.rs
```diff
@@ -526,11 +550,20 @@ impl FetchFromDatabase for Database {
...
+
+            if let Some(transaction) = guard.as_ref() {
+                let cached_records = transaction.search_in_cache::<E, SearchKey>(search_key).await;
+                let merged_records = transaction
+                    .merge_records::<E>(
+                        cached_records.iter().map(Arc::as_ref).map(Cow::Borrowed),
+                        persisted_records.into_iter().map(Cow::Owned),
+                    )
+                    .await;
+                return Ok(merged_records);
```
SimonThormeyer, line 570: Maybe a comment would be helpful why we can, based on our assumptions, do the early return here (and in some other places above, too)?

keystore/src/mls.rs
```diff
@@ -171,7 +175,7 @@ impl CryptoKeystoreMls for crate::Database {
     }
 
     async fn mls_group_delete(&self, group_id: impl AsRef<[u8]> + Send) -> CryptoKeystoreResult<()> {
-        self.remove_borrowed::<PersistedMlsGroup>(group_id.as_ref()).await?;
+        <Self as CryptoTransaction>::remove_borrowed::<PersistedMlsGroup>(self, group_id.as_ref()).await?;
```
SimonThormeyer: Nit: It's just a matter of taste, but have you considered using another syntax that avoids that `self` becomes a regular function argument (applies to all changes in this commit)?

keystore/src/connection/mod.rs
SimonThormeyer: Yes, do we really need it for the entire scope? If so, the previous version is fine, though I'm wondering if we can do without heap allocation

keystore/src/connection/mod.rs
```diff
@@ -384,14 +479,16 @@ impl FetchFromDatabase for Database {
         E::PrimaryKey: Borrow<E::BorrowedPrimaryKey>,
         for<'a> &'a E::BorrowedPrimaryKey: KeyType,
     {
-        // If a transaction is in progress...
-        if let Some(transaction) = self.transaction.lock().await.as_ref()
```
SimonThormeyer, line 388: The removed comment is still relevant (even though it needs an update), right?

keystore/src/connection/mod.rs
fewerner: We definetly want to hold the semaphores when taking the connection out of the option, otherwise we could close the database while a transaction is ongoing.

We could avoid heap allocation by avoiding the loop and explicitly aquire each arc. But then we need to remember changing this in case the transaction count changes. Not sure if it's worth it.

SimonThormeyer: I'll leave this up to you.

keystore/src/connection/mod.rs
fewerner: I added a general comment to the transactions field as I would copy the same argument into every function otherwise. Also the commit message for this change explains why this is ok to do.

keystore/src/connection/mod.rs
```diff
@@ -69,11 +69,17 @@ pub trait DatabaseConnection<'a>: DatabaseConnectionRequirements {
...
 }
 
+const INDEPENDENT_TRANSACTIONS_COUNT: usize = 2;
+
 #[derive(Debug, Clone)]
 pub struct Database {
     pub(crate) conn: Arc<Mutex<Option<KeystoreDatabaseConnection>>>,
-    pub(crate) transaction: Arc<Mutex<Option<KeystoreTransaction>>>,
-    transaction_semaphore: Arc<Semaphore>,
+    // We assume that these transactions never hold the same entitx types.
```
SimonThormeyer: typo: `entitx` -> `entity`

SimonThormeyer (approved):
This looks good, pending green CI and the typo fix

coriolinus:
The key question I would really like an answer to: is this really simpler / more maintainable than just adding a transaction API to the PKI environment itself?

istankovic:
> The key question I would really like an answer to: is this really simpler / more maintainable than just adding a transaction API to the PKI environment itself?

Please see https://github.com/wireapp/core-crypto/pull/2116#discussion_r3219588168.

coriolinus:
> > The key question I would really like an answer to: is this really simpler / more maintainable than just adding a transaction API to the PKI environment itself?
> 
> Please see [#2116 (comment)](https://github.com/wireapp/core-crypto/pull/2116#discussion_r3219588168).
>
>>  why should clients care about that? 

It's symmetric: our clients pass a `Database` instance to create a PKI env, and they pass one to create a `CC` instasnce. For mutating operations on CC they have to construct a transaction. For mutating operations on the PKI env they have to construct a transaction. There can be exactly one concurrent transaction on a database at a time, and opening a transaction will block until the semaphore is available. The mental model is very simple, and the underlying code to implement this only a bit more complex.

Contrast this to: in CC you have to create a transaction, but for a PKI environment that is silently handled for you. And normally this could cause a deadlock but we've gone to extreme lengths (including killing the ability to rm all transaction-related code) to hide this from you the user. 

Honestly it's the fact that this PR seems to prevent us from ever being able to clear out the transaction-related code that's the heart of my discontent with the approach. That's a fairly major drawback that I want to be sure is fully justified before we merge this.

fewerner:
Closing this as we went with a much simpler solution in #2134

PR #2123, opened 2026-05-11
ci: run PR benchmark jobs on cloud runners

typfel:
# What's new in this PR

run PR benchmark jobs on cloud runners

----

- The **PR Title**

### DIFF
--- a/.github/workflows/benchmark-jvm.yml
+++ b/.github/workflows/benchmark-jvm.yml
@@ -21,6 +21,11 @@ on:
         required: false
         type: boolean
         default: true
+      runs_on:
+        description: JSON string or array to use for the benchmark runner
+        required: false
+        type: string
+        default: '"ubuntu-latest"'
   workflow_dispatch:
     inputs:
       benchmark_message_counts:
@@ -44,6 +49,11 @@ on:
         required: false
         type: boolean
         default: true
+      runs_on:
+        description: JSON string or array to use for the benchmark runner
+        required: false
+        type: string
+        default: '"ubuntu-latest"'
 
 env:
   RELEASE: 1
@@ -56,8 +66,7 @@ permissions:
 
 jobs:
   benchmark:
-    name: Performance regression check
-    runs-on: [self-hosted, benchmark]
+    runs-on: ${{ fromJSON(inputs.runs_on) }}
     env:
       BENCHMARK_MESSAGE_COUNTS: ${{ inputs.benchmark_message_counts }}
       BENCHMARK_MESSAGE_SIZES: ${{ inputs.benchmark_message_sizes }}
--- a/.github/workflows/benchmark-ts-browser.yml
+++ b/.github/workflows/benchmark-ts-browser.yml
@@ -21,6 +21,11 @@ on:
         required: false
         type: boolean
         default: true
+      runs_on:
+        description: JSON string or array to use for the benchmark runner
+        required: false
+        type: string
+        default: '"ubuntu-latest"'
   workflow_dispatch:
     inputs:
       benchmark_message_counts:
@@ -44,6 +49,11 @@ on:
         required: false
         type: boolean
         default: true
+      runs_on:
+        description: JSON string or array to use for the benchmark runner
+        required: false
+        type: string
+        default: '"ubuntu-latest"'
 
 env:
   RELEASE: 1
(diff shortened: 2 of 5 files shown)


.github/workflows/benchmark-jvm.yml
```diff
@@ -21,6 +21,11 @@ on:
         required: false
         type: boolean
         default: true
+      runs_on:
+        description: JSON string or array to use for the benchmark runner
+        required: false
```
coriolinus, line 26: If you set `required: true`, then wouldn't it always use the default if the user doesn't manually set it? Then we could avoid the `${{ fromJSON(inputs.runs_on || '["self-hosted", "benchmark"]') }}` syntax later one.

.github/workflows/benchmark-jvm.yml
```diff
@@ -57,7 +62,7 @@ permissions:
 jobs:
   benchmark:
     name: Performance regression check
-    runs-on: [self-hosted, benchmark]
+    runs-on: ${{ fromJSON(inputs.runs_on || '["self-hosted", "benchmark"]') }}
```
coriolinus: I don't love this syntax; it took a minute to figure out. (In part because I'm not familiar with using an array in `runs-on`.)

[docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idruns-on) say that using an array matches only machines for which all tags apply. Do our self-hosted runners have the `benchmark` tag?

.github/workflows/pipeline.yml
```diff
@@ -253,9 +255,10 @@ jobs:
 
   benchmark-ts-native:
     needs:
-      - bindings-ts-native-darwin
+      - bindings-ts-native-linux
```
istankovic, line 256: Nit: to me it looks confusing to name these jobs `benchmark-ts-browser` and `benchmark-ts-native`. It gives the impression that we're really benchmarking for real, with the intention to measure.

How about using the `check-` pattern, e.g.
```
check-benchmarks-ts-browser
check-benchmarks-ts-native
```
?

.github/workflows/benchmark-ts-browser.yml
```diff
@@ -21,6 +21,11 @@ on:
         required: false
         type: boolean
         default: true
+      runs_on:
+        description: JSON string or array to use for the benchmark runner
+        required: false
```
coriolinus, line 26: Same comments from the other file also apply here.

.github/workflows/pipeline.yml
istankovic: Similarly for `benchmark-jvm`.

.github/workflows/benchmark-ts-native.yml
```diff
@@ -21,6 +21,11 @@ on:
         required: false
         type: boolean
         default: true
+      runs_on:
+        description: JSON string or array to use for the benchmark runner
+        required: false
```
coriolinus, line 26: Same comments from the other file also apply here.

coriolinus (approved):
I'm approving because farming out the PR's benchmark jobs to GH cloud machines will be a huge immediate DX win. But there are a few things that might be improved.

.github/workflows/pipeline.yml
typfel: Yeah fair point I'll rename them

.github/workflows/benchmark-ts-native.yml
```diff
@@ -57,7 +62,7 @@ permissions:
 jobs:
   benchmark:
     name: Performance regression check
```
istankovic: This name will appear also in the pipeline, right? I think that's confusing, and I wonder if we could rename this to something like "Check benchmarks" in the pipeline.

.github/workflows/pipeline.yml
typfel: Done

.github/workflows/benchmark-jvm.yml
typfel: yes one our self host runners have the benchmark tag so that we always run it on the same machine

.github/workflows/benchmark-ts-native.yml
typfel: That doesn't work because this workflow is flow is also used for the real benchmarks.

.github/workflows/benchmark-jvm.yml
typfel: The this weird syntax to handle manual invocations of the workflow, If we don't care about them we can remove it.

typfel: Pushed a change to avoid this syntax, we repeat the runs-on a few times but it's less confusing now I think.

.github/workflows/benchmark-ts-native.yml
typfel: I mean can rename them of course but it needs to work for benchmark.yml and pipeline.yml

istankovic: Is it possible to specify the name at the usage site, i.e. in `pipeline.yml`?

PR #2134, opened 2026-05-12
feat(keystore): add `Database::try_new_transaction`

coriolinus:
We have a use case for the PKI environment to get a new transaction without waiting: it's a programming error if they try to mutate the PKI environment during a CC transaction, because the PKI environment needs to open its own transaction.

Without this API that error produces a deadlock. With this API the PKI environment can just return an error.

# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/keystore/src/connection/mod.rs
+++ b/keystore/src/connection/mod.rs
@@ -264,6 +264,21 @@ impl Database {
         Ok(())
     }
 
+    /// Start a new transaction if no other transaction is currently in progress.
+    ///
+    /// If a transaction is currently in progress, this will produce a `TransactionInProgress` error.
+    pub async fn try_new_immediate_transaction(&self) -> CryptoKeystoreResult<()> {
+        let semaphore =
+            self.transaction_semaphore
+                .try_acquire_arc()
+                .ok_or_else(|| CryptoKeystoreError::TransactionInProgress {
+                    attempted_operation: "create an immediate transaction".into(),
+                })?;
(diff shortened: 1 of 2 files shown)


istankovic (approved):

keystore/src/connection/mod.rs
```diff
@@ -264,6 +264,19 @@ impl Database {
         Ok(())
     }
 
+    /// Start a new transaction if no other transaction is currently in progress.
+    ///
+    /// If a transaction is currently in progress, this will produce a `TransactionInProgress` error.
+    pub async fn try_new_transaction(&self) -> CryptoKeystoreResult<()> {
```
SimonThormeyer: I don't dislike this name, but would prefer naming that makes it clearer how this function distinguishes from `new_transaction()`.

We could rename the other function to `new_queued_transaction()`, or this one to `try_new_exclusive_transaction()` or `new_exclusive_transaction()` or `try_immediate_transaction()`?

istankovic: I actually like this name, it's in line with the naming used in other locking types. Even with "queued" transactions, one could say you have exclusivity -- every transaction runs has exclusive right to execute during its own period.

coriolinus: Renamed to `try_new_immediate_transaction` in bb48718c7.

SimonThormeyer (approved):
Nice, thanks! Looks good pending a reword of the commit message and green CI

PR #2135, opened 2026-05-12
chore: replace `find` with `fd`

fewerner:
# What's new in this PR

This PR replaces `find` with `fd` for better performance and simpler usage.

----

- The **PR Title**


make/docs.mk
```diff
@@ -19,8 +19,8 @@ $(STAMPS)/docs-rust-wasm: $(RUST_SOURCES)
 docs-rust-wasm: $(STAMPS)/docs-rust-wasm ## Generate Rust docs for wasm32-unknown-unknown
 
 # Kotlin docs
-KOTLIN_SOURCES := $(shell find crypto-ffi/bindings/shared/src/commonMain/kotlin \
-	                           -type f -name '*.kt' 2>/dev/null | LC_ALL=C sort)
```
coriolinus, line 23: I don't think `fd` sorts by default; we might need the explicit sort command still.

coriolinus (commented):
There was a reason we were sorting the sources. I don't remember what that reason was. Do we know if it still applies? I'm reluctant to approve before sorting this out.

SimonThormeyer:
> There was a reason we were sorting the sources. I don't remember what that reason was. Do we know if it still applies? I'm reluctant to approve before sorting this out.

I think sorting makes the hashes more robust; we might have had problems where we were getting different hashes based on the environment (or just randomly, I don't remember).

istankovic:
> I think sorting makes the hashes more robust; we might have had problems where we were getting different hashes based on the environment (or just randomly, I don't remember).

Aye, sorting is critical to get the hashes right and to make the whole CI optimization based on artifacts work. Note that also the locale part (`LC_ALL`) is important.

make/docs.mk
fewerner: I tested it and the output of fd was the same as find and sort.

coriolinus (approved):

make/docs.mk
istankovic: Can you guarantee the deterministic sort order, though? The man page for `fd` says this:
```
     -l, --list-details
            Use  a  detailed listing format like 'ls -l'. This is basically an alias for '--exec-batch ls -l' with some additional 'ls' options. This can be used to see more metadata, to
            show symlink targets and to achieve a deterministic sort order.
```

which would imply that without this option the sort order is not deterministic.

PR #2149, opened 2026-05-18
chore: ensure certificate key and algo match the signing keypair

istankovic:


e2e-identity/src/acquisition/checks.rs
```diff
@@ -44,11 +45,22 @@ async fn verify_leaf_certificate(
...
     pki_env: &RjtPkiEnvironment,
     outer_pki_env: &PkiEnvironment,
+    sign_kp: &Pem,
     cert: &Certificate,
 ) -> Result<(), CertificateError> {
     pki_env.validate_cert(cert)?;
 
-    // TODO: verify that cert is signed by enrollment.sign_kp
+    // Make sure that the public key in the certificate matches the one from the signing keypair.
+    let sign_kp_bytes = crate::utils::public_key_bytes(config.sign_alg, sign_kp).expect("sign_kp must be valid PEM");
```
SimonThormeyer, line 64: What does "must" mean here? Can we be sure that this won't panic at runtime? If so, why?

SimonThormeyer (commented):
One question I'd like to clarify before approving

e2e-identity/src/acquisition/checks.rs
istankovic: Pushed a comment to clarify the situation regarding that `expect`. In short, if it fails, that means the implementation is broken.

SimonThormeyer (approved):

coriolinus (approved):
Thanks for the comment explaining the `expect`, I found it helpful!

PR #2156, opened 2026-05-20
docs: write up the x509 acquisition flow in the book

coriolinus:
# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/cc-book/src/cc10/migration-guide.md
+++ b/cc-book/src/cc10/migration-guide.md
@@ -59,6 +59,61 @@ the sub-pages:
    public-key form, or `ExternalSender.parse()` to try both in turn. Parse errors are reported at parse time rather than
    during conversation creation. Call `externalSender.serialize()` to recover the raw bytes when needed.
 
+## X509 Credential Acquisition
+
+The enrollment API that previously drove the ACME and OIDC exchanges step-by-step from the client has been **removed**.
+The whole ACME / DPoP / OIDC sequence is now hidden behind a single object, `X509CredentialAcquisition`, which the
+client constructs once and then calls `finalize()` on to obtain a `Credential`. CoreCrypto reaches back into the client
+only via well-defined hook points; the client no longer threads nonces, account responses, order requests, or challenge
+payloads through its own code.
+
+Where you previously called `e2eiNewEnrollment`, `e2eiNewActivationEnrollment`, `e2eiNewRotateEnrollment`,
+`directoryResponse`, `newAccountRequest`/`Response`, `newOrderRequest`/`Response`, `newAuthzRequest`/`Response`,
+`createDpopToken`, `newDpopChallengeRequest`/`Response`, `newOidcChallengeRequest`/`Response`,
+`checkOrderRequest`/`Response`, `finalizeRequest`/`Response`, or `certificateRequest` — delete all of it. The new flow
+replaces every one of those calls.
+
+### Acquiring an X509 credential
+
+1. **Implement `PkiEnvironmentHooks`**. CoreCrypto will call these hooks during acquisition:
+
+   - `httpRequest` — perform HTTP requests against the ACME server, CRL distributors, and similar.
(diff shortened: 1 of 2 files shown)


cc-book/src/cc10/migration-guide.md
```diff
@@ -59,6 +59,61 @@ the sub-pages:
...
+
+Where you previously called `e2eiNewEnrollment`, `e2eiNewActivationEnrollment`, `e2eiNewRotateEnrollment`,
+`directoryResponse`, `newAccountRequest`/`Response`, `newOrderRequest`/`Response`, `newAuthzRequest`/`Response`,
+`createDpopToken`, `newDpopChallengeRequest`/`Response`, `newOidcChallengeRequest`/`Response`,
+`checkOrderRequest`/`Response`, `finalizeRequest`/`Response`, or `certificateRequest` — delete all of it. The new flow
+replaces every one of those calls.
+
+### Acquiring an X509 credential
+
+1. **Implement `PkiEnvironmentHooks`** if you have not already. CoreCrypto will call these hooks during acquisition:
```
istankovic: Nit: `if you have not already` is not necessary because the hooks are not there in prior releases.
Maybe it would be better to mention hooks in a separate migration section, where we highlight the fact that they only need one hook impl.

cc-book/src/cc10/migration-guide.md
```diff
@@ -59,6 +59,61 @@ the sub-pages:
...
+
+   - `httpRequest` — perform HTTP requests against the ACME server, CRL distributors, and similar.
+   - `authenticate` — drive the IdP authorization code flow with PKCE and return the resulting ID token.
+   - `getBackendNonce` — obtain a nonce from the Wire backend.
+   - `fetchBackendAccessToken` — exchange the DPoP token for a backend access token.
+
+1. **Create a `PkiEnvironment`** with `PkiEnvironment.new(hooks, database)` (TypeScript) or
+   `createPkiEnvironment(hooks, database)` (Swift/Kotlin). The `Database` can be the same used by the `CoreCrypto`
+   instance or distinct; they use unrelated tables.
```
istankovic: Should we maybe recommend that clients use the same db? I'm primarily thinking, if they use another db, they will need a key and then the question becomes how that key is managed, is it the same as the "main" db key etc. Might be worth to provide some light guidance.

cc-book/src/cc10/migration-guide.md
```diff
@@ -59,6 +59,61 @@ the sub-pages:
...
+   - `displayName`, `clientId`, `handle`, `domain`, optional `team`
+   - `validityPeriodSecs`
+
+1. **Construct the acquisition**: `X509CredentialAcquisition.new(pkiEnvironment, configuration)`.
+
+1. **Call `await acquisition.finalize()`**. This drives the DPoP and OIDC challenges to completion, calling your hooks
+   as needed, and returns the acquired `Credential`. The acquisition can only be finalized once; a second call throws.
+
+1. **Attach the credential** to the client with `transactionContext.addCredential(credential)`, exactly as for a basic
+   credential. This persists it to the keystore and returns a `CredentialRef`.
```
istankovic: Nit: not sure how we call it elsewhere in the book/docs, but since the type is called `Database`, I'd probably say "persists it to the client's database" instead of "persists it to the keystore".

cc-book/src/cc10/migration-guide.md
```diff
@@ -59,6 +59,61 @@ the sub-pages:
...
+
+### Pausing across the IdP redirect
+
+The IdP authentication flow typically requires an external redirect, after which the app may have been suspended,
+backgrounded, or restarted. To support this, the `authenticate` hook receives an `acquisitionSnapshot: bytes` parameter
+capturing the acquisition state at the point the DPoP challenge has completed and OIDC is about to begin. Persist these
+bytes before launching the IdP flow.
```
istankovic: I would add a note/warning here that the persisted bytes must be stored in encrypted storage.

istankovic (approved):
A few comments, but this looks very good, thanks!

PR #2161, opened 2026-05-21
refactor: d param from session [WPB-25784]

coriolinus:
# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/crypto/src/immutable_database.rs
+++ b/crypto/src/immutable_database.rs
@@ -1,0 +1,67 @@
+use std::borrow::Borrow;
+
+use async_trait::async_trait;
+use core_crypto_keystore::{
+    CryptoKeystoreResult, Database,
+    connection::KeystoreDatabaseConnection,
+    traits::{BorrowPrimaryKey, Entity, EntityGetBorrowed, FetchFromDatabase, KeyType, SearchableEntity},
+};
+
+/// This database only exposes immutable operations.
+///
+/// This makes it easier on a type level to get things right:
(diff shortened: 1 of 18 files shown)


crypto/src/immutable_database.rs
```diff
@@ -0,0 +1,68 @@
...
+#[derive(Debug, Clone, derive_more::From)]
+pub struct ImmutableDatabase(Database);
+
+impl ImmutableDatabase {
+    /// Get access to the version of this database which exposes
+    /// mutating operations.
+    pub(crate) fn mutable(&self) -> &Database {
+        &self.0
+    }
+}
```
SimonThormeyer: It would be ideal if `Session` and `CoreCrypto` had different database types: `CoreCrypto` could have one which implements `mutable(&self) -> &Database`, and `Session` one which doesn't. This way we would entirely preserve the previous guarantees, but require an additional new type.
What do you think?

coriolinus: To be honest I'm not sure the additional work is worth it in this case. It's simple enough to ensure that we never call `.mutable()` within the `session` directory tree. We should trust ourselves to do the right thing with this type IMO.

SimonThormeyer: I do think it is worth it, given how little additional complexity/work this is with just a single newtype.

The introduction of the generic parameter was motivated by us occasionally forgetting that mutable interfaces don't belong onto the session, and I don't see why we (or any potential future contributor) cannot forget that we shouldn't call `.mutable()` in the `session` module.

SimonThormeyer (approved):

PR #2162, opened 2026-05-21
test(kotlin): simplify kotlin test helpers

fewerner:
# What's new in this PR

This PR simplifies kotlin test helpers and aligns them to what we do in ts tests.

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/bindings/shared/src/commonTest/kotlin/com/wire/crypto/testutils/TestUtils.kt
+++ b/crypto-ffi/bindings/shared/src/commonTest/kotlin/com/wire/crypto/testutils/TestUtils.kt
@@ -183,22 +183,6 @@ fun randomIdentifier(n: Int = 12): String {
         .joinToString("")
 }
 
-
-/** Shorthand for creating a conversation with defaults */
-suspend fun CoreCryptoContext.createConversationShort(
-    id: ConversationId
-) {
-    val credentials = findCredentials(
-        clientId = null,
-        publicKey = null,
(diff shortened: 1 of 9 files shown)


crypto-ffi/bindings/shared/src/commonTest/kotlin/com/wire/crypto/testutils/TestUtils.kt
```diff
@@ -40,6 +40,16 @@ interface MockDeliveryService : MlsTransport {
...
+    companion object {
+        @Volatile
+        private var instance: MockMlsTransportSuccessProvider? = null
+
+        fun getInstance() =
+            instance ?: synchronized(this) {
+                instance ?: MockMlsTransportSuccessProvider().also { instance = it }
+            }
+    }
+
```
SimonThormeyer, line 53: Is this a singleton? If yes, this could potentially get us into trouble when running tests in parallel, right? If not, could you please write a comment explaining what this is for non-kotlin engineers?

SimonThormeyer (approved):
Just one question I'd like to clarify, but nice in general!

crypto-ffi/bindings/shared/src/commonTest/kotlin/com/wire/crypto/testutils/TestUtils.kt
fewerner: Indeed, it is. It can be a problem if the processes are not isolated. I decided to go for this to easily access them from benchmarks as well, where I couldn't use inheritence on the class.

PR #2164, opened 2026-05-22
feat: expose the public key on the CredentialRef

typfel:
# What's new in this PR

Expose the public key on the CredentialRef

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/src/core_crypto/credential.rs
+++ b/crypto-ffi/src/core_crypto/credential.rs
@@ -1,0 +1,11 @@
+use std::sync::Arc;
+
+use crate::{CoreCryptoFfi, CoreCryptoResult, CredentialRef};
+
+#[uniffi::export]
+impl CoreCryptoFfi {
+    /// Get the public key associated with this credential
+    pub async fn public_key(&self, credential_ref: Arc<CredentialRef>) -> CoreCryptoResult<Vec<u8>> {
+        self.inner.public_key(&credential_ref.0).await.map_err(Into::into)
+    }
+}
--- a/crypto-ffi/src/core_crypto/mod.rs
+++ b/crypto-ffi/src/core_crypto/mod.rs
@@ -7,6 +7,7 @@ pub(crate) mod logger;
 pub(crate) mod mls_transport;
 mod proteus;
 
+mod credential;
 #[cfg(feature = "wasm")]
 mod randomness;
 
--- a/crypto-ffi/src/e2ei/mod.rs
+++ b/crypto-ffi/src/e2ei/mod.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 
 use async_lock::Mutex;
 use jwt_simple::prelude::{ES256KeyPair, ES384KeyPair, ES512KeyPair, Ed25519KeyPair};
-use wire_e2e_identity::{HashAlgorithm, JwsAlgorithm, acquisition::states};
+use wire_e2e_identity::{E2eIdentityError, HashAlgorithm, JwsAlgorithm, acquisition::states};
 use x509_cert::der::Encode as _;
 
 use crate::{CipherSuite as FfiCiphersuite, ClientId, CoreCryptoError, CoreCryptoResult, Credential, PkiEnvironment};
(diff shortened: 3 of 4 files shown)


crypto-ffi/src/credential_ref.rs
```diff
@@ -28,6 +28,11 @@ impl CredentialRef {
         self.0.public_key_hash().to_vec()
     }
 
+    /// Get the public key associated with this credential ref.
```
istankovic: Nit: I would say "this credential", rather than "this credential ref" because the ref is just pointing to the credential and the ref is just a thin proxy for the credential.

crypto/src/mls/credential/credential_ref/mod.rs
```diff
@@ -41,6 +42,7 @@ impl CredentialRef {
         Self {
             client_id: credential.client_id().to_owned(),
             public_key_hash: Sha256Hash::hash_from(credential.signature_key_pair.public()),
+            public_key: credential.signature_key_pair.to_public_vec().clone(),
```
istankovic: Now that I am looking at this, I don't like the fact that we're computing SHA-256 every time we construct a credential ref. Now that we have the public key with the credential ref, it is wasteful to compute and store the hash like that. It would be better to have the clients call a function to get the hash, if and when they need it and not store the hash at all.

crypto/src/mls/credential/credential_ref/mod.rs
```diff
@@ -28,6 +28,7 @@ use crate::{CipherSuite, ClientId, ClientIdRef, CredentialType};
 pub struct CredentialRef {
     client_id: ClientId,
     public_key_hash: Sha256Hash,
+    public_key: Vec<u8>,
```
SimonThormeyer: Do we really want to keep the public key in the credential ref struct?
Note WPB-22950.
We could instead provide a getter for the public key on the `CoreCrypto` instance instead.

crypto/src/mls/credential/credential_ref/mod.rs
SimonThormeyer: I don't know if and when consumers should need the hash. Storing the hash was just an optimization (see WPB-22950). So we might as well just revert what we did in that item entirely.

crypto/src/mls/credential/credential_ref/mod.rs
istankovic: This is a good point, especially having in mind  the coming PQ algorithms, where we will have pubkeys larger than 1kb (e.g. ML-KEM-768 pubkey is 1184 bytes, and that's likely the minimum we will encounter with PQ ciphersuites).

If the public key and hash are relatively rare operations, I would suggest to make both of them lazy, i.e. do not load/compute them until they are actually requested. I suspect most usage of credential refs will be purely "here's a handle to a credential, add/remove it from the conversation" so I think doing these operations lazily will be fine. If it later turns out that's not sufficiently quick, we can re-evaluate. At this point we should prioritize credential refs being minimal and quick to create.

crypto-ffi/src/core_crypto/credential.rs
```diff
@@ -0,0 +1,11 @@
+use std::sync::Arc;
+
+use crate::{CoreCryptoFfi, CoreCryptoResult, CredentialRef};
+
+#[uniffi::export]
+impl CoreCryptoFfi {
+    /// Get the public key associated with this credential
+    pub async fn public_key(&self, credential_ref: Arc<CredentialRef>) -> CoreCryptoResult<Vec<u8>> {
```
coriolinus, line 8: Every other function like this is a method on the `CredentialRef` instance itself in `crypto-ffi/src/credential_ref.rs`. I think that the API is more consistent if we keep that pattern: define 

```rust
impl CredentialRef {
  pub async fn public_key(&self, core_crypto_ffi: Arc<CoreCryptoFfi>) -> CoreCryptoResult<Vec<u8>> { ... }
}
```

But ultimately it's more important that we have the API at all, so I'm approving now.

coriolinus (approved):

crypto-ffi/src/core_crypto/credential.rs
SimonThormeyer: But that's only for cases where we don't need to load addtional data from the database. Whenever we need the DB we usually have a method on `CoreCryptoFfi`, or `TransactionContext`.

crypto-ffi/src/core_crypto/credential.rs
```diff
@@ -0,0 +1,11 @@
+use std::sync::Arc;
+
+use crate::{CoreCryptoFfi, CoreCryptoResult, CredentialRef};
+
+#[uniffi::export]
+impl CoreCryptoFfi {
+    /// Get the public key associated with this credential
```
SimonThormeyer, line 7: Maybe this doc comment should say `Load` instead of `Get`?

PR #2165, opened 2026-05-22
chore: e2e-identity: use the IdP URL provided by the ACME server in the OIDC challenge (WPB-25833)

istankovic:
Right now the new X509 credential acquisition process requires the user to specify the IdP server
URL in the configuration. However, that is both unnecessary and risky because the ACME server provides
the IdP URL in the target field of the OIDC challenge. We should always consider the data coming from the
ACME server as authoritative.

### DIFF
--- a/e2e-identity/src/acquisition/oidc_challenge.rs
+++ b/e2e-identity/src/acquisition/oidc_challenge.rs
@@ -23,9 +23,8 @@ impl X509CredentialAcquisition<states::DpopChallengeCompleted> {
         let snapshot = serde_json::to_vec(&self)?;
 
         let url = &self.data.oidc_challenge.url;
-        let id_token = hooks
-            .authenticate(self.config.idp_url.clone(), key_auth, url.to_string(), snapshot)
-            .await?;
+        let target = self.data.oidc_challenge.target.to_string();
+        let id_token = hooks.authenticate(target, key_auth, url.to_string(), snapshot).await?;
 
         let oidc_challenge_request = RustyAcme::oidc_chall_request(
             id_token,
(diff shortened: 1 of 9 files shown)


e2e-identity/src/acquisition/oidc_challenge.rs
```diff
@@ -23,9 +23,8 @@ impl X509CredentialAcquisition<states::DpopChallengeCompleted> {
         let snapshot = serde_json::to_vec(&self)?;
 
         let url = &self.data.oidc_challenge.url;
-        let id_token = hooks
-            .authenticate(self.config.idp_url.clone(), key_auth, url.to_string(), snapshot)
-            .await?;
+        let target = self.data.oidc_challenge.target.to_string();
+        let id_token = hooks.authenticate(target, key_auth, url.to_string(), snapshot).await?;
```
fewerner, line 27: Is it idomatic to use `.into() `instead of `.to_string()`?

fewerner (approved):
lgtm!

Just one comment

e2e-identity/src/acquisition/oidc_challenge.rs
istankovic: I'd say it's mostly a matter of taste. For example, I typically use `.into()` when there is a blanket impl of it because there's some `impl From` but specifying the type is too tedious (looking at our error type hierarchy :eyes: ).

In some other cases, like here, it's handy to immediately see that you're converting it to a string, without having to look at signatures, LSP etc.

Then there are situations like
```
struct Foo {
  pub s1: String,
  pub s2: String,
  pub s3: String,
  pub s4: String,
}

fn foo() -> Foo {
  Foo {
     s1: "gumo".into(),
     s2: "gumo".to_string(),
     s3: "gumo".to_owned,
     s4: String::from("gumo"),
 }
}
```
IMO, all of those are valid and it mostly comes down to personal taste.  (Naturally, the implicit assumption here is that the optimizer will produce the exact same code for all cases. :crossed_fingers: )

PR #2171, opened 2026-05-26
chore: e2e-identity: add logging to the acquisition process (WPB-21320)

istankovic:


e2e-identity/src/acquisition/oidc_challenge.rs
```diff
@@ -70,8 +83,16 @@ impl X509CredentialAcquisition<states::DpopChallengeCompleted> {
             .await?;
         let response = String::from_utf8(response.body).map_err(|e| RustyAcmeError::from(e.utf8_error()))?;
         let certificates = RustyAcme::certificate_response(response, self.data.order)?;
+        log::info!(
```
coriolinus: This might be `debug` also; we don't care much if we receive a certificate but it turns out not to be valid.

coriolinus (approved):
Nice!

e2e-identity/src/acquisition/oidc_challenge.rs
istankovic: Aye, makes sense. Changed.

PR #2185, opened 2026-05-29
feat: export pems from credentials

coriolinus:
# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/src/core_crypto/credential.rs
+++ b/crypto-ffi/src/core_crypto/credential.rs
@@ -9,7 +9,11 @@ use crate::{CipherSuite, ClientId, CoreCryptoFfi, CoreCryptoResult, CredentialRe
 impl CoreCryptoFfi {
     /// Get the public key associated with this credential
     pub async fn public_key(&self, credential_ref: Arc<CredentialRef>) -> CoreCryptoResult<Vec<u8>> {
-        self.inner.public_key(&credential_ref.0).await.map_err(Into::into)
+        credential_ref
+            .0
+            .public_key(&self.inner.database())
+            .await
+            .map_err(CoreCryptoError::generic())
     }
 
     /// Get all credentials from this client.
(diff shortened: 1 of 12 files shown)


crypto-ffi/src/core_crypto/credential.rs
```diff
@@ -1,24 +1,27 @@
 use std::sync::Arc;
 
-use core_crypto::{CipherSuite as CryptoCipherSuite, CredentialFindFilters};
+use core_crypto::{CipherSuite as CryptoCipherSuite, CredentialFindFilters, CredentialRef as CcCredentialRef};
```
typfel: nit: don't we usually rename this as `CryptoCredentialRef` in the ffi

crypto-ffi/src/core_crypto/credential.rs
```diff
@@ -109,4 +109,18 @@ impl CoreCryptoFfi {
             .map(|credentials| credentials.into_iter().map(CredentialRef::from).map(Arc::new).collect())
             .map_err(CoreCryptoError::generic())
     }
+
+    /// Export a PEM string containing the public portion of this credential.
+    ///
+    /// - Basic credentials export their public key.
+    /// - x509 credentials export the full certificate chain. This enables external tools such as `openssl` to validate
+    ///   the certificate chain.
+    pub async fn export_credential_pem(&self, credential_ref: Arc<CredentialRef>) -> CoreCryptoResult<String> {
```
fewerner, line 118: This change reminds me of the active record pattern discussion. If we somehow make the database available to the records (via Singleton) we could improve APIs like this.

typfel (changes requested):

fewerner (approved):
I love this PR. Great simplifications by placing the functions on the model instead.

Just one comment, nothing blocking.

fewerner:
One thing I forgot. Could you please add some documentation to the cc-book?

crypto-ffi/src/core_crypto/credential.rs
coriolinus: Yes, but we talked about this, I still think stateful singletons are a bad idea for us. CryptoProvider gets a pass because it's (effectively) stateless. Databases are not.

crypto-ffi/src/core_crypto/credential.rs
coriolinus: d3a670cd9

typfel (approved):

PR #2186, opened 2026-05-29
chore: re-enable x509 tests

istankovic:


crypto/src/mls/conversation/mod.rs
```diff
@@ -216,7 +214,7 @@ mod tests {
             .await
         }
 
-        // TODO: ignore this test for now, until we fix the test suite (WPB-25356)
+        // TODO: ignore this test for now, until we rework the test suite & CRL handling (WPB-19580).
```
coriolinus: Do we want to retain this reversion to an older ticket number?

coriolinus (approved):
Very impressive how small the required fixes were after restoring the tests!

crypto/src/mls/conversation/mod.rs
istankovic: Ah, good catch, I missed that. I brought back the reference to the other item. 👍

PR #2208, opened 2026-06-04
feat(ffi): add `check_credentials` to ffi

fewerner:
# What's new in this PR

This adds `TransactionContext.check_credentials` to the FFI.

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/bindings/shared/src/commonTest/kotlin/com/wire/crypto/testutils/TestUtils.kt
+++ b/crypto-ffi/bindings/shared/src/commonTest/kotlin/com/wire/crypto/testutils/TestUtils.kt
@@ -102,48 +102,56 @@ class MockPkiEnvironmentHooks : PkiEnvironmentHooks {
     }
 }
 
-sealed interface CcInitOptions {
-    val clientId: ClientId?
-    val database: Database?
+data class CcInitOptions(
+    val mode: Mode = Mode.WithBasicCredential(),
+    val clientId: ClientId? = null,
+    val database: Database? = null,
+    val withPkiEnvironment: Boolean = false,
+) {
+    sealed interface Mode {
+        data object WithoutBasicCredential : Mode
 
-    data class WithoutBasicCredential(
-        override val clientId: ClientId? = null,
-        override val database: Database? = null,
-    ) : CcInitOptions
-
-    data class WithBasicCredential(
-        val cipherSuite: CipherSuite = CIPHERSUITE_DEFAULT,
-        override val clientId: ClientId? = null,
-        override val database: Database? = null,
-    ) : CcInitOptions
+        data class WithBasicCredential(
+            val cipherSuite: CipherSuite = CIPHERSUITE_DEFAULT
+        ) : Mode
+    }
 }
 
 suspend fun ccInit(
-    options: CcInitOptions = CcInitOptions.WithBasicCredential()
+    options: CcInitOptions = CcInitOptions()
 ): CoreCrypto {
     val db = options.database ?: newDatabase()
     val cc = CoreCrypto(db)
 
     val clientId = options.clientId ?: genClientId()
 
-    cc.transaction { ctx ->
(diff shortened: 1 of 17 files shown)


istankovic:
We should add a simple test that just calls this function to all our bindings, to at least ensure the function is available.

crypto-ffi/bindings/shared/src/commonTest/kotlin/com/wire/crypto/testutils/TestUtils.kt
```diff
@@ -102,48 +102,56 @@ class MockPkiEnvironmentHooks : PkiEnvironmentHooks {
...
-        override val clientId: ClientId? = null,
-        override val database: Database? = null,
-    ) : CcInitOptions
+data class CcInitOptions(
+    val mode: Mode = Mode.WithBasicCredential(),
+    val clientId: ClientId? = null,
+    val database: Database? = null,
+    val withPkiEnvironment: Boolean = false,
+) {
+    sealed interface Mode {
```
istankovic, line 111: Shouldn't this be an enum?

fewerner: I don't think so. We couldn't attach the associated cipherSuite to `WithBasicCredential`.
In swift this is allowed, which is why we chose an enum there.

crypto/src/mls/credential/x509.rs
```diff
@@ -135,6 +138,28 @@ impl Credential {
         };
         Ok(cb)
     }
+
+    /// Check an X509 Credential for expiration or revocation
+    ///
+    /// For now we only care X509 credentials which require the pki_env as an external resource.
```
istankovic: Nit: typo? `... care X509 credentials...`

crypto/src/mls/credential/x509.rs
```diff
@@ -135,6 +138,28 @@ impl Credential {
...
+    /// As soon as we have validation logic for other types the pki_env parameter becomes very smelly.
+    /// Other credential types will just return.
+    pub(crate) async fn check(&self, pki_env: &PkiEnvironment) -> Result<()> {
+        if self.credential_type == CredentialType::X509 {
+            let cert = self
+                .mls_credential()
+                .parse_leaf_cert()
+                .map_err(RecursiveError::mls_credential("parsing leaf certificate"))?
+                // This can actually never happen and points to a type issue with credentials
+                .ok_or(Error::InvalidCredential)?;
```
istankovic: I don't understand this `ok_or` call and the comment -- could you please elaborate?

fewerner: It unwraps the `CertificateInner` from the X509 case returned `Result<Option<CertificateInner>>` and errors for the Basic case returned `Result<None>`.
If you look at the implementation of `parse_leaf_cert` you will see that because we already matched for X509 in the current scope we actually can never receive a `Result<None>` here but the compiler wants us to handle it.

To me this points to a typing issue with credentials. `parse_leaf_cert` should only exist on X509 Credentials`.

istankovic: Hm, right. Since we know that we're always going to get `Some(cert)`, I think using `expect(...)` here is more appropriate than returning a `Result`. Because if we ever hit that, it's not the error by the user of the library, but rather an error in our implementation and we want to panic in that case, not return `InvalidCredential` to the user.

crypto/src/mls/credential/x509.rs
```diff
@@ -151,7 +151,7 @@ impl Credential {
                 .parse_leaf_cert()
                 .map_err(RecursiveError::mls_credential("parsing leaf certificate"))?
                 // This can actually never happen and points to a type issue with credentials
-                .ok_or(Error::InvalidCredential)?;
+                .expect("parse_leaf_cert to return a Certifiacte");
```
istankovic: Typo in `Certifiacte`

istankovic (approved):

PR #2217, opened 2026-06-08
feat: allow removing trust anchors

istankovic:
This adds a way to remove a trust anchor, and makes use of that to re-enable one of our tests.

### DIFF
--- a/e2e-identity/src/pki_env/mod.rs
+++ b/e2e-identity/src/pki_env/mod.rs
@@ -16,7 +16,7 @@ use core_crypto_keystore::{
     CryptoKeystoreError,
     connection::Database,
     entities::{E2eiAcmeCA, E2eiCrl, E2eiIntermediateCert},
-    traits::FetchFromDatabase,
+    traits::{FetchFromDatabase, UnifiedUniqueEntity},
 };
 use openmls_traits::authentication_service::{CredentialAuthenticationStatus, CredentialRef};
 use x509_cert::{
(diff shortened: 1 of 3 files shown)


e2e-identity/src/pki_env/mod.rs
```diff
@@ -215,6 +215,49 @@ impl PkiEnvironment {
...
+                TrustAnchorChoice::Certificate(ref cert)
+                    if cert.tbs_certificate.serial_number.as_bytes() != serial_number =>
+                {
+                    Some(cert.clone())
+                }
+                _ => None,
+            })
+            .collect();
+
+        guard.clear_trust_anchor_sources();
```
typfel, line 238: So we removing all trust anchors and then adding them all back again except for the one we removed.

istankovic: Exactly. Unfortunately, there is no other way to do it, due to the underlying API of rust-pki. I'm not sure if the API is better on `cert_limbo`, but I wouldn't bet on it. I was thinking we could probably improve that by submitting a PR to rust-pki, but that would only be for later, once we move to a `cert_limbo`-based release of rust-pki.

e2e-identity/src/pki_env/mod.rs
```diff
@@ -215,6 +215,49 @@ impl PkiEnvironment {
         Ok(())
     }
 
+    /// Remove the trust anchor with serial number `serial_numer` from the PKI environment.
```
fewerner: nit: typo in `serial_number`

typfel (approved):

e2e-identity/src/pki_env/mod.rs
istankovic: Thanks, fixed! :purple_heart:

PR #2221, opened 2026-06-09
refactor: make `Database: !Clone` [WPB-25641]

coriolinus:
# What's new in this PR

----

- The **PR Title**

### DIFF
--- a/keystore/src/connection/mod.rs
+++ b/keystore/src/connection/mod.rs
@@ -195,7 +195,11 @@ impl Database {
     pub async fn conn(&self) -> CryptoKeystoreResult<ConnectionGuard<'_>> {
         self.conn.lock().await.try_into()
     }
+}
 
+// These and all other database impls shold not refer directly to `self.conn` but should go through the `self.conn()`
+// wrapper
+impl Database {
     /// Wait for any running transaction to finish, then take the connection out of this database,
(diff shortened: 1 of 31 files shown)


keystore/src/connection/mod.rs
```diff
@@ -131,19 +133,18 @@ impl DerefMut for ConnectionGuard<'_> {
 
 // Only the functions in this impl block directly mess with `self.conn`
 impl Database {
-    pub async fn open(location: ConnectionType<'_>, key: &DatabaseKey) -> CryptoKeystoreResult<Self> {
+    pub async fn open(location: ConnectionType<'_>, key: &DatabaseKey) -> CryptoKeystoreResult<Arc<Self>> {
```
istankovic, line 136: I'm thinking, wouldn't it be better to leave `Arc` creation to the caller? That is, to leave the decision whether to use an `Arc` to the caller -- in some cases you don't actually need it.

coriolinus: It's immaterial here: we're always being called via FFI, and the FFI version of a `Database` always contains an `Arc`'d version of the keystore `Database`. So IMO it's simplest to just always create it.

istankovic (approved):
Looks good!

PR #2222, opened 2026-06-09
refactor!: unify client ids on ffi level [WPB-26090]

SimonThormeyer:
# What's new in this PR
This replaces `String` client id types with `ClientId` on the ffi layer and ensures that all client ids (except history clients) conform to the format `<uuid>:<device id>@domain`.

----

- The **PR Title**


crypto/src/mls/session/id.rs
```diff
@@ -7,50 +7,123 @@ use std::{
...
-    derive_more::Into,
-    serde::Serialize,
-    serde::Deserialize,
+    core_crypto_macros::Debug, Clone, Eq, PartialOrd, Ord, Hash, derive_more::Into, serde::Serialize, serde::Deserialize,
 )]
 #[sensitive]
-pub struct ClientId(pub(crate) Vec<u8>);
+pub struct ClientId(Vec<u8>);
+
+pub struct SerializedClientId {
```
coriolinus: Wouldn't this be `DeserializedClientId`?

crypto/src/mls/session/id.rs
```diff
@@ -7,50 +7,123 @@ use std::{
...
+    /// Create a new client ID.
+    pub fn new(user_id: &str, device_id: &str, domain: &str) -> Result<Self> {
+        let delimiter = Self::DELIMITER;
+        let string = format!("{user_id}{delimiter}{device_id}@{domain}");
+        let bytes = string.into_bytes();
+        Self::try_parse_bytes(&bytes)?;
+        Ok(Self(bytes))
+    }
+
+    pub fn serialize(&self) -> SerializedClientId {
```
coriolinus: Same, it's weird that we "serialize" a byte vector to produce a struct with distinct fields.

crypto/src/mls/session/id.rs
```diff
@@ -7,50 +7,123 @@ use std::{
...
 )]
 #[sensitive]
-pub struct ClientId(pub(crate) Vec<u8>);
+pub struct ClientId(Vec<u8>);
+
+pub struct SerializedClientId {
+    pub user_id: Uuid,
+    pub device_id: String,
+    pub domain: String,
+}
```
coriolinus, line 28: I assume that at some future point in the commit sequence we'll swap the semantics of which is serialized vs the default client ID?

crypto/src/mls/session/id.rs
```diff
@@ -43,6 +43,14 @@ impl ClientId {
         Ok(Self(bytes))
     }
 
+    pub(crate) fn new_ephemeral() -> Self {
+        // generate a new completely arbitrary client id
+        let session_id = uuid::Uuid::new_v4();
+        let session_id = format!("{HISTORY_CLIENT_ID_PREFIX}-{session_id}");
```
coriolinus: There's no particular reason we have to violate the format. It could as easily be `format!("{session_id}:ephemeral@{HISTORY_CLIENT_ID_SUFFIX}")`.

crypto/src/mls/conversation/mutable/commit.rs
```diff
@@ -127,23 +127,29 @@ impl ConversationMut {
     /// * `id` - group/conversation id
     /// * `clients` - list of client ids to be removed from the group
     pub async fn remove_members(&mut self, clients: &[impl Borrow<ClientIdRef>]) -> Result<()> {
+        self.remove_members_or_history_clients(clients.iter().map(|e| e.borrow().as_ref()))
+            .await
+    }
+
+    pub(crate) async fn remove_members_or_history_clients(
```
coriolinus, line 139: IMO adjusting the ephemeral client IDs to conform to the format is a better choice than adding this API.

crypto/src/identity.rs
```diff
@@ -1,17 +1,15 @@
...
-use crate::CredentialType;
+use crate::{ClientId, CredentialType, RecursiveError};
 
 /// Represents the identity claims identifying a client
 /// Those claims are verifiable by any member in the group
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
 pub struct WireIdentity {
     /// Unique client identifier e.g. `T4Coy4vdRzianwfOgXpn6A:6add501bacd1d90e@whitehouse.gov`
-    pub client_id: String,
+    pub client_id: Option<ClientId>,
```
coriolinus, line 12: Why is this optional? I wouldn't have expected it to be, previously.

crypto-ffi/src/client_id/serialize.rs
```diff
@@ -0,0 +1,29 @@
+use std::sync::Arc;
+
+use crate::ClientId;
+
+/// This directlry represents a `ClientId` of the `<userid>-<device-id>@<domain>` format.
+/// Instantiate via [ClientId::serialize].
+#[derive(Debug, uniffi::Record)]
+pub struct SerializedClientId {
```
coriolinus: Maybe `ParsedClientId`? But it's not obvious to me why this parsed representation isn't how the base `ClientId` type is implemented.

crypto-ffi/bindings/js/packages/browser/shared/utils.ts
```diff
@@ -73,13 +73,16 @@ export async function sharedSetup() {
             /**
              * Construct a new ClientId
              **/
-            static newClientId(clientIdStr?: string): ClientId {
-                if (clientIdStr === undefined) {
-                    clientIdStr = window.crypto.randomUUID();
-                }
-                const encoder = new TextEncoder();
+            static newClientId(): ClientId {
+                const userId = crypto.randomUUID();
```
coriolinus, line 77: We could probably export a `ClientId::random()` function which clients could use for tests; it should be obvious (and documented) that in real life you'd get a client id from the backend, but it seems harmless to move this kind of thing into the Rust layer.

SimonThormeyer: That's a good call.

crypto/src/mls/session/id.rs
SimonThormeyer: After discussion with backend, we decided to leave this as is

crypto/src/mls/conversation/mutable/commit.rs
SimonThormeyer: After discussion with backend we decided to leave this as is

crypto/src/identity.rs
SimonThormeyer: Previously, in some cases, it could be an empty string. In those cases it will be `None` now. Ensuring that these cases cannot occur anymore would be out of scope of this PR imo.

crypto-ffi/bindings/js/packages/browser/shared/utils.ts
```diff
@@ -73,14 +73,8 @@ export async function sharedSetup() {
...
-            static newClientId(clientIdStr?: string): ClientId {
-                if (clientIdStr === undefined) {
-                    clientIdStr = window.crypto.randomUUID();
-                }
-                const encoder = new TextEncoder();
-                return new window.ccModule.ClientId(
-                    encoder.encode(clientIdStr)
-                );
+            static newClientId(): ClientId {
+                return window.ccModule.ClientId.random();
```
fewerner: At this point we should drop this helper entirely as we do on ts-native.

SimonThormeyer: Really? I still find `window.helpers.newClientId()` nicer than `window.ccModule.ClientId.random()`

fewerner: Imo, we should aim for symmetry with other platforms. I find it hard to read and write tests when every platform has its own flavor of helpers and we've been fighting this recently.
I know it isn't fully aligned yet. I do this as part of my ongoing work.

These lines are obviously a negligible difference, but I'd like us to keep this contract.

fewerner: I'd also imagine that if we wanted to come up with a shared test infra, it's easier if things like these are structurally equivalent between platforms.

crypto/src/mls/session/id.rs
```diff
@@ -7,50 +7,124 @@ use std::{
...
+}
 
 impl ClientId {
+    /// user-id & device-id separator
+    pub const DELIMITER: &'static str = ":";
+
+    /// Create a new client ID.
+    pub fn new(user_id: &str, device_id: &str, domain: &str) -> Result<Self> {
+        let delimiter = Self::DELIMITER;
+        let string = format!("{user_id}{delimiter}{device_id}@{domain}");
```
coriolinus: Kind of surprising that we use a constant for the `:` but not for the `@`. I'd prefer to use constants for both or neither.

crypto/src/mls/session/id.rs
```diff
@@ -7,50 +7,124 @@ use std::{
...
+        let bytes = string.into_bytes();
+        Self::try_parse_bytes(&bytes)?;
```
coriolinus, line 44: This parse attempt is only useful as a double-check of our format string; it's entirely wasted in production. How about using `#[cfg(debug_assertions)]` to omit the parse in release builds?

crypto-ffi/bindings/js/packages/browser/shared/utils.ts
SimonThormeyer: But what do we do in cases where a helper provides real value on platform A and is just silly in Platform B? Drop it at the cost of ergonomics or implement it for both platforms?

crypto/src/mls/session/id.rs
```diff
@@ -7,50 +7,124 @@ use std::{
...
+    }
+
+    /// Parse the user id, assuming string representation of a UUIDv4.
+    /// TODO(SimonThormeyer): Should this be the base64-encoded string instead?
+    fn parse_user_id(user_id: &str) -> Result<Uuid> {
+        // let user_id = base64::prelude::BASE64_URL_SAFE_NO_PAD
+        //     .decode(user_id)
+        //     .map_err(|_| Error::InvalidQualifiedClientId)?;
+        let uuid = Uuid::try_parse(user_id).map_err(|_| Error::InvalidQualifiedClientId)?;
+        Ok(uuid)
```
coriolinus: Probably simplify this whole thing down to the one-liner

```rust
Uuid::try_parse(user_id).map_err(|_| Error::InvalidQualifiedClientId)
```

crypto/src/mls/session/id.rs
```diff
@@ -7,50 +7,124 @@ use std::{
...
+    }
+
+    fn parse_device_id(device_id: &str) -> Result<u64> {
+        u64::from_str_radix(device_id, 16).map_err(|_| Error::InvalidQualifiedClientId)
```
coriolinus, line 106: Are their length expectations for the device ID that we could parse, or is this anything which parses from hexadecimal into a u64?

crypto/src/mls/session/id.rs
```diff
@@ -7,50 +7,124 @@ use std::{
...
 }
 
-impl<const N: usize> From<[u8; N]> for ClientId {
-    fn from(value: [u8; N]) -> Self {
-        Self(value.into())
+impl TryFrom<Box<[u8]>> for ClientId {
+    type Error = Error;
+
+    fn try_from(value: Box<[u8]>) -> Result<Self> {
+        value.try_into()
```
coriolinus: Isn't this implementation infinitely recursive? I think you may have meant `(&value).try_into()`.

crypto/src/mls/session/id.rs
```diff
@@ -7,50 +7,124 @@ use std::{
...
 }
 
-impl<'a> From<&'a [u8]> for ClientId {
-    fn from(value: &'a [u8]) -> Self {
-        Self(value.into())
+impl<const N: usize> TryFrom<[u8; N]> for ClientId {
+    type Error = Error;
+
+    fn try_from(value: [u8; N]) -> Result<Self> {
+        value.try_into()
```
coriolinus: Same here.

crypto/src/mls/session/id.rs
```diff
@@ -40,6 +40,22 @@ impl ClientId {
         Ok(Self(bytes))
     }
 
+    /// In Some cases, we still have a base64-encoded user id, e.g., certifactes.
```
coriolinus: capitalization typo: "Some cases"

crypto-ffi/bindings/js/packages/browser/shared/utils.ts
SimonThormeyer: Also, whether we have this or not, helpers are currently structurally inequivalent in native and browser anyway: in native, they're free functions you have to know specifically about, while in browser, you can access them on `window.helpers`. This situation doesn't get worse with a single helper that exists in the browser helpers, and doesn't have a counter part in native.

However, I don't feel strongly about this, do you prefer to have the helper removed in browser or re-added in native?

crypto-ffi/src/client_id/mod.rs
```diff
@@ -1,18 +1,29 @@
...
+    Eq,
+    Hash,
+    derive_more::From,
+    derive_more::Into,
+    derive_more::Deref,
+    derive_more::DerefMut,
+    uniffi::Object,
+)]
+#[uniffi::export(Eq, Hash)]
+pub struct ClientId(core_crypto::ClientId);
```
coriolinus, line 27: This is ok, but what went wrong with just switching to 

```rust
bytes_wrapper!(ClientId fallibly wraps core_crypto::ClientId);
```

?

crypto-ffi/src/client_id/serialize.rs
```diff
@@ -0,0 +1,29 @@
+use std::sync::Arc;
+
+use crate::ClientId;
+
+/// This directlry represents a `ClientId` of the `<userid>-<device-id>@<domain>` format.
```
coriolinus: Typo: "directlry"

crypto-ffi/src/client_id/serialize.rs
```diff
@@ -0,0 +1,29 @@
...
+/// Instantiate via [ClientId::deserialize].
+#[derive(Debug, uniffi::Record)]
+pub struct DeserializedClientId {
+    /// The client id this was deserialized from
+    pub client_id: Arc<ClientId>,
+    /// The string representation of a UUID
+    pub user_id: String,
+    /// A hex-encoded unsigned 64-bit integer
+    pub device_id: String,
```
coriolinus, line 14: Should this be a u64 then?

crypto-ffi/src/client_id/mod.rs
```diff
@@ -42,6 +42,16 @@ impl ClientId {
         Ok(Self(inner))
     }
 
+    /// This should only be used for testing purposes, where generating a fresh `ClientId` is useful. In reality, the
+    /// client id has to be agreed on with the delivery service.
```
coriolinus: Verbiage nit: I'd say "is provided by" instead of "has to be agreed on with". It's not like there's a client id negotiation process.

coriolinus (commented):
- https://github.com/wireapp/core-crypto/commit/198f600b0492b966bfa5d86106d2b5b43979e07f has a misleading commit message; it doesn't conform to the format anymore. Also it references a function which hasn't been defined yet in the commit sequence; e5ee91a3243efb554086cb4fb7dc77be76ba6639 should be moved ahead of it in sequence.
- 90339b29d958b70cadfb355083cbfda98d8df973 has a typo in its commit message: "credenital"
- Might add a section to the kotlin-specific migration guide that we removed ".toClientId()"

Lots of nits, but overall the direction here is good! I think we're nearly there.

crypto-ffi/bindings/js/packages/browser/shared/utils.ts
fewerner: Ofc, we would allow ergonomics.

I just don't see why we would decide differently between 
`newClientId()` and `ClientId.random()` in ts-native 
and `window.helpers.newClientId()` vs. `window.ccModule.ClientId.random()` in ts-browser.

Both avoid exactly one access.

crypto/src/mls/session/id.rs
SimonThormeyer: No, it's useful to check the data a library user passes in production:
- is the user id string really a UUID?
- is the device id string really a hex-encoded u64?

crypto/src/mls/session/id.rs
SimonThormeyer: The string always has a length of 16, so we expect leading `0`s. Is that what you were asking?

crypto/src/mls/session/id.rs
SimonThormeyer: Good catch!

crypto-ffi/src/client_id/mod.rs
SimonThormeyer: Maybe I read the macro wrong, but wouldn't this emit a `new(bytes: Vec<u8>)` constructor? We don't want to expose a constructor like that, consumers are expected to use the `new(user_id, device_id, domain)`

crypto-ffi/src/client_id/serialize.rs
SimonThormeyer: Purely from a data-perspective, yes, but I think in this case we really want a string (but also don't feel strongly about it), because, AFAIK, users are used to seeing this string and passing it in as such a string.

crypto/src/mls/session/id.rs
coriolinus: We could trivially require them to just pass a `Uuid` instance and `u64` and avoid that work.

crypto/src/mls/session/id.rs
coriolinus: Yes. If that's the format, we can enforce it here.

crypto-ffi/src/client_id/mod.rs
coriolinus: Fair enough, that's reasonable.

crypto-ffi/src/client_id/serialize.rs
coriolinus: It feels to me like the deserialized type should focus on getting the data model correct. We can always add helpers like `device_id_encoded_string()` if people turn out to want them.

coriolinus (approved):
At this point I'm happy with it once CI is green. We can of course talk more about the interior data model etc., but nothing there is really blocking, just style differences.

Nice work!

crypto/src/mls/session/id.rs
```diff
@@ -1,56 +1,162 @@
...
+            "{user_id}{}{device_id}{}{domain}",
+            Self::DELIMITER,
+            Self::DOMAIN_SEPERATOR
+        );
+        let bytes = string.into_bytes();
+        Self::try_parse_bytes(&bytes)?;
+        Ok(Self(bytes))
+    }
+
+    /// In some cases, we still have a base64-encoded user id, e.g., certifactes.
```
istankovic: Nit: typo in `certifactes`

crypto/src/mls/session/id.rs
```diff
@@ -1,56 +1,162 @@
...
+        let user_id = Uuid::new_v4().hyphenated().to_string();
+        let bytes = format!("{HISTORY_CLIENT_ID_PREFIX}{}{user_id}", Self::DELIMITER).into_bytes();
+        Self(bytes)
+    }
+
+    /// This should only be used for testing purposes, where generating a fresh `ClientId` is useful. In reality, the
+    /// client id is provided by the delivery service.
+    ///
+    /// Note: only the user id and device id is going to be random, the domain is going to be `wire.com`.
+    pub fn random() -> Self {
```
istankovic: This should be `#[cfg(test)]`.

coriolinus: No it shouldn't: we want to expose it over FFI.

istankovic: Why?

istankovic:
A number of trait impls are unused and seem unnecessary (most of them were added before this PR, I think).
It would be nice to clean that up:

```diff
diff --git a/crypto/src/mls/session/id.rs b/crypto/src/mls/session/id.rs
index 16775e102..1794e5f35 100644
--- a/crypto/src/mls/session/id.rs
+++ b/crypto/src/mls/session/id.rs
@@ -144,28 +144,6 @@ impl TryFrom<&[u8]> for ClientId {
     }
 }
 
-impl TryFrom<Box<[u8]>> for ClientId {
-    type Error = Error;
-
-    fn try_from(value: Box<[u8]>) -> Result<Self> {
-        value.as_ref().try_into()
-    }
-}
-
-impl<const N: usize> TryFrom<[u8; N]> for ClientId {
-    type Error = Error;
-
-    fn try_from(value: [u8; N]) -> Result<Self> {
-        value.as_ref().try_into()
-    }
-}
-
-impl From<ClientId> for Box<[u8]> {
-    fn from(value: ClientId) -> Self {
-        value.0.into_boxed_slice()
-    }
-}
-
 impl Deref for ClientId {
     type Target = ClientIdRef;
 
@@ -186,26 +164,12 @@ impl AsRef<ClientIdRef> for ClientId {
     }
 }
 
-impl From<ClientId> for Cow<'_, [u8]> {
-    fn from(value: ClientId) -> Self {
-        Cow::Owned(value.0)
-    }
-}
-
 impl std::fmt::Display for ClientId {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "{}", hex::encode(self.0.as_slice()))
     }
 }
 
-impl std::str::FromStr for ClientId {
-    type Err = Error;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        Ok(Self(hex::decode(s).unwrap_or_else(|_| s.as_bytes().to_vec())))
-    }
-}
-
 impl<T> PartialEq<T> for ClientId
 where
     ClientIdRef: PartialEq<T>,
@@ -215,13 +179,6 @@ where
     }
 }
 
-#[cfg(test)]
-impl From<&str> for ClientId {
-    fn from(value: &str) -> Self {
-        Self(value.as_bytes().into())
-    }
-}
-
 /// Reference to a [`ClientId`].
 ///
 /// This type is `!Sized` and is only ever seen as a reference, like `str` or `[u8]`.
```

istankovic:
We have
```
impl std::fmt::Display for ClientId {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}", hex::encode(self.0.as_slice()))
    }
}
```
while at the same time we take care to mark `ClientId` as `sensitive` and derive `core_crypto_macros::Debug` on it.
I think we should just remove this impl.

crypto/src/mls/session/id.rs
SimonThormeyer: https://github.com/wireapp/core-crypto/pull/2222#discussion_r3381220180

istankovic: I must say I disagree with that.

We should not expose such things across FFI. Client code, as well as our own binding tests, should have their own test utilities, not rely on us to provide that.

This needlessly increases the binary size, if only by a bit, and will never be used in production. Hopefully.
This is one of those things that may look at first like a good idea, but I think it's just wrong.

SimonThormeyer:
> We have
> 
> ```
> impl std::fmt::Display for ClientId {
>     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
>         write!(f, "{}", hex::encode(self.0.as_slice()))
>     }
> }
> ```
> 
> while at the same time we take care to mark `ClientId` as `sensitive` and derive `core_crypto_macros::Debug` on it. I think we should just remove this impl.

Just removing this blows up in some places, we'd need an alternative. If that's desired, I'd like to do that as a follow-up item

crypto-ffi/src/client_id/serialize.rs
SimonThormeyer: I'll make a follow-up PR that solves this and also the unnecessary parsing you're mentioning above. We can then engage in a focused discussion about API and data

PR #2228, opened 2026-06-10
test: unignore last ignored test

istankovic:


crypto/src/mls/conversation/mod.rs
```diff
@@ -124,16 +124,16 @@ mod tests {
 
         async fn check_identities_device_status<const N: usize>(
             conversation: &Conversation,
-            client_ids: &[ClientId; N],
-            name_status: &[(impl ToString, DeviceStatus); N],
+            client_status: &[(ClientId, DeviceStatus); N],
         ) {
-            let mut identities = conversation.get_device_identities(client_ids).await.unwrap();
+            let client_ids: Vec<_> = client_status.iter().map(|(client_id, _)| client_id).collect();
```
coriolinus: I'm not really a fan of this factoring of the parameters, because it forces an otherwise totally unnecessary iteration/collection cycle. Why not write it like

```rust
async fn check_identities_device_status<const N: usize>(
    conversation: &Conversation,
    client_ids: &[ClientId; N],
    device_status: &[DeviceStatus; N],
)
```

? 

It means you don't have to go back and reconstruct `client_ids` to pass to `get_device_identities`, but you still know that there is a 1:1 match between client id and device status.

For the later iteration, it's simple enough to just do 

```rust
for (client_id, status) in client_ids.iter().zip(device_status.iter()) { ... }
```

crypto/src/test_utils/mod.rs
```diff
@@ -203,8 +203,15 @@ impl SessionContext {
...
-        let bundle = crate::CertificateBundle::new_with_exact_client_id(client_id, chain.find_local_intermediate_ca());
+        let issuer = chain.find_local_intermediate_ca();
+        let actor = chain
+            .actors
+            .iter()
+            .find(|actor| actor.client_id == client_id.to_string_triple()[0]);
+        let bundle = match actor {
+            Some(actor) => CertificateBundle::from_certificate_and_issuer(&actor.certificate, issuer),
+            None => CertificateBundle::new_with_exact_client_id(client_id, chain.find_local_intermediate_ca()),
+        };
```
coriolinus: Could simplify:

```rust
        let bundle = chain
            .actors
            .iter()
            .find(|actor| actor.client_id == client_id.to_string_triple()[0])
            .map(|actor| CertificateBundle::from_certificate_and_issuer(&actor.certificate, issuer))
            .unwrap_or_else(|| {
                CertificateBundle::new_with_exact_client_id(client_id, chain.find_local_intermediate_ca())
            });
```

coriolinus (commented):
We can do better in `check_identities_device_status`, but otherwise, great work!

crypto/src/mls/conversation/mod.rs
istankovic: Heh, it's so obvious now that you say it.  :-)
Changed, thanks.

crypto/src/test_utils/mod.rs
istankovic: Done, thanks.

coriolinus (approved):

PR #2233, opened 2026-06-11
test(ts): align ts-browser and ts-native tests [WPB-26181]

fewerner:
# What's new in this PR

This PR aligns ts-browser and ts-native tests.
We first align some util and then resolve test differences step by step.

----

- The **PR Title**

### DIFF
--- a/crypto-ffi/bindings/js/packages/browser/shared/utils.ts
+++ b/crypto-ffi/bindings/js/packages/browser/shared/utils.ts
@@ -337,7 +337,7 @@ export async function sharedSetup() {
              * @param message The message encrypted, sent, and decrypted once in each
              * direction.
              *
-             * @returns {Promise<(string | null)[]>} A two-element list, containing the decrypted {@link message} by
+             * @returns {Promise<(string)[]>} A two-element list, containing the decrypted {@link message} by
              * {@link client1} and {@link client2}, in that order.
              */
             static async roundTripMessage(
@@ -345,7 +345,7 @@ export async function sharedSetup() {
                 cc2: CoreCrypto,
                 conversationId: ConversationId,
                 message: string
-            ): Promise<(string | null)[]> {
+            ): Promise<string[]> {
                 const encoder = new TextEncoder();
                 const messageBytes = encoder.encode(message);
 
@@ -384,14 +384,8 @@ export async function sharedSetup() {
                 );
 
                 const decoder = new TextDecoder();
-                const result1 =
-                    decryptedByClient1.message !== undefined
-                        ? decoder.decode(decryptedByClient1.message)
-                        : null;
-                const result2 =
-                    decryptedByClient2.message !== undefined
-                        ? decoder.decode(decryptedByClient2.message)
-                        : null;
+                const result1 = decoder.decode(decryptedByClient1.message);
+                const result2 = decoder.decode(decryptedByClient2.message);
 
                 return [result1, result2];
             }
--- a/crypto-ffi/bindings/js/packages/browser/shared/utils.ts
+++ b/crypto-ffi/bindings/js/packages/browser/shared/utils.ts
@@ -178,9 +178,9 @@ export async function sharedSetup() {
              * Records logs by setting a logger and maximum log level in the browser's context.
              * The logs are stored in a global variable `window.recordedLogs` for further retrieval.
              *
-             * @return {Promise<void>}
+             * @return {void}
              */
-            static async recordLogs(): Promise<void> {
+            static recordLogs(): void {
                 const { setMaxLogLevel, CoreCryptoLogLevel, setLogger } =
                     window.ccModule;
                 window.recordedLogs = [];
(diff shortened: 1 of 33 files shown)


crypto-ffi/bindings/js/packages/browser/test/clientIdentity.test.ts
```diff
@@ -16,7 +16,7 @@ describe("client identity", () => {
             const cc = await window.helpers.ccInit();
             return (await cc.getCredentials())[0]!.publicKeyHash().byteLength;
         });
-        await expect(result).toBe(32);
+        expect(result).toBe(32);
     });
```
SimonThormeyer, line 20: So 72cca478950cd5b17a463ff16cf6cb3f30187c59 was wrong?

crypto-ffi/bindings/js/packages/browser/shared/utils.ts
```diff
@@ -342,7 +342,7 @@ export async function sharedSetup() {
                 cc2: CoreCrypto,
                 conversationId: ConversationId,
                 message: string
-            ): Promise<(string | null)[]> {
+            ): Promise<string[]> {
```
coriolinus, line 337: You didn't update the `@params` doc above.

SimonThormeyer:
The commit message of faafd9aa137f5114f9f069d7cfd26cc73174f59a is confusing:
> We also allow to specify the clientId, aligning with the browser helper.

But that's not the case.

> This also fixes a bug.
`Uint8Array.from(uuid)` treats the string as an Iterable of chars, which
is wrong. Instead we want proper binary encoding.

But there's nothing like that in the hunk. 

Could it be that the message describes a change that was overwritten?

crypto-ffi/bindings/js/packages/native/test/epochObserver.test.ts
```diff
@@ -11,63 +10,53 @@ afterEach(async () => {
...
-                    convIdSerialized,
-                };
-            });
-
-        expect(length).toEqual(1);
+    test("should observe new epochs", async () => {
+        // set up the observer. this just keeps a list of all observations.
+        type ObservedEpoch = {
+            // @ts-expect-error `window` is not present when ts is checking, but is present in the browser
+            conversationId: window.ccModule.ConversationId;
```
SimonThormeyer: Why are we using `window` in native context?

crypto-ffi/bindings/js/packages/native/test/epochObserver.test.ts
```diff
@@ -11,63 +10,53 @@ afterEach(async () => {
...
+            epoch: bigint;
+        };
+        class Observer {
+            observations: ObservedEpoch[];
+            constructor() {
+                this.observations = [];
+            }
+            async epochChanged(
+                // @ts-expect-error `window` is not present when ts is checking, but is present in the browser
+                conversationId: window.ccModule.ConversationId,
```
SimonThormeyer: Why are we using `window` in native context?

crypto-ffi/bindings/js/packages/native/test/logger.test.ts
```diff
@@ -10,138 +23,120 @@ afterEach(async () => {
...
+}
+
+/**
+ * Retrieves the logs recorded on the browser-side.
+ *
+ * @return {Promise<LogEntry[]>} A promise that resolves to an array of log entries
+ */
+async function retrieveLogs(): Promise<LogEntry[]> {
+    return RECORDED_LOGS.map((entry) => {
+        return entry;
```
SimonThormeyer: What does this `map()` call do? Woudln't it be sufficient to just return `RECORDED_LOGS` directly? And does the function need to be async?

SimonThormeyer (approved):
Some nits/open questions still, but nothing critical

crypto-ffi/bindings/js/packages/native/test/epochObserver.test.ts
fewerner: I didn't catch that because of the `ts-expect-error`. Good Catch!

fewerner:
> The commit message of [faafd9a](https://github.com/wireapp/core-crypto/commit/faafd9aa137f5114f9f069d7cfd26cc73174f59a) is confusing:
> 
> > We also allow to specify the clientId, aligning with the browser helper.
> 
> But that's not the case.
> 
> > This also fixes a bug.
> > `Uint8Array.from(uuid)` treats the string as an Iterable of chars, which
> > is wrong. Instead we want proper binary encoding.
> 
> But there's nothing like that in the hunk.
> 
> Could it be that the message describes a change that was overwritten?

yeah, that message doesn't make sense anymore. That happened due to rebasing, the message is obsolete.

crypto-ffi/bindings/js/packages/native/test/logger.test.ts
fewerner: I think this is completely obsolete also on browser. We should just access the logs directly.

PR #2248, opened 2026-06-12
refactor!: finalize client id initialization API [WPB-26090]

SimonThormeyer:
# What's new in this PR
See above

----

- The **PR Title**


crypto/src/mls/session/id.rs
```diff
@@ -34,15 +34,17 @@ impl ClientId {
     pub const DOMAIN_SEPERATOR: &'static str = "@";
 
     /// Create a new client ID.
-    pub fn new(user_id: &str, device_id: &str, domain: &str) -> Result<Self> {
+    pub fn new(user_id: &Uuid, device_id: u64, domain: &str) -> Self {
```
coriolinus: `Uuid` is `Copy`, so we don't need a reference.

crypto/src/mls/conversation/immutable/credential.rs
```diff
@@ -60,17 +60,20 @@ impl super::Conversation {
     }
 
     /// Returns all members credentials with their signature public key from the group/conversation
-    pub async fn members_with_key(&self) -> HashMap<Vec<u8>, CredentialWithKey> {
-        self.group().await.members().fold(HashMap::new(), |mut acc, kp| {
+    pub async fn members_with_key(&self) -> Result<HashMap<ClientId, CredentialWithKey>> {
+        self.group().await.members().try_fold(HashMap::new(), |mut acc, kp| {
```
coriolinus: This is better written as `.map().collect()` than `.try_fold()`.

crypto/src/mls/conversation/immutable/e2ei.rs
```diff
@@ -61,35 +62,31 @@ impl super::Conversation {
...
+    pub async fn get_user_identities(&self, user_ids: &[Uuid]) -> Result<HashMap<Uuid, Vec<WireIdentity>>> {
         if user_ids.is_empty() {
             return Err(Error::CallerError(
                 "This function accepts a list of IDs as a parameter, but that list was empty.",
             ));
         }
         let mls_provider = &self.session.crypto_provider;
         let auth_service = mls_provider.authentication_service();
-        let user_ids = user_ids.iter().map(|uid| uid.as_bytes()).collect::<Vec<_>>();
+        let user_ids = user_ids.iter().collect::<Vec<_>>();
```
coriolinus: This iteration/collection isn't required at all.

coriolinus (approved):
This is all fine with some nits, and I'm approving it here. 

However, I do think that this would be an excellent opportunity to reconsider why `ClientId(Vec<u8>)` is the primary type and `DeserializedClientId` is the supporting type. This would be an excellent opportunity to rename things such that `ClientId` just stores the fields as `DeserializedClientId` currently does, and then instead of providing a `deserialize` method producing the values, we could have a `serialize` method producing a `Vec<u8>`. Or a type-safe wrapper if that seems like it would be useful.

PR #2303, opened 2026-07-02
refactor: use unified connection [WPB-23775]

coriolinus:
# What's new in this PR

Partially implements [23775](https://wearezeta.atlassian.net/browse/WPB-23775). This accomplishes _most_ of that PR, except for two more steps:

- [x] Rename all `Unified*` entity trait family traits to rm the `Unified` prefix
- [x] Fix all the errors which fall out of this API changes in CC.

One issue I see right away: this new `Database` wraps a connection directly, where the old one wrapped a `Mutex<Option<Connection>>`. This means we can no longer _use_ the `Database::close` method, because it consumes `self` and consumers all always have an `Arc<Database>`. 

Whoever picks this up will need to either adjust that and all the access patterns, or determine that we can omit `close` methods entirely now.

----

- The **PR Title**


keystore/src/mls.rs
```diff
@@ -104,125 +153,35 @@ pub trait CryptoKeystoreMls: Sized {
...
     ///
     /// # Arguments
     /// * `id` - hash of the enrollment and unique identifier
-    async fn pop_e2ei_enrollment(&self, id: &[u8]) -> CryptoKeystoreResult<Option<Vec<u8>>>;
-}
-
-#[cfg_attr(target_os = "unknown", async_trait::async_trait(?Send))]
-#[cfg_attr(not(target_os = "unknown"), async_trait::async_trait)]
-impl CryptoKeystoreMls for crate::Database {
-    async fn mls_fetch_key_packages<V: MlsEntity>(&self, count: u32) -> CryptoKeystoreResult<Vec<V>> {
```
SimonThormeyer, line 126: I can't put into words how thankful I am that this is removed.

keystore/tests/z_entities.rs
```diff
@@ -223,7 +203,6 @@ mod tests {
         ));
 
         // It's required by cleanup to have a running transaction before finishing the test
-        store.rollback_transaction().await.unwrap();
```
SimonThormeyer, line 226: Is this comment still required and correct?

crypto/src/lib.rs
```diff
@@ -33,7 +33,7 @@ use std::sync::Arc;
 use async_lock::Mutex;
 use async_lock::RwLock;
 pub(crate) use bytes_wrapper::bytes_wrapper;
-pub use core_crypto_keystore::{ConnectionType, Database, DatabaseKey};
+pub use core_crypto_keystore::{Database, DatabaseKey};
```
SimonThormeyer, line 36: 0c994df0a3f8e8d5e81b8b727cd6056cb8723e82 could be folded into 30c322a5e29434c686bb4f9c4bea7ab7105f8771

crypto/src/mls/session/mod.rs
```diff
@@ -101,19 +101,6 @@ impl Session {
...
-    /// KeyStore errors, such as IO, and if there is more than one strong reference
-    /// to the connection.
-    pub async fn close(&self) -> crate::mls::Result<()> {
-        self.crypto_provider
-            .close()
-            .await
-            .map_err(OpenMlsError::wrap("closing connection with keystore"))
-            .map_err(Into::into)
-    }
-
```
SimonThormeyer, line 116: It's great that we can remove this, but I'd appreciate if the message of the removing commit, e0cf43dd8a17dd49bddfdeaf81afe1462124b12c, provided more details about why this can now be safely removed.

crypto-ffi/bindings/shared/src/commonTest/kotlin/com/wire/crypto/GeneralTest.kt
```diff
@@ -91,27 +92,6 @@ class DatabaseKeyTest {
...
-
-        val newKey = genDatabaseKey()
-        migrateDatabaseKeyTypeToBytes(path.absolutePathString(), oldKey, newKey)
-        val db = openDatabase(path.absolutePathString(), newKey)
-
-        CoreCrypto(db)
-
-        tmpdir.toFile().deleteRecursively()
-    }
-
```
SimonThormeyer, line 114: We decided to keep the `migrateDatabaseKeyTypeToBytes()` function  IIRC? If so, we should also keep this test.

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCrypto.swift
```diff
@@ -225,8 +225,8 @@ extension Database {
     /// Initialise an in-memory Database whose data will be lost when the instance is dropped.
     ///
     /// - Parameter key: secret key to unlock the database
-    public static func open(key: DatabaseKey) async throws -> Database {
-        return try await inMemoryDatabase(key: key)
+    public static func open() async throws -> Database {
+        return try await inMemoryDatabase()
```
SimonThormeyer, line 240: Thanks for correctly marking this commit as breaking!
Technically, if we wanted to avoid having to release 11.x next, we should keep the old API though and add an item to the CC11 epic, to remove the unused argument?

keystore/tests/z_entities.rs
fewerner: Yes that's still correct. The next line starts a new transaction at the end of the test.

crypto/src/lib.rs
fewerner: I kept separate commits as it's separate crates. But have no preference here

keystore/src/mls.rs
coriolinus: Once I realized what the fix was, I could barely believe that nobody had removed it already!

crypto/src/lib.rs
coriolinus: ~~Let's put it on the rebase-queue along with removing reverted commits entirely.~~

Heh, hadn't even looked at the commits. IMO one commit per crate makes sense; I'd leave these two alone.

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCrypto.swift
coriolinus: Reverted the breaking commit: this function is now deprecated, and points to the zero-argument version, so that overall this PR doesn't become breaking.

crypto/src/mls/session/mod.rs
coriolinus: Ok, just also needs to happen on rebase.

crypto-ffi/bindings/shared/src/commonTest/kotlin/com/wire/crypto/GeneralTest.kt
coriolinus: Reversion is 0bd53b8fb7

SimonThormeyer (approved):
I am amazed that this actually works. This is the greatest improvement to the keystore crate this year, if not ever.

Great work @fewerner @coriolinus!

PR #2324, opened 2026-07-13
test(keystore/idb): use unique db name for each migration test

SimonThormeyer:
Also, clear the db after every test. Unfortunately, we can't do this automatically because we don't have async drop.


fewerner (approved):
Thanks! lgtm

PR #2330, opened 2026-07-15
Pq mls ciphersuites

emil-wire:
Adds the eleven post-quantum cipher suites from draft-ietf-mls-pq-ciphersuites-06, numbered `0xF001` to `0xF00B`. They go into core-crypto's own crypto provider and into the cipher suite enum the bindings use. Needs wireapp/openmls#110, pinned here by revision.

The provider now supports five post-quantum key exchanges from `hpke` 0.14.1:
- X-Wing (ML-KEM-768 with X25519)
- ML-KEM-768 with P-256
- ML-KEM-1024 with P-384
- ML-KEM-768 on its own
- ML-KEM-1024 on its own

It can also sign, verify and check keys for ML-DSA-44, 65 and 87, using `ml-dsa` 0.1.1.

The new suites are in `DEFAULT_SUPPORTED_CIPHERSUITES`, so every client advertises them in its leaf node.

The ML-DSA suites only work with basic credentials. Certificate enrolment returns `UnsupportedAlgorithm` for them and the JWK thumbprint comes back empty, because there is no JOSE key type for ML-DSA yet.

The numbers are placeholders in the private-use range of RFC 9420. IANA has assigned nothing.

Tests: `pq_suite_full_conversation_roundtrip` in `crypto/src/mls/conversation/mutable/encrypt.rs` creates a conversation with each of the eleven suites, adds a second member and sends messages both ways. The provider tests in `crypto/src/mls_provider` cover the rest.


crypto/src/mls/conversation/mutable/encrypt.rs
```diff
@@ -78,6 +78,58 @@ mod tests {
...
+            assert_ne!(&reply[..], &encrypted[..]);
+            let decrypted = conversation
+                .guard()
+                .await
+                .decrypt_message(encrypted)
+                .await
+                .unwrap()
+                .app_msg
+                .unwrap();
+            assert_eq!(&decrypted[..], &reply[..]);
```
SimonThormeyer: Everything after 
```rust
let conversation = case.create_conversation([&alice, &bob]).await;
```

could be replaced with
```rust
assert!(conversation.is_functional_and_contains([&alice, &bob]).await);
```

emil-wire: Thanks for the review, done :)

PR #2340, opened 2026-07-17
feat(swift): transactions run `withTaskCancellationHandler` [WPB-27117]

SimonThormeyer:
Make core crypto ffi transactions cancellable and run the ffi transaction `withTaskCancellationHandler` in the wrapper.

### DIFF
--- a/crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCrypto.swift
+++ b/crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCrypto.swift
@@ -63,8 +63,19 @@ public final class CoreCrypto: CoreCryptoProtocol, @unchecked Sendable {
         let filePath = dbLocation.map { FilePath(stringLiteral: $0) }
         let transactionExecutor = try TransactionExecutor<Result>(
             keystorePath: filePath, block)
+        let token = CoreCryptoCancellationToken()
+
         do {
-            try await coreCryptoFfi.transactionFfi(command: transactionExecutor)
+            try await withTaskCancellationHandler {
+                try await coreCryptoFfi.transactionFfiCancellable(
+                    command: transactionExecutor,
+                    cancellation: token
+                )
+            } onCancel: {
+                token.cancel()
+            }
+        } catch CoreCryptoError.TransactionCanceled {
+            throw CancellationError()
         } catch {
             throw await transactionExecutor.innerError ?? error
         }
--- a/crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
+++ b/crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
@@ -294,6 +294,96 @@ final class WireCoreCryptoTests: XCTestCase {
         )
     }
 
+    // swiftlint:disable:next function_body_length
+    func testCancellingTransactionStopsWaitingForLongRunningCallbacks() async throws {
+        // Verifies that cancellation stops waiting for PKI and MLS transport callbacks.
+        let database = try await newDatabase()
+
+        // This struct implements send commit bundle and x509 credential finalization by just waiting.
+        // Also, we're passing in expectations that are used as checkpoints.
+        let sendCommitStarted = expectation(description: "send commit started")
+        let sendCommitExited = expectation(description: "send commit exited")
+        let httpRequestStarted = expectation(description: "http request started")
+        let httpRequestExited = expectation(description: "http request exited")
+        let longRunningCallbacks = LongRunningCallbacks(
+            sencCommitStarted: sendCommitStarted,
+            sencCommitExited: sendCommitExited,
+            httpRequestStarted: httpRequestStarted,
+            httpRequestExited: httpRequestExited
+        )
+
+        let clientId = genClientId()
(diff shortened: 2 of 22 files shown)


crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
```diff
@@ -294,6 +294,96 @@ final class WireCoreCryptoTests: XCTestCase {
...
+            }
+        }
+
+        // Wait until both long-running callbacks have started before cancelling.
+        await fulfillment(of: [sendCommitStarted, httpRequestStarted])
+        transactionTask.cancel()
+
+        // Assert that cancellation is triggered without waiting for either callback result.
+        do {
+            try await transactionTask.value
```
typfel, line 370: We have a `XCTAssertThrowsErrorAsync` helper you can use here.

typfel (changes requested):

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
SimonThormeyer: When trying to use it like this

```swift
        let expectedError = CancellationError()
        // Assert that cancellation is triggered without waiting for either callback result.
        await XCTAssertThrowsErrorAsync(
            expectedError,
            when: {
                try await transactionTask.value
            }
        )
```
I get
`Instance method 'XCTAssertThrowsErrorAsync(_:when:_:file:line:)' requires that 'CancellationError' conform to 'Equatable'`

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
```diff
@@ -853,6 +943,66 @@ final class WireCoreCryptoTests: XCTestCase {
...
+        ) {
+            self.sencCommitStarted = sencCommitStarted
+            self.sendCommitExited = sencCommitExited
+            self.httpRequestStarted = httpRequestStarted
+            self.httpRequestExited = httpRequestExited
+        }
+
+        func sendCommitBundle(commitBundle: CommitBundle) async {
+            sencCommitStarted.fulfill()
+            try? await Task.sleep(for: .milliseconds(9999))
```
johnxnguyen: question: do I understand correctly that when the Task containing the transaction is cancelled, this cancellation proposes to this code here (which in the real case would be the `MLSTransport` on the Swift side)?

If that is correct, then the cancellation is picked up by `Task.sleep`, which explains why there's no explicit cancellation check. That's not a problem, just wanted to make it explicit.

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCrypto/CoreCrypto.swift
```diff
@@ -63,8 +63,19 @@ public final class CoreCrypto: CoreCryptoProtocol, @unchecked Sendable {
...
+            try await withTaskCancellationHandler {
+                try await coreCryptoFfi.transactionFfiCancellable(
+                    command: transactionExecutor,
+                    cancellation: token
+                )
+            } onCancel: {
+                token.cancel()
+            }
+        } catch CoreCryptoError.TransactionCanceled {
+            throw CancellationError()
```
johnxnguyen, line 78: praise: thanks for making bridging back to Swift

johnxnguyen (approved):
I only looked at the Swift code, and it looks good to me 👍🏼

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
typfel: Ah then it's not worth it.

typfel (approved):

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
samwyndham: I don't understand the rust 🐵 so looking at the tests to understand behavior. I assume that using some magic `try? await Task.sleep(for: .milliseconds(9999))` will receive the cancellation right?

Would it make sense to do a try catch to ensure that the Task.sleep is cancelled and not just expiring after 9999 milliseconds?

samwyndham (commented):
From what I understand it looks good!

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
SimonThormeyer: Yes, that's correct

SimonThormeyer: 4e7fca6b9ef9ec546925ae573ed310ef89c483e5

SimonThormeyer:
Closing in favor of #2342, which simplifies the rust side of things.

PR #2342, opened 2026-07-20
feat(swift): transactions run `withTaskCancellationHandler` [WPB-27117]

SimonThormeyer:
Make core crypto ffi transactions cancellable and run the ffi transaction `withTaskCancellationHandler` in the wrapper.

### DIFF
--- a/crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
+++ b/crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
@@ -294,6 +294,96 @@ final class WireCoreCryptoTests: XCTestCase {
         )
     }
 
+    // swiftlint:disable:next function_body_length
+    func testCancellingTransactionStopsWaitingForLongRunningCallbacks() async throws {
+        // Verifies that cancellation stops waiting for PKI and MLS transport callbacks.
+        let database = try await newDatabase()
+
+        // This struct implements send commit bundle and x509 credential finalization by just waiting.
+        // Also, we're passing in expectations that are used as checkpoints.
+        let sendCommitStarted = expectation(description: "send commit started")
+        let sendCommitExited = expectation(description: "send commit exited")
+        let httpRequestStarted = expectation(description: "http request started")
+        let httpRequestExited = expectation(description: "http request exited")
+        let longRunningCallbacks = LongRunningCallbacks(
+            sencCommitStarted: sendCommitStarted,
+            sencCommitExited: sendCommitExited,
+            httpRequestStarted: httpRequestStarted,
+            httpRequestExited: httpRequestExited
+        )
+
+        let clientId = genClientId()
+        let coreCrypto = try CoreCrypto(database: database)
+
+        let pkiEnvironment = try await PkiEnvironment(
+            hooks: longRunningCallbacks, database: database)
+        await coreCrypto.setPkiEnvironment(pkiEnvironment: pkiEnvironment)
+        let acquisition = try X509CredentialAcquisition(
+            pkiEnvironment: pkiEnvironment,
+            config: X509CredentialAcquisitionConfiguration(
+                acmeDirectoryUrl: "https://acme.example.com/directory",
+                cipherSuite: cipherSuiteDefault(),
+                displayName: "Alice Smith",
+                clientId: clientId,
+                handle: "alice_wire",
+                domain: "world.com",
+                team: nil,
+                validityPeriodSecs: 3600
+            )
+        )
+
+        let conversationId = genConversationId()
+        let credential = try Credential.basic(
+            cipherSuite: cipherSuiteDefault(),
+            clientId: clientId
+        )
+
+        try await coreCrypto.transaction { context in
+            try await context.mlsInit(clientId: clientId, transport: longRunningCallbacks)
+            let credentialRef = try await context.addCredential(credential: credential)
+            try await context.createConversation(
+                conversationId: conversationId,
+                credentialRef: credentialRef,
+                externalSender: nil
+            )
+        }
+
+        let transactionTask = Task {
+            try await coreCrypto.transaction { context in
+                try await context.setData(data: Data("This data should not be committed.".utf8))
+
+                // Start the long-running callbacks concurrently within the cancellable transaction.
+                async let update: () = context.updateKeyingMaterial(
+                    conversationId: conversationId
+                )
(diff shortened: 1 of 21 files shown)


crypto-ffi/src/cancellation/token.rs
```diff
@@ -0,0 +1,65 @@
+use std::sync::{Arc, Mutex};
+
+use futures_channel::oneshot;
+use futures_util::{FutureExt as _, future::Shared};
+
+/// Use this to cancel a `CoreCrypto` transaction and running foreign callbacks. Should be used in the Swift wrapper
+/// only.
+#[derive(Debug, uniffi::Object)]
+pub struct CoreCryptoCancellationToken {
+    inner: Arc<Inner>,
```
coriolinus: I'd love to see some inline docs about how this design was evolved. Right now `CCCancellationToken::new()` resolves to `Arc<CCCancellationToken{ inner: Arc<Inner>}>`, and it's not immediately obvious why we need both arcs and the separation between the two layers.

crypto-ffi/src/cancellation/slot.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub(crate) fn enter(
+        self: &Arc<Self>,
+        token: Arc<CoreCryptoCancellationToken>,
+    ) -> CoreCryptoResult<CancellationGuard> {
+        let mut current = self.current.lock().map_err(CoreCryptoError::ad_hoc)?;
+
+        assert!(
+            current.is_none(),
+            "only one transaction cancellation token may be in the slot; correct wrapper implementation never hits this"
+        );
```
coriolinus, line 34: If a correct wrapper implementation never hits this, could we replace this with `debug_assert!` and skip the check? Or is this sufficiently far out in the FFI layer that we only run tests in release mode?

crypto-ffi/src/cancellation/slot.rs
```diff
@@ -0,0 +1,57 @@
...
+/// Clears the active cancellation token from the slot when dropped.
+#[derive(Debug)]
+pub(crate) struct CancellationGuard {
+    slot: Arc<CancellationSlot>,
+}
+
+impl CancellationSlot {
+    /// Installs the token for the transaction currently holding the semaphore.
+    ///
+    /// Only one token may be installed at a time.
```
coriolinus: If we do retain the unconditional `assert!` line, then we should add to this doc line that it will panic if that constraint is violated.

crypto-ffi/src/core_crypto/mls_transport.rs
```diff
@@ -53,22 +60,56 @@ impl core_crypto::MlsTransport for MlsTransportShim {
...
+    else {
+        return callback.await;
+    };
+
+    // Futures used with select_biased!{} need to be `Fuse` instances. select_biased!{} prefers the first declared
+    // future in case both finish at the same time.
+    let result = futures_util::select_biased! {
+        result = callback.fuse() => result,
+        _ = token.cancelled().fuse() => Err(core_crypto::Error::ErrorDuringMlsTransport("cancelled via cancellation token".into())),
+    };
```
coriolinus, line 127: In other `select_biased!` invocations we've preferred cancellation over completion. Any particular reason the policy is different here?

coriolinus (approved):
I like these changes! Some nits, but ultimately this is quite good.

crypto-ffi/src/cancellation/token.rs
```diff
@@ -0,0 +1,65 @@
+use std::sync::{Arc, Mutex};
+
+use futures_channel::oneshot;
+use futures_util::{FutureExt as _, future::Shared};
+
+/// Use this to cancel a `CoreCrypto` transaction and running foreign callbacks. Should be used in the Swift wrapper
+/// only.
+#[derive(Debug, uniffi::Object)]
+pub struct CoreCryptoCancellationToken {
```
coriolinus, line 9: Just out of curiosity, how much of this API came directly from [`tokio_util::sync::CancellationToken`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html)?

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
```diff
@@ -853,6 +943,79 @@ final class WireCoreCryptoTests: XCTestCase {
         }
     }
 
+    private final actor LongRunningCallbacks: MlsTransport, PkiEnvironmentHooks {
```
fewerner, line 946: You could inherit from our dummy implementations here and only override the functions that you care about.

fewerner (approved):
GJ, nice solution! Just one additional comment.

crypto-ffi/src/cancellation/token.rs
SimonThormeyer: b22710dc83da6d495916f1c8f9472b9ebc371885

crypto-ffi/src/cancellation/slot.rs
SimonThormeyer: Correct, we only test the binary in release mode.

crypto-ffi/src/cancellation/token.rs
SimonThormeyer: Decided to add a ticket to consider replacing the local impl with the library

crypto-ffi/src/cancellation/slot.rs
SimonThormeyer: 8992bf394ba9d614190c3ca511dbc6de888d5873

crypto-ffi/src/core_crypto/mls_transport.rs
SimonThormeyer: No, good catch! 5a7c2913efd2acc2baabb870fefa1682f768a308

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
SimonThormeyer: I tried it, unfortunately that doesn't work because we have one type that implements both interfaces, and there's no multiple inheritance in swift.

Anyway I have the following counter argument against doing this:
I see that one would like to do this to avoid some boilerplate (which isn't that large in this case though). Semantically, I also don't directly see why `LongRunningCallbacks` would share behavior with anything that doesn't have a long running callback (which is the case for our mock impls). So we would effectively override anything relevant.

PR #2343, opened 2026-07-20
test(ts): unify ts benchmarks [WPB-27138]

fewerner:
This PR factors out common logic of ts-native and ts-browser benchmarks.


crypto-ffi/bindings/js/packages/native/benches/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+        globalThis.tinybench = await import("tinybench");
+    }
+}
+
+export async function runBenchmark(benchmarkSetup: () => Promise<void>) {
+    await setup();
+    await benchmarkSetup();
+    console.log(`Starting ${bench.name}`);
+    await bench.run();
+    await logResults(bench.name, bench.table());
```
SimonThormeyer, line 23: what is `bench`? I can't see where this variable is assigned a value.

EDIT: I can see that it is declared in `global` in a later commit. Probably this hunk would be less confusing if it was folded into that?

SimonThormeyer (approved):
One nit regarding history, other than that, it looks good to me!

crypto-ffi/bindings/js/packages/native/benches/utils.ts
fewerner: Ah sorry, yep that's indeed confusing. It is the global delcaration

PR #2344, opened 2026-07-20
test: fix mixed cipher suite test

SimonThormeyer:
The previous version of the test was incorrectly combining cipher suites and sign algorithms. This fixed version does everything in a simpler way: we create the key package with the other cipher suite from a fresh `TestContext`, so it's easier to get it right.


fewerner (approved):
lgtm!

PR #2353, opened 2026-07-21
ci: ensure release tag matches package versions

SimonThormeyer:
see above

### DIFF
--- a/.github/workflows/prepare-publish.yml
+++ b/.github/workflows/prepare-publish.yml
@@ -50,6 +50,39 @@ jobs:
             exit 1
           fi
 
+      - name: Ensure the tag name matches all package versions to publish
+        run: |
+          TAG_NAME=${GITHUB_REF#refs/tags/}
+          EXPECTED_VERSION=${TAG_NAME#v}
+
+          if [[ "$TAG_NAME" != v* || -z "$EXPECTED_VERSION" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/prepare-publish.yml
```diff
@@ -50,6 +50,35 @@ jobs:
...
+          if [[ "$GRADLE_VERSION" != "$EXPECTED_VERSION" ]]; then
+            echo "Gradle packages version '$GRADLE_VERSION' does not match release tag '$TAG_NAME'"
+            exit 1
+          fi
+
+          if [[ "$NPM_VERSION" != "$EXPECTED_VERSION" ]]; then
+            echo "npm package @wireapp/core-crypto version '$NPM_VERSION' does not match release tag '$TAG_NAME'"
+            exit 1
+          fi
+
```
coriolinus, line 83: Might be better not to fail fast here: 

```bash
exitcode=0
if [[ "$CARGO_VERSION != "$EXPECTED_VERSION" ]]; then
  echo "cargo version mismatch";
  exitcode=1
fi
# other two checks go here
exit "$exitcode"
```

Realistically we normally expect either all to pass or none of them, but just in case somehow they get out of sync, it's nicer to get all the feedback at once.

coriolinus (approved):
I have a suggestion here but it's not essential.

PR #2356, opened 2026-07-21
docs: add docs describing how `CoreCryptoCancellationToken` works

SimonThormeyer:
see above

### DIFF
--- a/crypto-ffi/src/cancellation/mod.rs
+++ b/crypto-ffi/src/cancellation/mod.rs
@@ -1,3 +1,35 @@
+//! # CoreCryptoCancellationToken Implementation Details
+//!
+//! For a more high-level description and a sequence diagram, see the `Transaction` chapter of the CoreCrypto book.
+//!
+//! The [CoreCryptoCancellationToken] is supposed, but in theory not limited, to be used in
+//! [crate::CoreCryptoFfi::transaction_ffi_cancellable]. This describes the implementation and intended usage which
+//! matches the current usage in that function.
+//!
+//! ## How Cancellation Works
+//! To make a future cancellable, we race the future against [CoreCryptoCancellationToken::cancelled] using
+//! [futures_util::select_biased], with a bias for [CoreCryptoCancellationToken::cancelled] - in case both futures
+//! complete at the same time, cancellation is preferred. If cancellation wins, we return an appropriate error.
+//!
+//! ## Sharing the Cancellation Token With Foreign Callbacks
+//! Cancellation needs to cause rust futures currently waiting for foreign callbacks to stop waiting, i.e., be dropped.
+//! This is because when they are dropped, the uniffi-generated swift code will cancel the task that is running the
+//! foreign callback. To share an existing token with callbacks, [crate::CoreCryptoFfi] has a clonable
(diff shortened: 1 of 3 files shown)


crypto-ffi/src/cancellation/mod.rs
```diff
@@ -1,3 +1,33 @@
...
+//! Cancellation needs to cause rust futures currently waiting for foreign callbacks to stop waiting, i.e., be dropped.
+//! This is because when they are dropped, the uniffi-generated swift code will cancel the task that is running the
+//! foreign callback. To share an existing token with callbacks, [crate::CoreCryptoFfi] has a clonable
+//! [CancellationSlot], with space for a single token.
+//! To fill a slot, call [CancellationSlot::enter]. This will return a [slot::CancellationGuard] which empties the slot
+//! when dropped. [CancellationSlot::enter] will panic if called while the slot is still filled.
+//! [crate::CoreCryptoFfi::transaction_ffi_cancellable] ensures that this constraint isn't violated by only filling the
+//! slot after the transaction semaphore was acquired. *This constraint makes the [CancellationSlot] currently only
+//! usable in transactions*.
+//! The slot is cloned to this crate's structs implementing foreign traits, currently [crate::MlsTransport] and
```
istankovic: What does it mean to "clone the slot to this crate's structs"? Did you mean "add" instead of "clone"?

crypto-ffi/src/cancellation/mod.rs
```diff
@@ -1,3 +1,33 @@
...
+//!
+//! The [CoreCryptoCancellationToken] is supposed, but in theory not limited, to be used in
+//! [crate::CoreCryptoFfi::transaction_ffi_cancellable]. This describes the implementation and intended usage which
+//! matches the current usage in that function.
+//!
+//! ## How Cancellation Works
+//! To make a future cancellable, we race the future against [CoreCryptoCancellationToken::cancelled] using
+//! [futures_util::select_biased], with a bias for [CoreCryptoCancellationToken::cancelled] - in case both futures
+//! complete at the same time, cancellation is preferred. If cancellation wins, we return an appropriate error.
+//!
```
istankovic, line 13: This already helps quite a bit, but I think we can go one step further and really make things clear by outlining the sequence of steps that occur, both on the Rust side and the foreign language side.
IOW, something like a sequence diagram would be helpful to show when tokens are created, when the slots are filled, when the guards are dropped etc.

crypto-ffi/src/core_crypto/command/mod.rs
```diff
@@ -94,6 +94,8 @@ impl CoreCryptoFfi {
     ///
     /// Cancelling the token aborts the transaction and stops waiting for any
     /// in-flight `MlsTransport` callback or `PkiEnvironment` hook associated with it.
+    ///
+    /// For implementation details, see the rust module docs of this crate's `cancellation` module.
```
coriolinus: We could get an internal link here, i.e.

```rust
/// For implementation details, see the [`cancellation` docs][crate::cancellation].
```

coriolinus (approved):
Nice work with the diagram!

PR #2365, opened 2026-07-23
[DO NOT MERGE] feat: automatically recover from `ConversationAlreadyExists` caused by epoch-0 groups

SimonThormeyer:
See above.

Don't merge, because we haven't decided yet that this is the path we'd like to go.

### DIFF
--- a/crypto/src/transaction_context/conversation/persistence.rs
+++ b/crypto/src/transaction_context/conversation/persistence.rs
@@ -86,8 +86,28 @@ impl TransactionContext {
 
         let id = ConversationId::from(group.group_id().as_slice());
 
-        if self.conversation_exists(&id).await? || self.pending_conversation_exists(&id).await? {
-            return Err(LeafError::ConversationAlreadyExists(id).into());
+        // If the conversation exists locally as a pending conversation or as a conversation with epoch 0, we can clear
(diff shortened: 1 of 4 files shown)


coriolinus (approved):
I like this approach, thanks for writing it!

(Approval obviously also depends on fixing CI though.)

crypto/src/transaction_context/conversation/persistence.rs
```diff
@@ -87,8 +87,28 @@ impl TransactionContext {
 
         let id = ConversationId::from(group.group_id().as_slice());
 
-        if self.conversation_exists(&id).await? || self.pending_conversation_exists(&id).await? {
-            return Err(LeafError::ConversationAlreadyExists(id).into());
+        // If the conversation exists locally as a pending conversation or as a conversation with epoch 0, we can clear
+        // it and keep the one we just created, because this will lead to the desired state, compatible with what we had
+        // locally. A fresh group with this `id`, just not created by us.
+        // This resolves WPB-21116.
```
istankovic, line 92: It is not clear to me that this is the best solution for WPB-21116, in part because it is unclear how exactly the problem arises. So I would prefer to not proceed with this until we get more clarity on the original issue.

SimonThormeyer:
The rollback mechanism is sufficient for clients to recover from this. So no need to do this implicitly.

PR #2366, opened 2026-07-23
feat(conversation): process welcome: retain key package on error [WPB-27413]

SimonThormeyer:
This is needed for correctness, see corresponding item.

Unfortunately it's a bit more complex than expected, and we need a condition to hold for this to work: the keypackage must have been created in an earlier transaction. It's reasonable to assume that, though, imo

### DIFF
--- a/crypto/src/transaction_context/conversation/welcome.rs
+++ b/crypto/src/transaction_context/conversation/welcome.rs
@@ -31,10 +31,21 @@ impl TransactionContext {
             ..Default::default()
         };
 
-        let conversation = self
-            .persist_conversation_from_welcome_message(welcome, configuration)
-            .await?;
+        let welcome_clone = welcome.clone();
+        let conversation_result = self
+            .persist_conversation_from_welcome_message(welcome_clone, configuration)
+            .await;
 
+        if conversation_result.is_err() {
+            // If persisting failed, we want to pretend we didn't process the welcome at all and restore the key
+            // package that may have been marked for deletion by openmls:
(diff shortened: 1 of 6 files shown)


crypto/src/transaction_context/conversation/welcome.rs
```diff
@@ -96,11 +96,26 @@ mod tests {
...
                     .unwrap();
 
+                // Bob's key package count before processing the welcome
+                let key_package_refs_before = bob.transaction.get_key_package_refs().await.unwrap();
+
                 let welcome = conversation.transport().await.latest_welcome_message().await;
+
+                // We need Bob's created key package to be persisted, so we can restore it on error.
+                // This matches real-world usage: the key package must have been created in an earlier transaction
+                // before the welcome is processed, because otherwise Bob couldn't have been invited.
```
coriolinus: We need it to be persisted, sure. And I'll buy that it matches real-world usage. But I'm not sure that it would be impossible for a client to start one megatransaction, create a keypackage, send it to the server for broadcast, and then get a welcome based on that keypackage while still in the megatransaction.

Good idea? Absolutely not. But I bet we could write a test case which demonstrated that failure mode, if we wanted to. It's weird and unlikely enough that I wouldn't bother, though. I'd just adjust the comment here.

coriolinus (approved):
I like how you cleared four of the nits in the pending review with fixups before I even had the chance to post the review!

One more non-blocking nit. But good work!

PR #2367, opened 2026-07-23
chore: replace eslint and prettier with oxlint and oxfmt

fewerner:
oxc is a dropin replacement for eslint and prettier.


coriolinus (approved):

crypto-ffi/bindings/js/.oxfmtrc.json
```diff
@@ -0,0 +1,19 @@
+{
+    "$schema": "./node_modules/oxfmt/configuration_schema.json",
+    "semi": true,
+    "singleQuote": false,
+    "tabWidth": 4,
+    "trailingComma": "es5",
+    "arrowParens": "always",
+    "printWidth": 80,
+    "sortPackageJson": false,
+    "ignorePatterns": ["node_modules", "rust_modules", "out"],
```
SimonThormeyer: Do we even have an `out` folder still?

crypto-ffi/bindings/js/package.json
```diff
@@ -59,5 +58,8 @@
     "license": "GPL-3.0",
     "trustedDependencies": [
         "uniffi-bindgen-react-native"
-    ]
+    ],
+    "dependencies": {
+        "oxlint": "^1.75.0"
+    }
```
SimonThormeyer: Why isn't this in `devDependencies`?

crypto-ffi/bindings/js/.oxfmtrc.json
fewerner: nope, good catch!

crypto-ffi/bindings/js/package.json
fewerner: that's also a very good catch!

SimonThormeyer (approved):
Two comments, otherwise good to go

crypto-ffi/bindings/js/.oxfmtrc.json
```diff
@@ -0,0 +1,19 @@
+{
+    "$schema": "./node_modules/oxfmt/configuration_schema.json",
+    "semi": true,
+    "singleQuote": false,
+    "tabWidth": 4,
+    "trailingComma": "es5",
```
SimonThormeyer, line 6: Would it be possible to allow, or even enforce, trailing comma? That would be amazing

PR #2371, opened 2026-07-24
chore(ts): update to ts 7

fewerner:
This PR upgrades to ts 7.
We get [type-aware](https://oxc.rs/docs/guide/usage/linter/type-aware) lints with that.

TypeDoc doesn't support ts7, yet. We use a separate package with ts 6 to create docs for the time being.


crypto-ffi/bindings/js/package.json
```diff
@@ -52,11 +52,15 @@
...
-        "typescript": "^6.0.3",
+        "typescript": "^7.0.2",
         "typescript-eslint": "^8.63.0",
+        "typescript6": "npm:typescript@6",
         "uniffi-bindgen-react-native": "^0.31.0-3",
         "yarn": "^1.22.22"
     },
+    "overrides": {
+        "typedoc/typescript": "6.x"
+    },
```
SimonThormeyer: Mentioning in the commit message why typedoc needs this version, perhaps even linking the open issue would be really useful, because we can't comment in `json` files.

SimonThormeyer (approved):
Just one comment. Thanks for doing this!

crypto-ffi/bindings/js/package.json
fewerner: That is actually a mistake. We don't need this as we define the separate typedoc package

fewerner: The actual commit for that has a comment, I'll link to the gh issue as well there

PR #2375, opened 2026-07-27
ci: fix injection vulnerabilities [WPB-27207]

fewerner:
See
https://docs.github.com/en/actions/reference/security/secure-use#good-practices-for-mitigating-script-injection-attacks

### DIFF
--- a/.github/actions/make/action.yml
+++ b/.github/actions/make/action.yml
@@ -110,14 +110,17 @@ runs:
         tar -xzf artifact.tar.gz
         # This is necessary because the github checkout action doesn't preserve
         # timestamps on checked out files.
-        paths="${{ inputs.target-path }}"
-        for dir in $paths; do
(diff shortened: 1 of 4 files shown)


.github/actions/make/action.yml
```diff
@@ -110,14 +110,17 @@ runs:
...
+
           if [ -d "$dir" ]; then
             find "$dir" -type f -exec touch {} +
           elif [ -f "$dir" ]; then
             touch "$dir"
           fi
-        done
+        done <<< "$TARGET_PATH"
+      env:
+        TARGET_PATH: ${{ inputs.target-path }}
```
fewerner, line 123: Had to change the iteration here so that `$TARGET_PATH` isn't treated as a single argument when it provides multiple artifacts.

coriolinus (approved):

PR #2376, opened 2026-07-27
refactor!: `DecryptedMessage` is an enum

SimonThormeyer:
This changes the `DecryptedMessage` struct to an enum in both ffi crate and crypto crate.


interop/src/clients/corecrypto/web/mls.ts
```diff
@@ -113,13 +113,16 @@ export async function encryptMessage() {
...
     const conversationId = new ConversationId(Uint8Array.from(Object.values(cId)));
     const encryptedMessage = Uint8Array.from(Object.values(encMessage));
 
-    const { message } = await window.cc.transaction((ctx) =>
+    const decryptedMessage = await window.cc.transaction((ctx) =>
         ctx.decryptMessage(conversationId, encryptedMessage)
     );
-    return new Uint8Array(message);
+    if (DecryptedMessage.Text.instanceOf(decryptedMessage)) {
+        return new Uint8Array(decryptedMessage.inner.plaintext);
```
fewerner: plaintext already is a `Uint8Array`

fewerner (approved):
Nice improvement, I like it!

crypto/src/mls/conversation/mutable/decrypt/mod.rs
```diff
@@ -33,52 +33,161 @@ use crate::{
...
+    pub fn as_text(&self) -> Option<&Text> {
+        let Self::Text(message) = self else {
+            return None;
+        };
+
+        Some(message)
+    }
+
+    /// Relevant data after processing if the decrypted message was an MLS application message.
+    pub fn into_text(self) -> Option<Text> {
```
coriolinus: We can generate this (and related) methods with [`derive_more::TryInto`](https://docs.rs/derive_more/latest/derive_more/derive.TryInto.html#enums).

crypto/src/mls/conversation/mutable/decrypt/mod.rs
```diff
@@ -33,52 +33,161 @@ use crate::{
...
+    pub fn identity(&self) -> &WireIdentity {
+        match self {
+            DecryptedMessage::Text(text) => &text.identity,
+            DecryptedMessage::Commit(commit) => &commit.identity,
+            DecryptedMessage::Proposal(proposal) => &proposal.identity,
+        }
+    }
+
+    /// Relevant data after processing if the decrypted message was an MLS application message.
+    pub fn as_text(&self) -> Option<&Text> {
```
coriolinus: Alternately, we can generate this (and related) methods with [`enum_as_inner::EnumAsInner`](https://docs.rs/enum-as-inner/latest/enum_as_inner/).

crypto/src/mls/conversation/mutable/decrypt/mod.rs
```diff
@@ -300,31 +304,28 @@ impl ConversationMut {
...
             .identity()
             .try_into()
-            .inspect_err(|e| log::info!("sender client id couldn't be parsed into the expected format: {e:?}"))
-            .ok();
+            .map_err(RecursiveError::mls_client("client id from credential"));
 
         let decrypted = match message.into_content() {
             ProcessedMessageContent::ApplicationMessage(app_msg) => {
                 let conversation = self;
+                let sender_client_id = sender_client_id_result?;
```
coriolinus, line 264: Only propagate `sender_client_id_result`'s error in the event it was an application message, right? Any reason not to move its parsing here?

coriolinus (approved):
That's quite good!

crypto/src/mls/conversation/mutable/decrypt/mod.rs
SimonThormeyer: Not sure if parsing the message above is strictly necessary, however, `message.into_content()` (3 lines above) moves `message`, and producing `sender_client_id_result` requires borrowing it (via `message.credential()`), so we have to do it before if we're not working around that in another way.

SimonThormeyer: I added a comment in the place we produce `sender_client_id_result`.

PR #2406, opened 2026-07-29
ci(docs): replace docs within the same folder

SimonThormeyer:
`keep_files: true` was thought to be the best solution to not interfere
with other folders. However, `destination_dir` is a much better fit.

Not using `keep_files: true` anymore, solves the problems of
ever-increasing directory sizes and incorrect results in case of struct
renames, where the old struct doc would be kept.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -35,7 +35,7 @@ jobs:
 
     - name: deploy static files
       if: github.ref == 'refs/heads/main'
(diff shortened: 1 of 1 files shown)


coriolinus (approved):
I'm worried that we're going to break all our links if we do this. I just don't know how `destination_docs` works yet. However, it seems clear that deploying the versioned docs with `keep_files: false` is safe, because we do in fact set `destination_dir`, preventing clobbering.

I'd say go for it, we can fix the links later if we discover they are broken.

PR #2410, opened 2026-07-29
refactor: decouple transaction cache from database [WPB-23951]

coriolinus:
- [x] rename `KeystoreTransaction` -> `Transaction`
- [x] keystore transaction keeps a reference to the database
- [x] database keeps a weak reference to the active transaction
  - [x] getters and guards
- [x] transactions are passed explicitly through the keystore
- [x] mutating methods move from `Database` to `Transaction`
- [x] transactions impl `FetchFromDatabase`
- [x] transactions impl `OpenMlsKeyStore`
- [x] transactions impl `PreKeyStore`
- [x] transactions playback upsert events in order
- [x] simplify CC `TransactionContextInner`
  - [x] getters and guards
- [ ] ~~database no longer impls `OpenMlsKeyStore`~~ unfortunately, this breaks things, because we sometimes have to hand over an `OpenMlsCryptoProvider`, and of that, [`KeyStoreProvider: OpenMlsKeyStore`](https://github.com/wireapp/openmls/blob/c9cde17076508968c9cbead5728454f0a1f60c4f/traits/src/traits.rs#L20). OpenMLS's bad API design strikes again!
  But seriously, I can't wait to use `mls-rs` and not tie all these random traits together all the time.
- [x] database no longer impls `PreKeyStore
- [x] adjust everything else so it works


keystore/tests/searchable_entities.rs
```diff
@@ -55,9 +55,9 @@ mod persisted_mls_groups {
         let entity = random_entity();
         let search_key = get_search_key(&entity);
 
+        let tx = store.new_transaction().await.unwrap();
         store.save(entity.clone()).await.unwrap();
-        store.commit_transaction().await.unwrap();
-        store.new_transaction().await.unwrap();
+        tx.commit().await.unwrap();
```
fewerner, line 60: In d945ef6 I asked myself why the mutating methods still live on the database, and you adjust that in the next two commits.
I think you could fold d945ef6 into 1a2cdbf. It would give a better picture on how the two changes interact.

keystore/src/mls.rs
```diff
@@ -3,15 +3,33 @@ use openmls_basic_credential::SignatureKeyPair;
...
+    }
+
+    async fn remove_borrowed<E>(&self, id: &E::BorrowedPrimaryKey) -> CryptoKeystoreResult<()>
+    where
+        E: Entity + EntityDatabaseMutation + BorrowPrimaryKey + EntityDeleteBorrowed,
+    {
+        self.with_transaction(async |tx| tx.remove_borrowed::<E>(id).await)
+            .await
+    }
+}
```
fewerner, line 30: Main purpose for these functions to still exist on database is the implementation of the openmls Keystore trait, right?

I assume this is also the reason why we can't get rid of the `MutatingOperationWithoutTransaction` Error? Ideally, we should not need this error.

fewerner (commented):
I have to say, the way this PR is structured and the quality of documentation and commit messages is so helpful to understand the changes.
I really appreciate this!

I have only a few comments.

keystore/src/transaction/specializations.rs
```diff
@@ -39,16 +39,15 @@ impl KeystoreTransaction {
...
+        {
+            let mut cache_guard = self.cache.write().await;
+            cache_guard.retain(|_entity_id, entity| {
+                let Some(pending_message) = entity.downcast::<MlsPendingMessage>() else {
+                    return true;
+                };
                 pending_message.foreign_id != conversation_id
             });
         }
-        drop(cache_guard);
```
SimonThormeyer: In many other places we've preferred using explicit `drop()` over introducing a block and indentation. 
Is there any specific reason why we don't want to do that here?

I personally like avoiding indentation, but I don't feel strongly about it.

keystore/src/connection/entity_extension_methods.rs
```diff
@@ -37,9 +37,10 @@ impl Database {
...
         self.with_transaction(async |transaction| {
-            Ok(transaction
+            transaction
                 .remove_pending_messages_by_conversation_id(conversation_id)
-                .await)
+                .await;
+            Ok(())
         })
         .await
     }
```
SimonThormeyer: Could this commit 28c6166b583e083afe41874aaf5580d78b4e422f easily be folded into the original change?

crypto/src/transaction_context/mod.rs
```diff
@@ -30,21 +31,21 @@ pub mod test_utils;
...
-        core_crypto: Arc<CoreCrypto>,
-        pending_epoch_changes: Arc<Mutex<Vec<(ConversationId, u64)>>>,
-    },
-    Invalid,
+#[derive(derive_more::Debug)]
+struct TransactionContextInner {
+    core_crypto: Arc<CoreCrypto>,
+    pending_epoch_changes: Arc<Mutex<Vec<(ConversationId, u64)>>>,
+    #[debug(skip)]
+    transaction: UniqueArc<core_crypto_keystore::Transaction>,
```
SimonThormeyer, line 48: This is much better! And makes me wonder why we didn't switch away from an enum earlier.

keystore/src/connection/entity_extension_methods.rs
fewerner: I thought the same when going over the separate formatting commit. What I usually do if I notice I committed with wrong formatting/linting before is using a macro to get the last commit hash for that change and ammending to that commit.

keystore/src/transaction/entity_read.rs
```diff
@@ -3,14 +3,13 @@
...
     CryptoKeystoreResult,
     traits::{BorrowPrimaryKey, Entity, KeyType, SearchableEntity},
     transaction::Transaction,
 };
 
-use super::dynamic_dispatch::EntityId;
-
 impl Transaction {
     async fn find_in_cache<E>(&self, entity_id: &EntityId) -> Option<Arc<E>>
     where
```
SimonThormeyer, line 15: Would it, with reasonable effort, be possible to fold this formatting commit into the original changes?

SimonThormeyer (approved):
Great work! Just some nits, mostly related to history

keystore/src/transaction/specializations.rs
coriolinus: Hm, I didn't realize we had that policy! As you can see my instinct runs in the opposite direction: adding an explicit block is faster to read IMO than an inline `drop` statement, and encourages us to keep the block small. 

I'd prefer to keep the block structure, in this case at least.

keystore/src/mls.rs
coriolinus: Yes, you're right on both counts. I was really hoping to be able to get rid of the error variant at least, but the design of OpenMLS's trait hierarchy doesn't let us, unfortunately.

keystore/tests/searchable_entities.rs
coriolinus: Reordered and squashed those.

keystore/src/connection/entity_extension_methods.rs
coriolinus: Took some rebasing but I got those merged into the history in the right places.

keystore/src/transaction/entity_read.rs
coriolinus: Yes, accomplished that.

PR #2411, opened 2026-07-30
test(ts): replace wdio with puppeteer

fewerner:
This PR simplifies running browser code by switching from wdio to puppeteer.
Benchmarks can now be long-running and awaited.
We reduce test execution time by ~50%.


crypto-ffi/bindings/js/package.json
```diff
@@ -37,6 +37,7 @@
         "@wireapp/core-crypto": "workspace:*",
         "chai": "^6.2.2",
         "dts-bundle-generator": "^9.5.1",
+        "mocha": "^11.7.6",
```
SimonThormeyer, line 45: This commit 7c16bd60af9c6b35903e7b064ac86974ef13b3fb message has a typo: We no install -> we now install

crypto-ffi/bindings/js/packages/browser/benches/utils.ts
```diff
@@ -1,20 +1,6 @@
-import {
-    logResults,
-    setup as sharedSetup,
-} from "../../../shared/benches/utils";
+import { setup as sharedSetup } from "../../../shared/benches/utils";
```
SimonThormeyer, line 1: Would it be worth to introduce a named import here, too, to make it stable across file movements (here and in other places)? Up to you if you'd like to defer this potential refactoring to another time.

.github/workflows/benchmark-ts-browser.yml
```diff
@@ -81,17 +81,14 @@ jobs:
...
+          bun x puppeteer browsers install chrome-headless-shell
 
-      - name: Install chromedriver
-        if: ${{ runner.environment == 'self-hosted' }}
+      # See https://chromium.googlesource.com/chromium/src/+/main/docs/security/apparmor-userns-restrictions.md
+      - name: disable Ubuntu's security feature globally
         run: |
-          bun x @puppeteer/browsers install chromedriver@latest --path $PWD
-          echo "CHROMEDRIVER_PATH=$(echo $PWD/chromedriver/*/*/chromedriver)" >> $GITHUB_ENV
+          echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns
```
SimonThormeyer, line 93: We should log a link to the markdown document in the workflow logs, otherwise seeing the step name without any other context is going to raise eyebrows

.github/workflows/test-ts-browser.yml
```diff
@@ -32,6 +32,15 @@ jobs:
...
           artifact-generation: ${{ vars.ARTIFACT_GENERATION }}
 
+      - name: install chrome via puppeteer
+        run: |
+          bun x puppeteer browsers install chrome-headless-shell
+
+      # See https://chromium.googlesource.com/chromium/src/+/main/docs/security/apparmor-userns-restrictions.md
+      - name: disable Ubuntu's security feature globally
+        run: |
+          echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns
```
SimonThormeyer, line 44: Same here

SimonThormeyer (approved):
Two nits, one optional and one mandatory (the latter), but nothing fundamental.

Thanks for this huge improvement!

crypto-ffi/bindings/js/packages/browser/benches/utils.ts
fewerner: I'll leave that for later. I'm not sure what to do with the different utils. We are at a point where the whole file structure got reduced to almost only shared files for tests and benches. We could probably reduce the complexity of the folder structure as a next refactoring.

.github/workflows/benchmark-ts-browser.yml
felannan: Investigated first: confirmed 35 spec files (not 29 — a few more landed on main since this was written), 28 covered by e2e-nightly.yml's two jobs, and the same 7 orphans (agents, auth, dm-room-selection, harness, navigation, rooms, serverless) running in neither workflow. Traced their origin — agents/auth/navigation/rooms all go back to 8317869d, predating e2e-nightly.yml's own introduction; dm-room-selection landed incidentally in an unrelated refactor commit; harness/serverless are each intentional but never got a workflow step.

Ran all seven for real against this branch (not just read) rather than guessing from content — every one passed, every one is a single per-test launch (dm-room-selection launches no Electron at all — it's a plain function test), none touches a vendor API key or the network. Folded all seven into both e2e-pr-fast.yml and e2e-nightly.yml's e2e-smoke job, so every one of the 35 spec files now runs somewhere in CI (commit 67f4c86c). Along the way, fixed auth.spec.ts/harness.spec.ts's stale comments about E2E_SKIP_SERVERLESS (RIP-274 made serverless.spec.ts hermetic too, so that flag is now a dead manual escape hatch, not a real behavioral difference).

The header now carries an explicit coverage-accounting section with the current totals rather than a static list, so this doesn't go stale silently again. CI green on the current head (34f52564): https://github.com/felannan/carmen/actions/runs/30665279284

PR #2414, opened 2026-07-31
feat(conversation): sender nonce

SimonThormeyer:
Further preparation for T&T messages.


crypto-macros/src/entity_derive/column_type.rs
```diff
@@ -112,11 +121,13 @@ impl TryFrom<Type> for ColumnType {
...
         } else if bytes_types().iter().flat_map(optional_types).any(|o_type| o_type == ty) {
             Ok(Self::OptionalBytes)
+        } else if u32_types().contains(&ty) {
+            Ok(Self::U32)
         } else {
             let type_string = ty.to_token_stream().to_string();
             Err(syn::Error::new_spanned(
                 ty,
-                format!("Expected `String`, `Vec<u8>`, or `Option<Vec<u8>>`, not `{type_string}`"),
+                format!("Expected `String`, `Vec<u8>`, `Option<Vec<u8>>`, or `u32, not `{type_string}`"),
```
fewerner: closing  back tick is missing here on `u32`

crypto/src/mls/conversation/mutable/group_mutation.rs
```diff
@@ -35,18 +35,25 @@ impl ConversationMut {
...
+                state: core_crypto_keystore::ser(group.mls_group())
+                    .map_err(KeystoreError::wrap("serializing group state"))?,
                 parent_id: None,
+                sender_nonce: group.sender_nonce(),
             })
             .await
             .map_err(KeystoreError::wrap("persisting mls group"))?;
 
-            group.set_state(InnerState::Persisted);
+            group.mls_group_mut().set_state(InnerState::Persisted);
```
fewerner: We could do some inversion here and reduce nesting:
```
if group.state_changed() != InnerState::Changed {
        return Ok(result);
}
```

But not sure if it's worth it. I usually prefer less nesting.

crypto/src/transaction_context/conversation/mod.rs
```diff
@@ -106,7 +106,8 @@ impl TransactionContext {
         .await
         .map_err(OpenMlsError::wrap("creating group with id"))?;
 
-        self.persist_conversation_from_mls_group(group, configuration).await?;
+        self.persist_conversation_from_mls_group(group, configuration, u32::MIN)
```
fewerner: These and other `u32::MIN` are just placeholders , right?

fewerner (commented):
Thanks! lgtm. Only a few comments.

crypto/src/transaction_context/conversation/mod.rs
SimonThormeyer: Exactly, not just using `0` was supposed to remind me that I'm going to pass `SenderNonce::default()` here; Though now I realize `Default::default()` would have worked now and later. Going to put that there now.

fewerner (approved):

PR #2417, opened 2026-07-31
chore(keystore): drop `e2ei_enrollment` table [WPB-24480]

coriolinus:
We're not using that table at all anymore; it's been superseded by the new E2EI implementation from CC 10.

Proof: the only changes required in `crypto` to make this PR pass all tests were to remove the entity from some test functions; nothing in live code.


fewerner (approved):
lgtm!

PR #2419, opened 2026-07-31
chore: misc keystore cleanups [WPB-23776]

coriolinus:
Some cleanups of the keystore which weren't worth breaking out into separate tickets.


fewerner (approved):
Thanks, lgtm!

PR #2421, opened 2026-07-31
feat: add `encrypt_targeted_message()` [WPB-27214]

SimonThormeyer:
This is further partial work for WPB-27214.

Remaining after this
- nonce handling on the receiver side
- `decrypt_targeted_message()`
- encrypt/decrypt tests

### DIFF
--- a/crypto-ffi/src/core_crypto_context/mls.rs
+++ b/crypto-ffi/src/core_crypto_context/mls.rs
@@ -62,6 +62,24 @@ bytes_wrapper!(
     Welcome fallibly wraps core_crypto::mls::conversation::WelcomeMessage
 );
 
+/// The policy to encrypt the targeted message with.
+#[derive(Debug, Clone, Copy, uniffi::Enum)]
+pub enum TargetedMessagePolicy {
+    /// Won't be persisted and will only be visible to currently online members who immediately process it.
+    Transient,
+    /// May be persisted and buffered, will also be delivered and processed by currently offline members.
+    Persisted,
+}
+
+impl From<TargetedMessagePolicy> for core_crypto::mls::conversation::TargetedMessagePolicy {
+    fn from(value: TargetedMessagePolicy) -> Self {
+        match value {
+            TargetedMessagePolicy::Transient => core_crypto::mls::conversation::TargetedMessagePolicy::Transient,
+            TargetedMessagePolicy::Persisted => core_crypto::mls::conversation::TargetedMessagePolicy::Persisted,
+        }
+    }
+}
+
 #[uniffi::export]
 impl Welcome {
     /// TLS-serialize this message
@@ -284,6 +302,24 @@ impl CoreCryptoContext {
     ) -> CoreCryptoResult<Vec<u8>> {
         let mut conversation = self.inner.conversation(conversation_id.as_ref()).await?;
         conversation.encrypt_message(message).await.map_err(Into::into)
+    }
+
+    /// Encrypts a plaintext message for one member of the given conversation.
+    ///
+    /// Any feature using a targeted, transient or transient targeted message MUST specify why the lower security
+    /// guarantees (compared to MLS application messages) are acceptable and/or how they are mitigated.
+    pub async fn encrypt_targeted_message(
+        &self,
+        conversation_id: &ConversationId,
+        recipient: Arc<ClientId>,
+        policy: TargetedMessagePolicy,
+        message: Vec<u8>,
+    ) -> CoreCryptoResult<Vec<u8>> {
+        let mut conversation = self.inner.conversation(conversation_id.as_ref()).await?;
+        conversation
+            .encrypt_targeted(recipient.as_ref().as_ref(), policy.into(), message)
+            .await
+            .map_err(Into::into)
     }
 
     /// Joins an existing conversation by constructing an external commit from the given group info.
--- a/crypto/src/mls/conversation/config.rs
+++ b/crypto/src/mls/conversation/config.rs
@@ -35,7 +35,8 @@ pub struct ConversationConfiguration {
 }
 
 impl ConversationConfiguration {
-    const PADDING_SIZE: usize = 128;
+    /// This is a) passed to openmls when creating a group and b) for transient and targeted messages padding.
+    pub(crate) const PADDING_SIZE: usize = 128;
 
     /// Default protocol
     pub(crate) const DEFAULT_PROTOCOL_VERSION: ProtocolVersion = ProtocolVersion::Mls10;
--- a/crypto/src/mls/conversation/immutable/mod.rs
+++ b/crypto/src/mls/conversation/immutable/mod.rs
@@ -12,14 +12,16 @@ use openmls::group::MlsGroup;
 use super::{ConversationIdRef, Error, Result, SecretKey};
 use crate::{
     CipherSuite, ConversationConfiguration, ConversationId, CredentialRef, ExternalSender, OpenMlsError, Session,
+    mls::TntMessageCounter,
 };
 
-#[derive(Debug, derive_more::Constructor, derive_more::Deref)]
+#[derive(derive_more::Constructor, derive_more::Deref, derive_more::DerefMut, derive_more::Debug)]
 pub(crate) struct MlsGroupState {
     #[deref]
+    #[deref_mut]
     group: MlsGroup,
-    // Note: this is going to change to a new type SenderNonce(u32) in an upcoming PR
-    sender_nonce: u32,
+    #[debug(skip)]
+    tnt_message_counter: TntMessageCounter,
 }
 
 impl MlsGroupState {
@@ -31,17 +33,19 @@ impl MlsGroupState {
         &mut self.group
     }
 
-    pub(in crate::mls::conversation) fn sender_nonce(&self) -> u32 {
-        self.sender_nonce
+    pub(in crate::mls::conversation) fn tnt_message_counter(&self) -> TntMessageCounter {
+        self.tnt_message_counter
     }
 
-    #[expect(dead_code)]
-    pub(in crate::mls::conversation) fn increment_sender_nonce(&mut self) {
-        self.sender_nonce += 1;
+    /// Get the tnt message counter bound to this conversation after incrementing it.
+    pub(in crate::mls::conversation) fn obtain_tnt_message_counter(&mut self) -> Result<TntMessageCounter> {
+        self.tnt_message_counter.increment()?;
+        self.group.set_state(openmls::group::InnerState::Changed);
+        Ok(self.tnt_message_counter)
     }
 
-    pub(in crate::mls::conversation) fn reset_sender_nonce(&mut self) {
-        self.sender_nonce = 0
+    pub(in crate::mls::conversation) fn reset_tnt_message_counter(&mut self) {
+        self.tnt_message_counter = Default::default()
     }
 }
 
--- a/crypto/src/mls/conversation/mutable/group_mutation.rs
+++ b/crypto/src/mls/conversation/mutable/group_mutation.rs
@@ -1,10 +1,10 @@
 use core_crypto_keystore::{Transaction, entities::PersistedMlsGroup};
-use openmls::group::{InnerState, MlsGroup};
+use openmls::group::InnerState;
 
 use super::{ConversationMut, Result};
 use crate::{
     KeystoreError, RecursiveError,
-    mls::conversation::{Conversation, ConversationIdRef},
+    mls::conversation::{Conversation, ConversationIdRef, MlsGroupState},
 };
 
 impl ConversationMut {
@@ -24,7 +24,7 @@ impl ConversationMut {
     /// if any lock already exists on that conversation.
     pub(super) async fn mutate_group<T>(
         &mut self,
-        operation: impl AsyncFnOnce(&Transaction, &mut MlsGroup, &ConversationIdRef) -> Result<T>,
+        operation: impl AsyncFnOnce(&Transaction, &mut MlsGroupState, &ConversationIdRef) -> Result<T>,
     ) -> Result<T> {
         // we can't get the transaction if the transaction context has been invalidated,
         // and we want to have that error first before evaluating anything in the operation.
@@ -36,21 +36,21 @@ impl ConversationMut {
         let Conversation { group, id, .. } = &*self.inner;
         let mut group = group.write().await;
         let epoch_before_operation = group.epoch();
-        let ok_result = operation(tx, &mut *group.mls_group_mut(), id).await?;
+        let ok_result = operation(tx, &mut *group, id).await?;
 
         if group.state_changed() == InnerState::Persisted {
             return Ok(ok_result);
         }
 
         if epoch_before_operation < group.epoch() {
-            group.reset_sender_nonce();
+            group.reset_tnt_message_counter();
         }
 
         tx.save(PersistedMlsGroup {
             id: id.to_bytes(),
             state: core_crypto_keystore::ser(group.mls_group())
                 .map_err(KeystoreError::wrap("serializing group state"))?,
-            sender_nonce: group.sender_nonce(),
+            tnt_message_counter: group.tnt_message_counter().into(),
         })
         .await
         .map_err(KeystoreError::wrap("persisting mls group"))?;
@@ -64,7 +64,7 @@ impl ConversationMut {
     #[cfg(test)]
     pub(crate) async fn mutate_group_test<T>(
         &mut self,
-        operation: impl AsyncFnOnce(&Transaction, &mut MlsGroup, &ConversationIdRef) -> Result<T>,
+        operation: impl AsyncFnOnce(&Transaction, &mut MlsGroupState, &ConversationIdRef) -> Result<T>,
     ) -> Result<T> {
         self.mutate_group(operation).await
     }
--- a/crypto/src/mls/conversation/mutable/tnt/mod.rs
+++ b/crypto/src/mls/conversation/mutable/tnt/mod.rs
@@ -1,0 +1,104 @@
+mod targeted;
+mod tnt_message_counter;
+
+use openmls::prelude::Signature;
+use tls_codec::{Serialize, TlsDeserialize, TlsSerialize, TlsSize};
+
+use self::targeted::TargetedMessage;
+pub use self::targeted::encrypt::TargetedMessagePolicy;
+pub(crate) use self::tnt_message_counter::TntMessageCounter;
+
+/// The version of the Transient and Targeted Messages protocol.
+#[derive(TlsSize, TlsSerialize, TlsDeserialize)]
+#[repr(transparent)]
+struct ProtocolVersion(u16);
+
+impl ProtocolVersion {
+    /// Transient and Targeted Messages Version 1
+    const V1: Self = Self(1);
+
+    const fn as_u16(&self) -> u16 {
+        self.0
+    }
+}
+
+/// The body of [TntMessageTBS], with currently three variants. The wire format is encoded in the leading two
+/// bytes of the TLS-serialized version of this struct. The wire-format values are contained in the range reserved for
+/// private use in RFC 9420.
+#[derive(TlsSize, TlsSerialize, TlsDeserialize)]
+#[repr(u16)]
+#[non_exhaustive]
+enum TntMessageBody {
+    /// A transient message.
+    #[tls_codec(discriminant = 0xF000)]
+    Transient(()),
+    /// A targeted message.
+    #[tls_codec(discriminant = 0xF001)]
+    Targeted(TargetedMessage),
+    /// A transient, targeted message.
+    #[tls_codec(discriminant = 0xF002)]
+    TransientTargeted(TargetedMessage),
+}
+
+/// The to-be-signed [TntMessage].
+#[derive(TlsSize, TlsSerialize, TlsDeserialize)]
+struct TntMessageTBS {
+    protocol_version: ProtocolVersion,
+    body: TntMessageBody,
+}
+
+impl TntMessageTBS {
+    #[expect(dead_code)]
+    fn new_transient(transient: ()) -> Self {
+        Self {
+            protocol_version: ProtocolVersion::V1,
+            body: TntMessageBody::Transient(transient),
+        }
+    }
+
+    fn new_targeted(targeted: TargetedMessage) -> Self {
+        Self {
+            protocol_version: ProtocolVersion::V1,
+            body: TntMessageBody::Targeted(targeted),
+        }
+    }
+
+    fn new_transient_targeted(targeted: TargetedMessage) -> Self {
+        Self {
+            protocol_version: ProtocolVersion::V1,
+            body: TntMessageBody::TransientTargeted(targeted),
+        }
+    }
+}
+
+/// A message containing [TntMessageBody] with the corresponding signature.
+#[derive(TlsSize, TlsSerialize, TlsDeserialize)]
+struct TntMessage {
+    content: TntMessageTBS,
+    signature: Signature,
+}
+
+impl openmls::prelude::Signable for TntMessageTBS {
+    type SignedOutput = TntMessage;
+
+    fn unsigned_payload(&self) -> Result<Vec<u8>, tls_codec::Error> {
+        self.tls_serialize_detached()
+    }
+
+    fn label(&self) -> &str {
+        match self.body {
+            TntMessageBody::Transient(_) => todo!(),
+            TntMessageBody::Targeted(_) => TargetedMessage::SIGN_LABEL_PERSISTED,
+            TntMessageBody::TransientTargeted(_) => TargetedMessage::SIGN_LABEL_TRANSIENT,
+        }
+    }
+}
+
+impl openmls::prelude::SignedStruct<TntMessageTBS> for TntMessage {
+    fn from_payload(payload: TntMessageTBS, signature: openmls::prelude::Signature) -> Self {
+        Self {
+            content: payload,
+            signature,
+        }
+    }
+}
--- a/crypto/src/mls/conversation/mutable/tnt/targeted/encrypt.rs
+++ b/crypto/src/mls/conversation/mutable/tnt/targeted/encrypt.rs
@@ -1,0 +1,140 @@
+use std::borrow::Borrow;
+
+use openmls::prelude::{Ciphersuite, Member, Signable as _};
+use tls_codec::{Serialize as _, TlsSerialize, TlsSize, VLBytes};
+
+use super::{PskId, TargetedMessage, TargetedMessageContext};
+use crate::{
+    ClientIdRef, CryptoProvider, OpenMlsError, RecursiveError,
+    mls::conversation::{
+        ConversationMut, Error, MlsGroupState, Result,
+        mutable::tnt::{TntMessage, TntMessageTBS},
+    },
+};
+
+/// The policy to encrypt the targeted message with.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, TlsSize, TlsSerialize)]
+#[repr(u8)]
+pub enum TargetedMessagePolicy {
+    /// Won't be persisted and will only visible to currently online members who immediately process it.
+    Transient,
+    /// May be persisted and buffered, will also be delivered and processed by currently offline members.
+    Persisted,
+}
+
+impl ConversationMut {
+    /// Targeted messages are messages distributed to individual clients. They don't mutate regular group state (i.e, as
+    /// defined by RFC 9420). That is because many or most clients in a group will never receive those messages, and
+    /// will not have the appropriate cryptographic state to decrypt the messages.
+    ///
+    /// Any feature using a targeted, transient or transient targeted message MUST specify why the lower security
+    /// guarantees (compared to MLS application messages) are acceptable and/or how they are mitigated.
+    pub async fn encrypt_targeted(
+        &mut self,
+        recipient: impl Borrow<ClientIdRef>,
+        policy: TargetedMessagePolicy,
+        message: Vec<u8>,
+    ) -> Result<Vec<u8>> {
+        let cipher_suite = self.cipher_suite();
+        let crypto_provider = self
+            .tx_context
+            .crypto_provider()
+            .await
+            .map_err(RecursiveError::transaction("obtaining crypto provider"))?;
+
+        let targeted = self
+            .mutate_group(async |_, group_state, _| {
+                let recipient = group_state
+                    .mls_group()
+                    .members()
+                    .find(|member| ClientIdRef::new(member.credential.identity()) == recipient.borrow())
+                    .ok_or_else(|| Error::MemberNotFound(recipient.borrow().to_owned()))?;
+                Self::create_targeted_message(
+                    policy,
+                    group_state,
+                    &cipher_suite,
+                    &crypto_provider,
+                    &recipient,
+                    &message,
+                )
+            })
+            .await?;
+
+        let tbs = match policy {
+            TargetedMessagePolicy::Transient => TntMessageTBS::new_transient_targeted(targeted),
+            TargetedMessagePolicy::Persisted => TntMessageTBS::new_targeted(targeted),
+        };
+
+        let credential = self.find_current_credential().await?;
+        let signature_key = credential.signature_key();
+        let signed_message: TntMessage = tbs
+            .sign(signature_key)
+            .map_err(OpenMlsError::wrap("signing TntMessageTBS"))?;
+
+        signed_message
+            .tls_serialize_detached()
+            .map_err(Error::tls_serialize("TntMessage: targeted"))
+    }
+
+    fn create_targeted_message(
+        policy: TargetedMessagePolicy,
+        group_state: &mut MlsGroupState,
+        cipher_suite: &Ciphersuite,
+        crypto_provider: &CryptoProvider,
+        recipient: &Member,
+        message: &[u8],
+    ) -> Result<TargetedMessage, Error> {
+        let nonce = group_state.obtain_tnt_message_counter()?;
+        let mls_group = group_state.mls_group();
+        let aad = nonce
+            .tls_serialize_detached()
+            .map_err(Error::tls_serialize("TntMessageCounter"))?;
+        let context = TargetedMessageContext::new(
+            policy,
+            mls_group.export_group_context(),
+            mls_group.own_leaf_index(),
+            recipient.index,
+        );
+        let info = context
+            .tls_serialize_detached()
+            .map_err(Error::tls_serialize("TargetedMessageContext"))?;
+
+        // We can use an empty context because we're using a unique label.
+        let psk = mls_group
+            .export_secret(
+                crypto_provider,
+                TargetedMessage::PSK_LABEL,
+                &[],
+                cipher_suite.hash_length(),
+            )
+            .map_err(OpenMlsError::wrap("exporting targeted message psk"))?;
+        let psk_id = PskId::new(mls_group.group_id().clone(), mls_group.epoch());
+        let psk_id = psk_id.tls_serialize_detached().map_err(Error::tls_serialize("PskId"))?;
+        let message = tls_serialize_padded(message).map_err(Error::tls_serialize("TargetedMessageContent"))?;
+        let payload = crypto_provider
+            .hpke_seal_psk(
+                cipher_suite.hpke_config(),
+                &recipient.encryption_key,
+                &info,
+                &aad,
+                &psk,
+                &psk_id,
+                &message,
+            )
+            .map_err(OpenMlsError::wrap("encrypting targeted message"))?;
+
+        let sender = mls_group.own_leaf_index();
+        let group_id = mls_group.group_id();
+        let epoch = mls_group.epoch();
+
+        let targeted = TargetedMessage::new(nonce, sender, recipient.index, epoch, group_id.clone(), payload);
+        Ok(targeted)
+    }
+}
+
+fn tls_serialize_padded(message: &[u8]) -> Result<Vec<u8>, tls_codec::Error> {
+    // VLBytes includes a length prefix, so whoever decrypts the ciphertext will know the payload length.
+    let mut payload = VLBytes::from(message).tls_serialize_detached()?;
+    payload.resize(payload.len().next_multiple_of(TargetedMessage::PADDING_SIZE), 0);
+    Ok(payload)
+}
--- a/crypto/src/mls/conversation/mutable/tnt/targeted/mod.rs
+++ b/crypto/src/mls/conversation/mutable/tnt/targeted/mod.rs
@@ -1,0 +1,79 @@
+pub(super) mod encrypt;
+
+use const_format::concatcp;
+use derive_more::Constructor;
+use openmls::{
+    group::{GroupEpoch, GroupId, group_context::GroupContext},
+    prelude::{HpkeCiphertext, LeafNodeIndex},
+};
+use tls_codec::{TlsDeserialize, TlsSerialize, TlsSize};
+
+use self::encrypt::TargetedMessagePolicy;
+use super::{ProtocolVersion, tnt_message_counter::TntMessageCounter};
+use crate::ConversationConfiguration;
+
+/// Used to parameterize HPKE Seal/Open.
+/// Not carried with the payload, constructed freshly when decrypting.
+#[derive(TlsSize, TlsSerialize)]
+pub(super) struct TargetedMessageContext {
+    protocol_version: ProtocolVersion,
+    policy: TargetedMessagePolicy,
+    sender: LeafNodeIndex,
+    recipient: LeafNodeIndex,
+    group_context: GroupContext,
+}
+
+impl TargetedMessageContext {
+    fn new(
+        policy: TargetedMessagePolicy,
+        group_context: &GroupContext,
+        sender: LeafNodeIndex,
+        recipient: LeafNodeIndex,
+    ) -> Self {
+        Self {
+            policy,
+            protocol_version: ProtocolVersion::V1,
+            sender,
+            recipient,
+            group_context: group_context.clone(),
+        }
+    }
+}
+
+/// Used to parametrize HPKE Seal/Open
+#[derive(Constructor, TlsSize, TlsSerialize)]
+pub(super) struct PskId {
+    group_id: GroupId,
+    epoch: GroupEpoch,
+}
+
+/// Targeted messages are messages distributed to individual clients. They don't mutate regular group state (i.e, as
+/// defined by RFC 9420). That is because many or most clients in a group will never receive those messages, and will
+/// not have the appropriate cryptographic state to decrypt the messages.
+///
+/// Any feature using a targeted, transient or transient targeted message MUST specify why the lower security
+/// guarantees (compared to MLS application messages) are acceptable and/or how they are mitigated.
+///
+/// Implementation note: we're preferring OpenMLS types here, because we're using OpenMLS encoding. Once we switch to
+/// mls-rs, we're going to use mls-rs types.
+#[derive(TlsSize, TlsSerialize, TlsDeserialize, Constructor)]
+pub(super) struct TargetedMessage {
+    nonce: TntMessageCounter,
+    sender: LeafNodeIndex,
+    recipient: LeafNodeIndex,
+    epoch: GroupEpoch,
+    group_id: GroupId,
+    /// Encrypted with the receipient's HPKE public key; authenticated with a PSK derived from the epoch secret,
+    /// guaranteeing that the sender was a member of the group during the epoch. An additional authentication layer is
+    /// provided by the signature over the entire message. Before encryption, the HPKE plaintext is padded.
+    payload: HpkeCiphertext,
+}
+
+impl TargetedMessage {
+    pub(super) const PADDING_SIZE: usize = ConversationConfiguration::PADDING_SIZE;
+    pub(super) const SIGN_LABEL_PERSISTED: &str =
+        concatcp!("TntMessageTBS-Persisted-Targeted v", ProtocolVersion::V1.as_u16());
+    pub(super) const SIGN_LABEL_TRANSIENT: &str =
+        concatcp!("TntMessageTBS-Transient-Targeted v", ProtocolVersion::V1.as_u16());
+    pub(super) const PSK_LABEL: &str = concatcp!("Tnt TargetedMessage Psk v", ProtocolVersion::V1.as_u16());
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1090,6 +1090,7 @@ dependencies = [
  "async-trait",
  "base64 0.23.1",
  "chacha20poly1305",
+ "const_format",
  "core-crypto-keystore",
  "core-crypto-macros",
  "derive_more",
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -22,6 +22,7 @@ test-all-cipher = []
 
 [dependencies]
 derive_more.workspace = true
+const_format.workspace = true
 enum-as-inner.workspace = true
 thiserror.workspace = true
 strum.workspace = true
--- a/crypto/src/error/open_mls.rs
+++ b/crypto/src/error/open_mls.rs
@@ -118,4 +118,7 @@ pub enum OpenMlsErrorKind {
     /// Keystore Error
     #[error(transparent)]
     KeystoreError(#[from] core_crypto_keystore::CryptoKeystoreError),
+    /// Signature Error
+    #[error(transparent)]
+    SignatureError(#[from] openmls::prelude::SignatureError),
 }
--- a/crypto/src/mls/conversation/error.rs
+++ b/crypto/src/mls/conversation/error.rs
@@ -4,6 +4,7 @@
 #![allow(missing_docs)]
 
 use super::config::MAX_PAST_EPOCHS;
+use crate::ClientId;
 
 pub type Result<T, E = Error> = core::result::Result<T, E>;
 
@@ -13,6 +14,13 @@ pub enum Error {
     InappropriateMessageBodyType,
     #[error("The current client id isn't authorized to perform this action")]
     Unauthorized,
+    #[error("Member with id {_0} doesn't exist")]
+    MemberNotFound(ClientId),
+    #[error(
+        "The tnt message counter has overflowed. You sent 2^32 targeted or transient messages in this epoch. Increment
+        the epoch and try again."
+    )]
+    TntMessageCounterOverflow,
     /// Message rejected by the delivery service
     #[error("Message rejected by the delivery service. Reason: {reason}")]
     MessageRejected {
--- a/crypto/src/mls/conversation/immutable/persistence.rs
+++ b/crypto/src/mls/conversation/immutable/persistence.rs
@@ -6,12 +6,15 @@ use openmls::group::MlsGroup;
 use super::Result;
 use crate::{
     ConversationConfiguration, ConversationId, KeystoreError, Session,
-    mls::conversation::{Conversation, MlsGroupState},
+    mls::{
+        TntMessageCounter,
+        conversation::{Conversation, MlsGroupState},
+    },
 };
 
 impl Conversation {
     /// restore the conversation from a persistence-saved serialized Group State.
-    fn from_serialized_state(session: Session, buf: Vec<u8>, sender_nonce: u32) -> Result<Self> {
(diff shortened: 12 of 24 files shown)


crypto/src/mls/core_crypto_message/mod.rs
```diff
@@ -11,3 +14,91 @@ impl TransientAndTargetedMessagesProtocolVersion {
...
+pub enum CoreCryptoMessageBody {
+    /// Transient message.
+    #[tls_codec(discriminant = 0xF000)]
+    Transient(()),
+    /// [TargetedMessage].
+    #[tls_codec(discriminant = 0xF001)]
+    Targeted(TargetedMessage),
+    /// A [TargetedMessage] encrypted with transient policy.
+    #[tls_codec(discriminant = 0xF002)]
+    TransientTargeted(TargetedMessage),
```
coriolinus: Not sure what "encrypted with transient policy" means. Maybe just say "A transient, targeted message" in line with the other variants' doc-comments.

crypto/src/mls/core_crypto_message/targeted.rs
```diff
@@ -0,0 +1,69 @@
...
+    mls::{SenderNonce, core_crypto_message::TransientAndTargetedMessagesProtocolVersion},
+};
+
+/// Used to parameterize HPKE Seal/Open.
+/// Not carried with the payload, constructed freshly when decrypting.
+#[derive(TlsSize, TlsSerialize, TlsDeserialize)]
+pub struct TargetedMessageContext {
+    protocol_version: TransientAndTargetedMessagesProtocolVersion,
+    group_context: GroupContext,
+}
```
coriolinus: I think that the derived code here is wrong, unfortunately, because it completely skips `Self::ENCRYPTION_LABEL`. We might have to implement those traits manually, or use a related struct or something.

crypto/src/mls/conversation/mutable/tnt/targeted/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// The label used when encrypting a Targeted Message.
+    pub const ENCRYPTION_LABEL: &str = "CoreCrypto TargetedMessageContext";
+}
+
+/// Used to parametrize HPKE Seal/Open
+#[derive(Constructor, TlsSize, TlsSerialize)]
+pub struct PskId {
+    group_id: GroupId,
+    epoch: GroupEpoch,
+}
```
coriolinus, line 48: Also missing a label.

crypto/src/mls/conversation/mutable/tnt/targeted.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// The encrypted payload will typically be GenericMessage protobuf.
+    /// Must have padding like the openmls ciphertexts.
+    payload: HpkeCiphertext,
+}
+
+impl TargetedMessage {
+    pub(crate) const PADDING_SIZE: usize = ConversationConfiguration::PADDING_SIZE;
+    pub(super) const SIGN_LABEL_PERISTED: &str = "CoreCryptoMessageTBS-Persisted-Targeted";
+    pub(super) const SIGN_LABEL_TRANSIENT: &str = "CoreCryptoMessageTBS-Transient-Targeted";
+    pub(crate) const PSK_LABEL: &str = "CoreCrypto TargetedMessage Psk";
```
coriolinus: This is not exactly what's written in the RFC. Intentional?

crypto/src/mls/core_crypto_message/targeted.rs
```diff
@@ -0,0 +1,69 @@
...
+    group_id: GroupId,
+    /// Ecrypted with the receipient's HPKE public key; authenticated with epoch secret as PSK.
+    /// The encrypted payload will typically be GenericMessage protobuf.
+    /// Must have padding like the openmls ciphertexts.
+    payload: HpkeCiphertext,
+}
+
+impl TargetedMessage {
+    pub(crate) const PADDING_SIZE: usize = ConversationConfiguration::PADDING_SIZE;
+    pub(super) const SIGN_LABEL_PERISTED: &str = "CoreCryptoMessageTBS-Persisted-Targeted";
```
coriolinus: Typo: "PERISTED" -> "PERSISTED"

crypto/src/mls/conversation/immutable/mod.rs
```diff
@@ -31,17 +32,16 @@ impl MlsGroupState {
...
 
-    pub(in crate::mls::conversation) fn sender_nonce(&self) -> u32 {
+    pub(in crate::mls::conversation) fn sender_nonce(&self) -> SenderNonce {
         self.sender_nonce
     }
 
-    #[expect(dead_code)]
     pub(in crate::mls::conversation) fn increment_sender_nonce(&mut self) {
-        self.sender_nonce += 1;
+        self.sender_nonce.increment();
```
coriolinus: This should probably mark the conversation as dirty so it saves.

crypto/src/mls/conversation/mutable/group_mutation.rs
```diff
@@ -51,7 +63,7 @@ impl ConversationMut {
             state: core_crypto_keystore::ser(group.mls_group())
                 .map_err(KeystoreError::wrap("serializing group state"))?,
             parent_id: None,
-            sender_nonce: group.sender_nonce(),
+            sender_nonce: group.sender_nonce().into(),
         })
```
coriolinus, line 54: If we mutate the sender nonce but not otherwise the group state, we return early before persisting the nonce.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,148 @@
...
+            .mutate_group_and_sender_nonce(async |_, group, _| {
+                group.increment_sender_nonce();
+                Ok(group.sender_nonce())
+            })
+            .await?;
+
+        let mls_group_guard = self.group().await;
+        let mls_group = mls_group_guard.mls_group();
+
+        let recipient = self.client_id_member(recipient).await?;
```
coriolinus: `Self::client_id_member` also acquires a short-lived read guard. Could we get into a deadlock if someone enqueued a write lock between the two reads?

Anyway, binding `recipient` first eliminates that potential, so we may as well reorder it.

crypto/src/mls/conversation/immutable/mod.rs
SimonThormeyer: Very good catch!

crypto/src/mls/conversation/mutable/group_mutation.rs
SimonThormeyer: Because of the review comment context window just spanning from these lines, I'm not 100% sure which early return your're referring to. Is it
```rust
        if group.state_changed() == InnerState::Persisted {
            return Ok(ok_result);
        }

```
?

Marking the group dirty as dirty when incrementing the nonce, as you hinted above, should fix this then.

crypto/src/mls/conversation/mutable/tnt/targeted/mod.rs
SimonThormeyer: As discussed, this is covered by the wire format.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,140 @@
...
+        recipient: impl Borrow<ClientIdRef>,
+        policy: TargetedMessagePolicy,
+        message: Vec<u8>,
+    ) -> Result<Vec<u8>> {
+        let nonce = self
+            .mutate_group_and_sender_nonce(async |_, group, _| {
+                group.increment_sender_nonce();
+                Ok(group.sender_nonce())
+            })
+            .await?;
```
istankovic, line 61: Would it be possible to combine `mutate_group_and_sender_nonce` and `increment_sender_nonce` into a single operation (e.g. `obtain_sender_nonce` or similar) that would not allow using previous nonce values?

The idea is that the only way to get a nonce would imply changing it and we'd get a compile-time guarantee about that.

crypto/src/mls/core_crypto_message/targeted.rs
```diff
@@ -0,0 +1,69 @@
...
+    group_id: GroupId,
+    epoch: GroupEpoch,
+}
+
+/// Targeted messages are messages distributed to individual clients. They cannot mutate group state because many or
+/// most clients in a group will never receive those messages, and will not have the appropriate cryptographic state to
+/// decrypt the messages.
+///
+/// Any feature using a targeted, transient or transient targeted message MUST specify why the compared to MLS
+/// application messages lower guarantees are acceptable and/or how they are mitigated.
```
fewerner: We should add a section in cc-book about different message types and a guide on when to use them. The second paragraph should not just be documented in code. Not necessarily a part of this PR.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,140 @@
...
+                    panic!("Cannot produce a transient message body in `encrypt_targeted()`")
+                }
+                CoreCryptoMessageBody::Targeted(_) => {
+                    assert!(policy == TargetedMessagePolicy::Persisted)
+                }
+                CoreCryptoMessageBody::TransientTargeted(_) => {
+                    assert!(policy == TargetedMessagePolicy::Transient)
+                }
+            }
+        }
```
istankovic: It feels strange that we need to do this. Can we ensure these invariants at compile time?

For instance, why not make `CoreCryptoMessageTBS` generic with a `policy` type parameter?

crypto/src/mls/conversation/mutable/tnt/targeted.rs
SimonThormeyer: Updated the RFC accordingly.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
SimonThormeyer: We don't really need to do that. I copied this from `ConversationMut::encrypt()` because I thought if it's there, we should also do it here for consistency.

I'm not sure about the generic TBS and if it would work as nicely with the serialization and wire format. But could give it a try if we think that it's worth it.

fewerner (commented):
Not much to add from my side. GJ so far!

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
SimonThormeyer: 522bc953d7bf4cda8a31293e21d6d969dc113558

crypto/src/mls/core_crypto_message/targeted.rs
SimonThormeyer: Absolutely. Will be added with a later PR.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
istankovic: I'm not sure how it's done there, but if we have a similar thing there, then it would be worth exploring an alternative approach that would provide us compile-time guarantees. Would be nice to have something to compare.

crypto/src/mls/conversation/mutable/group_mutation.rs
coriolinus: Yes, sorry for the confusion. Agree you've handled this now.

crypto-ffi/src/core_crypto_context/mls.rs
```diff
@@ -62,6 +62,24 @@ bytes_wrapper!(
     Welcome fallibly wraps core_crypto::mls::conversation::WelcomeMessage
 );
 
+/// The policy to encrypt the targeted message with.
+#[derive(Debug, Clone, Copy, uniffi::Enum)]
+pub enum TargetedMessagePolicy {
+    /// Won't be persisted and will only visible to currently online members who immediately process it.
```
istankovic: Nit: `only visible` -> `only be visible`?

crypto-ffi/src/core_crypto_context/mls.rs
```diff
@@ -286,6 +304,24 @@ impl CoreCryptoContext {
         conversation.encrypt_message(message).await.map_err(Into::into)
     }
 
+    /// Encrypts a plaintext message for one member of the given conversation.
+    ///
+    /// Any feature using a targeted, transient or transient targeted message MUST specify why the compared to MLS
+    /// application messages lower guarantees are acceptable and/or how they are mitigated.
```
istankovic: Nit: `why the compared to MLS application messages lower guarantees` is hard to read.

I suggest something like `why the lower security guarantees (compared to MLS application messages)`.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// Won't be persisted and will only visible to currently online members who immediately process it.
+    Transient,
+    /// May be persisted and buffered, will also be delivered and processed by currently offline members.
+    Persisted,
+}
+
+impl super::ConversationMut {
+    /// Targeted messages are messages distributed to individual clients. They cannot mutate group state because many or
+    /// most clients in a group will never receive those messages, and will not have the appropriate cryptographic state
+    /// to decrypt the messages.
```
istankovic: I think it would be good to point out that group state _is_ going to be mutated: both on the sender side and the receiver side. Perhaps it is useful to distinguish between "regular MLS group state", as defined by RFC 9420, and our extension to that state in order to support these messages.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,135 @@
...
+    ///
+    /// Any feature using a targeted, transient or transient targeted message MUST specify why the compared to MLS
+    /// application messages lower guarantees are acceptable and/or how they are mitigated.
```
istankovic: Same nit as in https://github.com/wireapp/core-crypto/pull/2421/changes#r3712952851.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,140 @@
...
```
coriolinus: I think some formatting was lost, or a paste went wrong, or something. This sentence is hard to parse.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,140 @@
...
+        message: Vec<u8>,
+    ) -> Result<TargetedMessage, Error> {
+        let cipher_suite = self.cipher_suite();
+        let aad = nonce
+            .tls_serialize_detached()
+            .map_err(Error::tls_serialize("SenderNonce"))?;
+        let context = TargetedMessageContext::new(mls_group.export_group_context());
+        let info = context
+            .tls_serialize_detached()
+            .map_err(Error::tls_serialize("TargetedMessagaeContext"))?;
```
coriolinus: typo: "Messagae"

crypto/src/mls/conversation/mutable/tnt/targeted/encrypt.rs
```diff
@@ -0,0 +1,140 @@
...
+            .tx_context
+            .crypto_provider()
+            .await
+            .map_err(RecursiveError::transaction("getting crypto provider"))?;
+
+        // We can use an empty context because we're using a unique label.
+        let psk = mls_group
+            .export_secret(
+                crypto_provider,
+                TargetedMessage::PSK_LABEL,
```
coriolinus, line 106: This label is constant regardless of whether this is a normal targeted message or a transient targeted message. 

I'm starting to think this is wrong. If the keys, and therefore the signed messages are the same, than an attacker (who collaborates with the backend) can switch between normal-targeted and transient-targeted semantics by intercepting and rewriting the envelope, and re-signing. They aren't changing or decrypting the encrypted bytes, but they're claiming attribution over them.

... That's true whether or not we distinguish between transient and persistent targeted messages, I suppose. I haven't finished thinking this through. But the immediate sense is that it would be better to at least prevent them from switching transient mode on or off arbitrarily, which is as easy as creating distinct labels here.

crypto/src/mls/conversation/mutable/tnt/targeted/encrypt.rs
```diff
@@ -0,0 +1,135 @@
...
+        nonce: SenderNonce,
+        recipient: Member,
+        message: Vec<u8>,
+    ) -> Result<TargetedMessage, Error> {
+        let mls_group_guard = self.group().await;
+        let mls_group = mls_group_guard.mls_group();
+        let cipher_suite = self.cipher_suite();
+        let aad = nonce
+            .tls_serialize_detached()
+            .map_err(Error::tls_serialize("SenderNonce"))?;
```
coriolinus: I think we should bind the indices of `sender` and `recipient` in the AAD here, to prevent the kind of substitution attack I mentioned earlier. Maybe also directly bind the `TargetedMessagePolicy`, to make it absolutely explicit.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub async fn encrypt_targeted(
+        &mut self,
+        recipient: impl Borrow<ClientIdRef>,
+        policy: TargetedMessagePolicy,
+        message: Vec<u8>,
+    ) -> Result<Vec<u8>> {
+        let nonce = self.obtain_sender_nonce().await?;
+        let recipient = self.client_id_member(recipient).await?;
+
+        let targeted = self.create_targeted_message(nonce, recipient, message).await?;
```
coriolinus: This takes three locks in a row; each of these functions locks internally.

So what happens if a commit comes in between the first and the last of those? We're now encrypting a message with a nonce obtained in epoch `N`, in epoch `N+1`. That can violate our claim that we never reuse a nonce: in the likely case, both the current message and the subsequent carry nonce 0.

I think we need to refactor these such that there is one lock ensuring that these operations are atomic.

crypto/src/mls/core_crypto_message/targeted.rs
```diff
@@ -0,0 +1,69 @@
...
+            protocol_version: TransientAndTargetedMessagesProtocolVersion::V1,
+            group_context: group_context.clone(),
+        }
+    }
+}
+
+impl TargetedMessageContext {
+    /// The label used when encrypting a Targeted Message.
+    pub const ENCRYPTION_LABEL: &str = "CoreCrypto TargetedMessageContext";
+}
```
coriolinus: This is dead code, replaced by the discriminant, right?

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,135 @@
...
+        Ok(targeted)
+    }
+}
+
+fn tls_serialize_padded(message: Vec<u8>) -> Result<Vec<u8>, tls_codec::Error> {
+    // VLBytes includes a length prefix, so whoever decrypts the ciphertext will know the payload length.
+    let mut payload = VLBytes::from(message).tls_serialize_detached()?;
+    payload.resize(payload.len().next_multiple_of(TargetedMessage::PADDING_SIZE), 0);
+    Ok(payload)
+}
```
coriolinus: This pads the plaintext. I believe that [openmls pads the ciphertext](https://github.com/wireapp/openmls/blob/c3bdf5a192013af53b456131440c2190aa61c00d/openmls/src/framing/private_message.rs#L267-L275).

I don't believe that there's a security difference between the two, but the RFC says "padding like the openmls ciphertext". This technically isn't in compliance. We might just amend the RFC to be more specific.

crypto/src/mls/core_crypto_message/sender_nonce.rs
```diff
@@ -0,0 +1,12 @@
+use tls_codec::{TlsDeserialize, TlsSerialize, TlsSize};
+
+/// A shared nonce type for Transient and Targeted Messages.
+#[derive(Clone, Default, Copy, TlsSize, TlsSerialize, TlsDeserialize, derive_more::Into, derive_more::From)]
+#[repr(transparent)]
+pub struct SenderNonce(u32);
+
+impl SenderNonce {
+    pub(crate) fn increment(&mut self) {
+        self.0 += 1;
```
coriolinus: Technically this can overflow in release mode. It's cheap to add an error variant and use `checked_add` here.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,135 @@
...
+                    panic!("Cannot produce a transient message body in `encrypt_targeted()`")
+                }
+                CoreCryptoMessageBody::Targeted(_) => {
+                    assert!(policy == TargetedMessagePolicy::Persisted)
+                }
+                CoreCryptoMessageBody::TransientTargeted(_) => {
+                    assert!(policy == TargetedMessagePolicy::Transient)
+                }
+            }
+        }
```
coriolinus: Can this even fail? We construct the `tbs` a few lines earlier within this same function. And it looks like `CoreCryptoMessage::body` only exists to serve this infallible assertion.

coriolinus (dismissed):
Two things are true simultaneously: this is great work, but also it has some real problems. I don't think we can merge this as-is quite yet.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
SimonThormeyer: I would also say that if we don't mess up the mapping between policy and message type, this assertion can never fail. So this is more of a regression check? However, if we say that we don't need this, wouldn't the same apply to the end of `encrypt_message()` (which is where I got the idea for this assertion from)? If we say that it's redundant (which it probably will be at the latest as soon as we have tests), should we remove both?

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
SimonThormeyer: ﻿﻿﻿8f4831e1af40630b4cc1c6c4e6f0c4beaf5c33cf

crypto/src/mls/conversation/mutable/tnt/targeted/encrypt.rs
SimonThormeyer: No, they cannot switch, at least not without the sender's signature private key: we have two different wire formats - `0xF001` for `TargetedMessage`, and `0xF002` for `TransientTargetedMessage`. This wire format is part of the signed payload, so nobody can swap it without the signature private key.

coriolinus: An in-conversation middleman can swap out the entire signed envelope as long as they claim authorship. This lets them swap the wire format.

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
SimonThormeyer: I updated the RFC accordingly

istankovic: Are you sure about that? It doesn't make sense to pad the ciphertext and to me it looks they actually pad before encryption, not after, see [https://github.com/wireapp/openmls/blob/c3bdf5a192013af53b456131440c2190aa61c00d/openmls/src/framing/private_message.rs#L183-L195](https://www.google.com/url?q=https://github.com/wireapp/openmls/blob/c3bdf5a192013af53b456131440c2190aa61c00d/openmls/src/framing/private_message.rs%23L183-L195&source=gmail-imap&ust=1786523517000000&usg=AOvVaw0xPLldOi94G0yLnMukFOvS).

SimonThormeyer: I updated the RFC with this:

`The plaintext that produced the ciphertext has been zero-padded to to a length of plaintext.len().next_multiple_of(128).`

crypto/src/mls/conversation/mutable/tnt/targeted/encrypt.rs
SimonThormeyer: befc9941aa7b08d929296f0efea6105558a7677c and 9e2e87cc5eb4138037f6e5a2d83bb30c316a3a66

crypto/src/mls/conversation/mutable/tnt/targeted/encrypt.rs
SimonThormeyer: https://github.com/wireapp/core-crypto/commit/befc9941aa7b08d929296f0efea6105558a7677c and https://github.com/wireapp/core-crypto/commit/9e2e87cc5eb4138037f6e5a2d83bb30c316a3a66

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
SimonThormeyer: d370db61c67fdb79ee645b0aee7eb40e5bdcf6ec

crypto/src/mls/core_crypto_message/sender_nonce.rs
SimonThormeyer: e54d6edddf47f0dfc1e5881f2924b4802bd018e3

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
SimonThormeyer: If I just look at these three lines, yes; but to change the epoch (be it via sending or receiving a commit), we'd need a reference to `&mut self`, which we already posess here. 

Would like us to hold a read lock to the `MlsGroup` across these functions, anyway? Then we'd need to implement these functions via an extension trait to `MlsGroup`, right? Or is there a more direct way? 
Anyway, because of the reason above I'm not sure if this is necessary.

crypto/src/mls/core_crypto_message/targeted.rs
SimonThormeyer: b2ed23f31b2346dc54fd2bec269b980033c30341

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
SimonThormeyer: As discussed, we can just remove this:
d370db61c67fdb79ee645b0aee7eb40e5bdcf6ec

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
SimonThormeyer: 8f4831e1af40630b4cc1c6c4e6f0c4beaf5c33cf

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
coriolinus: We can't trust the `&mut self` to guarantee exclusivity in that sense because we can construct an arbitrary number of instances of `ConversationMut`; `TransactionContext::conversation` just gets an `Arc<Conversation>` guard, then calls `ConversationMut::new` on that. So I think we do need to depend on an explicit lock, unfortunately.

I think that the natural approach, given that we have to do it all in a single lock context, is to implement everything as methods on `&mut MlsGroupState`. Seems simpler there than to use an extension trait on `MlsGroup`.

SimonThormeyer: 441e49a9739221069d71d5e3b2cd31a6ecba56f9 and 3e08d9e216eaf388bec7a3bca9e23073c3a8e9ee

crypto-ffi/src/core_crypto_context/mls.rs
```diff
@@ -286,6 +304,24 @@ impl CoreCryptoContext {
         conversation.encrypt_message(message).await.map_err(Into::into)
     }
 
+    /// Encrypts a plaintext message for one member of the given conversation.
+    ///
+    /// Any feature using a targeted, transient or transient targeted message MUST specify why why the lower security
```
istankovic: Nit: `why why` -> `why`

crypto/src/mls/conversation/immutable/mod.rs
```diff
@@ -31,17 +32,19 @@ impl MlsGroupState {
...
         self.sender_nonce
     }
 
-    #[expect(dead_code)]
-    pub(in crate::mls::conversation) fn increment_sender_nonce(&mut self) {
-        self.sender_nonce += 1;
+    /// Get the sender nonce bound to this conversation after incrementing it.
+    pub(in crate::mls::conversation) fn obtain_sender_nonce(&mut self) -> Result<SenderNonce> {
+        self.group.set_state(openmls::group::InnerState::Changed);
+        self.sender_nonce.increment()?;
```
istankovic: If we use this order, wouldn't we mark the state as changed regardless of whether we actually managed to update the nonce?

SimonThormeyer: Right, good catch!

SimonThormeyer: 9af16d6040ac98d095a2ca2d8d90ded7e53fb6ee

crypto/src/mls/conversation/mutable/tnt/targeted/encrypt.rs
```diff
@@ -0,0 +1,133 @@
...
+            .map_err(RecursiveError::transaction("obtaining crypto provider"))?;
+
+        let targeted = self
+            .mutate_group_and_sender_nonce(async |_, group_state, _| {
+                let nonce = group_state.obtain_sender_nonce()?;
+                let mls_group = group_state.mls_group();
+                let recipient = mls_group
+                    .members()
+                    .find(|member| ClientIdRef::new(member.credential.identity()) == recipient.borrow())
+                    .ok_or_else(|| Error::MemberNotFound(recipient.borrow().to_owned()))?;
```
istankovic, line 51: It would be better to move the member search prior to getting the nonce so that if the member search fails we don't unnecessarily increment the nonce and mark the state as dirty.

SimonThormeyer: b0779e8d990f93e34eba71f36b22db5d18318066

crypto/src/mls/core_crypto_message/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+pub enum CoreCryptoMessageBody {
+    /// A transient message.
+    #[tls_codec(discriminant = 0xF000)]
+    Transient(()),
+    /// A targeted message.
+    #[tls_codec(discriminant = 0xF001)]
+    Targeted(TargetedMessage),
+    /// A transient, targeted message.
+    #[tls_codec(discriminant = 0xF002)]
+    TransientTargeted(TargetedMessage),
```
istankovic: While these are mentioned in the RFC, they are buried in sample code, which is not obvious at a glance. Could you please add a section to the RFC and a nice table with these allocated numbers? Ideally, we would also reference the relevant section in RFC 9420 that specifies the reserved range.

crypto/src/mls/conversation/config.rs
```diff
@@ -35,7 +35,7 @@ pub struct ConversationConfiguration {
 }
 
 impl ConversationConfiguration {
-    const PADDING_SIZE: usize = 128;
+    pub(crate) const PADDING_SIZE: usize = 128;
```
istankovic, line 39: Please add a doc comment here to describe what this is for (thinking of myself 2 months from now). 
Is this also used for regular MLS messages?

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,147 @@
...
+        let cipher_suite = self.cipher_suite();
+        let crypto_provider = self
+            .tx_context
+            .crypto_provider()
+            .await
+            .map_err(RecursiveError::transaction("obtaining crypto provider"))?;
+
+        let targeted = self
+            .mutate_group_and_sender_nonce(async |_, group_state, _| {
+                // We're borrowing the mls group separately twice, because we cannot hold a reference to the grop across
```
istankovic: Nit: `grop` -> `group`

crypto/src/mls/conversation/mutable/encrypt_targeted.rs
```diff
@@ -0,0 +1,147 @@
...
+                // the call of  `obtain_sender_nonce()`, which needs a mutable reference to the state.
```
istankovic: Nit: extra space after `of`

crypto/src/mls/core_crypto_message/targeted.rs
```diff
@@ -0,0 +1,78 @@
...
+    mls::{
+        SenderNonce, conversation::TargetedMessagePolicy,
+        core_crypto_message::TransientAndTargetedMessagesProtocolVersion,
+    },
+};
+
+/// Used to parameterize HPKE Seal/Open.
+/// Not carried with the payload, constructed freshly when decrypting.
+#[derive(TlsSize, TlsSerialize)]
+pub struct TargetedMessageContext {
```
istankovic: This should probably be `pub(crate)`.

crypto/src/mls/conversation/config.rs
SimonThormeyer: Yes, it is passed to openmls when creating a group.

crypto/src/mls/core_crypto_message/targeted.rs
SimonThormeyer: That's what I had first, but then we can't link to it in any doc comment.

crypto/src/mls/conversation/mutable/tnt/targeted/encrypt.rs
```diff
@@ -0,0 +1,148 @@
...
+        let group_id = mls_group.group_id();
+        let epoch = mls_group.epoch();
+
+        let targeted = TargetedMessage::new(nonce, sender, recipient.index, epoch, group_id.clone(), payload);
+        Ok(targeted)
+    }
+}
+
+fn tls_serialize_padded(message: &[u8]) -> Result<Vec<u8>, tls_codec::Error> {
+    // VLBytes includes a length prefix, so whoever decrypts the ciphertext will know the payload length.
```
istankovic, line 136: I appreciate this comment! 💜

crypto/src/mls/conversation/mutable/tnt/targeted/encrypt.rs
```diff
@@ -0,0 +1,148 @@
...
+    let mut payload = VLBytes::from(message).tls_serialize_detached()?;
+    payload.resize(payload.len().next_multiple_of(TargetedMessage::PADDING_SIZE), 0);
```
istankovic, line 138: Interesting thing here is that `x.next_multiple_of(y)` will just return `x` if `x` is already a multiple of `y`.
For example, if plaintext is 128 bytes and `PADDING_SIZE = 128`,  there's going to be no added padding.
I guess this is probably fine, but good to keep in mind.

coriolinus: Least surprise, I guess. I'd be surprised if x changed when `x % y` was already 0.

crypto/src/mls/core_crypto_message/targeted.rs
istankovic: > That's what I had first, but then we can't link to it in any doc comment.

You tried linking inside this crate, or outside?

crypto/src/mls/conversation/mutable/tnt/targeted/encrypt.rs
SimonThormeyer: Agree. This is probably one of the simplest possible implementations of a padding.

crypto/src/mls/core_crypto_message/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use openmls::prelude::Signature;
+use tls_codec::{Serialize, TlsDeserialize, TlsSerialize, TlsSize};
+
+pub use crate::mls::core_crypto_message::targeted::{PskId, TargetedMessage, TargetedMessageContext};
+
+/// The version of the Transient and Targeted Messages protocol.
+#[derive(TlsSize, TlsSerialize, TlsDeserialize)]
+#[repr(transparent)]
+pub struct TransientAndTargetedMessagesProtocolVersion(u16);
```
istankovic: I'm getting some Java vibes here. :-)

How about changing the module name from `core_crypto_message` to just e.g. `targeted` and then this could be referenced as `targeted::ProtocolVersion`?

This should also probably be `pub(crate)` as I don't think there should be any usage outside `crypto`.

crypto/src/mls/core_crypto_message/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+    #[tls_codec(discriminant = 0xF001)]
+    Targeted(TargetedMessage),
+    /// A transient, targeted message.
+    #[tls_codec(discriminant = 0xF002)]
+    TransientTargeted(TargetedMessage),
+}
+
+/// The to-be-signed [CoreCryptoMessage].
+#[derive(TlsSize, TlsSerialize, TlsDeserialize)]
+pub struct CoreCryptoMessageTBS {
```
istankovic: Please make this `pub(crate)`. Actually, does anything outside this module even need this? It's fairly low-level.

crypto/src/mls/core_crypto_message/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+        Self {
+            protocol_version: TransientAndTargetedMessagesProtocolVersion::V1,
+            body: CoreCryptoMessageBody::TransientTargeted(targeted),
+        }
+    }
+}
+
+/// A message containing [CoreCryptoMessageBody] with the corresponding signature.
+#[derive(TlsSize, TlsSerialize, TlsDeserialize)]
+pub struct CoreCryptoMessage {
```
istankovic: Similar thing here: this is fairly low-level and should be at most `pub(crate)`, and ideally private to the module actually using it.

SimonThormeyer: 72a663c1d1450a28bdce6369e4f7d46b420763fe

crypto/src/mls/core_crypto_message/mod.rs
SimonThormeyer: 72a663c1d1450a28bdce6369e4f7d46b420763fe

crypto/src/mls/core_crypto_message/mod.rs
SimonThormeyer: Note that `targeted` is a child module of this and the protocol version will be used by `transient` also.

How about `CoreCryptoMessagesProtocolVersion` for a name?

crypto/src/mls/core_crypto_message/targeted.rs
SimonThormeyer: Making everything `pub(crate)` at once seemed to work, no more errors from rustdoc.

crypto/src/mls/core_crypto_message/sender_nonce.rs
```diff
@@ -0,0 +1,19 @@
+use tls_codec::{TlsDeserialize, TlsSerialize, TlsSize};
+
+/// A shared nonce type for Transient and Targeted Messages.
+#[derive(Clone, Default, Copy, TlsSize, TlsSerialize, TlsDeserialize, derive_more::Into, derive_more::From)]
+#[repr(transparent)]
+pub struct SenderNonce(u32);
```
istankovic: Similarly to other places, this should also be at most `pub(crate)`.

Additionally, why do we have `Clone` and `Copy` here?
Would it be possible to drop those? It would be preferable to restrict unlimited copying/cloning of nonces so we can more easily track and reason about them.

crypto/src/mls/core_crypto_message/targeted.rs
```diff
@@ -0,0 +1,78 @@
...
+    group_id: GroupId,
+    epoch: GroupEpoch,
+}
+
+/// Targeted messages are messages distributed to individual clients. They don't mutate regular group state (i.e, as
+/// defined by RFC 9420). That is because many or most clients in a group will never receive those messages, and will
+/// not have the appropriate cryptographic state to decrypt the messages.
+///
+/// Any feature using a targeted, transient or transient targeted message MUST specify why the compared to MLS
+/// application messages lower guarantees are acceptable and/or how they are mitigated.
```
istankovic: I think we already had one instance of
> ... why the compared to MLS application messages lower guarantees...

that was fixed, this one should be fixed too in the same manner.

crypto/src/mls/core_crypto_message/sender_nonce.rs
SimonThormeyer: Please see 361cd931f1191c4fe3212e3f758dae069b3158a9.
Is this better? The cost of this is that we have to use `u32` on `TargetedMessage` instead of the newtype and use the `From<&SenderNonce> for u32` impl.

coriolinus: Making a nonce `Copy` makes sense to me; it's in line with other numeric types. The real trick with a nonce is that you only ever acquire one in the first place from the nonce factory, and the factory never redistributes one.

Making it `pub(crate)` instead of fully public seems fine, I don't have a strong feeling about using the newtype vs `u32` directly. But I note that if it's important to have a non-cloneable type, then we should publish and use the newtype, not the integer.

SimonThormeyer: Decided to revert 361cd931f1191c4fe3212e3f758dae069b3158a9

crypto/src/mls/conversation/config.rs
```diff
@@ -35,6 +35,7 @@ pub struct ConversationConfiguration {
 }
 
 impl ConversationConfiguration {
+    /// This a) passed to openmls when creating a group and b) for transient and targeted messages padding.
```
istankovic: Nit: should probably be "This is a) ..."

crypto/src/mls/core_crypto_message/mod.rs
istankovic: > Note that `targeted` is a child module of this and the protocol version will be used by `transient` also.
> How about `CoreCryptoMessagesProtocolVersion` for a name?

Honestly, I don't think `CoreCryptoMessage` is a good name.

We're already in `core-crypto` and it feels unnecessary to repeat it in module names. I would prefer the name to be more specific. 

Right now this is our impl of the targeted messages feature. Once the other draft is finalized, it is conceivable that we switch to it, at which point it wouldn't make sense to talk about `CoreCryptoMessage`.

So I'm thinking, instead of
```
crypto/src/mls/
├── conversation
├── core_crypto_message
├── credential
└── session
```
have something like
```
crypto/src/mls/
├── conversation
├── credential
├── session
└── targeted   [*]
```

I'd put both targeted and transient impls under that `[*]` submodule since they are (somewhat) related and I don't see a big issue in having them both there.

And in fact, the many `pub(crate)` items that we now have should probably never be accessible outside of that submodule -- most of them are implementation details of T&T messages, including the protocol version above, so it makes no sense for them to be available to anything outside that narrow module.

coriolinus: > Honestly, I don't think `CoreCryptoMessage` is a good name.

I agree! 

>  targeted and transient impls under that [*] submodule

So name the module either `transient_and_targeted` or `tnt` or something? Either would be better than `core_crypto_message`.

> many pub(crate) items that we now have should probably never be accessible outside of that submodule

Also a good point.

SimonThormeyer: Agreed to make `tnt` a submodule of `conversation`, so that all `tnt` types and other details can stay module-internal.

crypto/src/mls/conversation/mutable/tnt/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+            body: TntMessageBody::Targeted(targeted),
+        }
+    }
+
+    pub(crate) fn new_transient_targeted(targeted: TargetedMessage) -> Self {
+        Self {
+            protocol_version: ProtocolVersion::V1,
+            body: TntMessageBody::TransientTargeted(targeted),
+        }
+    }
```
istankovic, line 71: None of these `fn`s need to be `pub(crate)`.

crypto/src/mls/conversation/mutable/tnt/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+    fn unsigned_payload(&self) -> Result<Vec<u8>, tls_codec::Error> {
+        self.tls_serialize_detached()
+    }
+
+    fn label(&self) -> &str {
+        match self.body {
+            TntMessageBody::Transient(_) => todo!(),
+            TntMessageBody::Targeted(_) => TargetedMessage::SIGN_LABEL_PERSISTED,
+            TntMessageBody::TransientTargeted(_) => TargetedMessage::SIGN_LABEL_TRANSIENT,
+        }
```
istankovic, line 93: Note that the actual label used for signing will be the concatenation of 
```rust
const LABEL_PREFIX: &str = "MLS 1.0 ";
```
(defined by OpenMLS/MLS) and the label we provide.

crypto/src/mls/conversation/mutable/tnt/sender_nonce.rs
```diff
@@ -0,0 +1,16 @@
+use tls_codec::{TlsDeserialize, TlsSerialize, TlsSize};
+
+use crate::mls::conversation::{Error, Result};
+
+/// A shared nonce type for Transient and Targeted Messages.
+#[derive(Clone, Default, Copy, TlsSize, TlsSerialize, TlsDeserialize, derive_more::Into, derive_more::From)]
+#[repr(transparent)]
+pub(crate) struct SenderNonce(u32);
```
istankovic: I would really like this to be renamed, in order to make it more obvious what this is about.
How about any of the following?
- `TntSenderNonce`
- `TntNonce`

crypto/src/mls/conversation/mutable/tnt/targeted.rs
```diff
@@ -0,0 +1,73 @@
...
+pub(super) struct TargetedMessage {
+    nonce: SenderNonce,
+    sender: LeafNodeIndex,
+    recipient: LeafNodeIndex,
+    epoch: GroupEpoch,
+    group_id: GroupId,
+    /// Ecrypted with the receipient's HPKE public key; authenticated with epoch secret as PSK.
+    /// The encrypted payload will typically be GenericMessage protobuf.
+    /// Must have padding like the openmls ciphertexts.
+    payload: HpkeCiphertext,
```
istankovic: I would rephrase this doc comment to be more precise regarding the auth, PSK etc.:
```rust
    /// Encrypted with the receipient's HPKE public key; authenticated with a PSK derived
    /// from the epoch secret, guaranteeing that the sender was a member of the group
    /// during the epoch. An additional authentication layer is provided by the signature over
    /// the entire message.
    /// Before encryption, the HPKE plaintext is padded like in OpenMLS.
```

crypto/src/mls/conversation/mutable/tnt/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+use self::targeted::{PskId, TargetedMessage, TargetedMessageContext};
+
+/// The version of the Transient and Targeted Messages protocol.
+#[derive(TlsSize, TlsSerialize, TlsDeserialize)]
+#[repr(transparent)]
+struct ProtocolVersion(u16);
+
+impl ProtocolVersion {
+    /// Transient and Targeted Messages Version 1
+    pub(crate) const V1: Self = Self(1);
```
istankovic: This does not need to be `pub(crate)`.

BTW, this looks so much better now!

crypto/src/mls/conversation/mutable/tnt/targeted.rs
```diff
@@ -0,0 +1,73 @@
...
+pub(super) struct TargetedMessageContext {
+    protocol_version: ProtocolVersion,
+    policy: TargetedMessagePolicy,
+    sender: LeafNodeIndex,
+    recipient: LeafNodeIndex,
+    group_context: GroupContext,
+}
+
+impl TargetedMessageContext {
+    pub(super) fn new(
```
istankovic: The fact that we need to have `pub(super)` here and on `TargetedMessageContext` leads me to think that the item placement in modules is less than ideal.

For example, if we moved everything in this module to `encrypt_targeted.rs`, we wouldn't need those `pub`s at all, these types and functions could all be private. (Same in the other direction, which I would prefer: all items in `encrypted.rs` moved to `targeted.rs`.)

That way we wouldn't need to thread the various items and adjust visibility across the whole `tnt` module:
- `targeted.rs` would have all items related to targeted (and transient targeted) messages
- (future) `transient.rs` would have all items related to transient messages
- `mod.rs` would just have the common items  (we might even move `sender_nonce.rs` items there since there's so little in it)

crypto/src/mls/conversation/mutable/group_mutation.rs
```diff
@@ -25,6 +25,18 @@ impl ConversationMut {
...
+            let mls_group = state.mls_group_mut();
+            operation(tx, mls_group, id).await
+        })
+        .await
+    }
+
+    /// Like [Self::mutate_group], but also allows mutating the sender nonce via [MlsGroupState].
+    pub(super) async fn mutate_group_and_sender_nonce<T>(
+        &mut self,
+        operation: impl AsyncFnOnce(&Transaction, &mut MlsGroupState, &ConversationIdRef) -> Result<T>,
```
istankovic, line 27: I think I understand why you added this additional function: there are places that need to only mutate the "inner" MLS group state, not the whole state that includes the inner + nonce.

At a first glance, I feel it would be better to have just one `mutate_<x>` function that gives you the entire state and then the operation implementer gets the state, from which they can trivially access the group.
That way we wouldn't need to think about which one to call and the nonce handling would be implicit and hidden from the callers of this function. Put another way: I don't think the caller of this should care about nonces at all.

The downside, of course, would be a tiny bit more work in operation implementers, but that should be an acceptable tradeoff, accessing the inner group is trivial.

istankovic:
We still have `CoreCryptoMessage` in a few places:
```
crypto/src/mls/conversation/mutable/tnt/targeted.rs:70:    pub(super) const SIGN_LABEL_PERSISTED: &str = "CoreCryptoMessageTBS-Persisted-Targeted";
crypto/src/mls/conversation/mutable/tnt/targeted.rs:71:    pub(super) const SIGN_LABEL_TRANSIENT: &str = "CoreCryptoMessageTBS-Transient-Targeted";
crypto/src/mls/conversation/mutable/tnt/encrypt_targeted.rs:69:            .map_err(OpenMlsError::wrap("signing CoreCryptoMessageTBS"))?;
crypto/src/mls/conversation/mutable/tnt/encrypt_targeted.rs:73:            .map_err(Error::tls_serialize("CoreCryptoMessage: targeted"))
```

crypto/src/mls/conversation/mutable/tnt/targeted.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// The encrypted payload will typically be GenericMessage protobuf.
+    /// Must have padding like the openmls ciphertexts.
+    payload: HpkeCiphertext,
+}
+
+impl TargetedMessage {
+    pub(super) const PADDING_SIZE: usize = ConversationConfiguration::PADDING_SIZE;
+    pub(super) const SIGN_LABEL_PERSISTED: &str = "CoreCryptoMessageTBS-Persisted-Targeted";
+    pub(super) const SIGN_LABEL_TRANSIENT: &str = "CoreCryptoMessageTBS-Transient-Targeted";
+    pub(super) const PSK_LABEL: &str = "CoreCrypto TargetedMessage Psk";
```
istankovic: These labels should also encode the version of the TNT protocol/structs.

SimonThormeyer: fa2f53ab6ca003b1fa115df4d68a65d09717bef2

crypto/src/mls/conversation/mutable/tnt/sender_nonce.rs
istankovic: After some more discussion with Emil and confusion about this type, I am of the opinion that we should avoid using the term `nonce` here as it is misleading.

I propose to switch to a `count(er)`-based terminology, e.g.
- `MessageCount`
- `MessageCounter`

crypto/src/mls/conversation/mutable/tnt/mod.rs
SimonThormeyer: Updated the RFC accordingly

crypto/src/mls/conversation/mutable/tnt/targeted.rs
SimonThormeyer: Decided to leave the sender nonce module as a separate module for now, as it might grow when we implement decryption.

istankovic (approved):
Looking good! 🍷

PR #2422, opened 2026-08-03
chore(keystore): rm `E2eiRefreshToken` [WPB-27716]

coriolinus:
It doesn't even have a database table anymore (V15); we don't need the entity.


SimonThormeyer (approved):
Thanks!

PR #2423, opened 2026-08-03
refactor: fix proteus auto prekey selection [WPB-27717]

coriolinus:
Get rid of the cached serialization of the id, use a non-terrible algorithm to find the next free id.

Note: this used to be [much simpler](https://github.com/wireapp/core-crypto/commit/e22432c99ef866661bf51f2493cd1062410d91a5), but it turns out that there are two problems with that:

1. Proteus uses `u16::MAX` as the last resort prekey id, so just selecting the max in the DB will run us out of available space immediately.
2. Deleting a prekey has to free its id for reuse, and there's a test case enforcing that behavior.


keystore/src/entities/proteus/prekey.rs
```diff
@@ -19,22 +19,48 @@ impl ProteusPrekey {
...
+            },
+            |row| row.get::<_, Option<u16>>(0),
+        )?
+        .ok_or(CryptoKeystoreError::NoFreePrekeyId)
+    }
+
+    pub async fn get_free_id(tx: &Transaction) -> CryptoKeystoreResult<u16> {
+        // have to check the transaction to see what's stored there first
+        tx.free_proteus_prekey_id().await
     }
```
fewerner, line 64: Would it make sense if  `new_prekey_auto` called `tx.free_proteus_prekey_id` direclty  and we drop `get_free_id` completely?

I don't think this single line function adds value.

fewerner (approved):
Great improvement and nice comments. 

Just one question:

keystore/src/entities/proteus/prekey.rs
coriolinus: I think this function does add value. Weak rationale: keeping it leads to a smaller PR and less code churn. Stronger rationale: `ProteusPrekey::get_free_id(tx)` is just a better calling pattern for external callers than `tx.free_proteus_prekey_id()`, even though one delegates directly to the other.

PR #2424, opened 2026-08-03
refactor(keystore): change spelling to `key_package` [WPB-24897]

fewerner:
This PR fixes the spelling of key package for remaining occurrences in keystore crate.
We fixed it in all other crates previously and waited with this PR for db unification as it involves a migration.


SimonThormeyer (approved):
Thanks!

PR #2425, opened 2026-08-03
feat(tnt): add `ConversationMut::decrypt_targeted()` [WPB-27214]

SimonThormeyer:


keystore/src/entities/mls/tnt_secret.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+impl EntityDatabaseMutation for TntSecret {
+    type AutoGeneratedFields = ();
+
+    fn save(&self, tx: &rusqlite::Transaction) -> crate::CryptoKeystoreResult<()> {
+        let mut stmt = tx.prepare_cached(&format!(
+            "INSERT OR REPLACE INTO {}
+                (conversation_id, epoch, hpke_private_key, group_context, targeted_message_psk)
+             VALUES (?, ?, ?, ?, ?)",
```
coriolinus: This is complicated enough to justify the use of named params, I think.

keystore/src/entities/mls/tnt_secret.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub epoch: u64,
+    #[sensitive]
+    pub hpke_private_key: Vec<u8>,
+    pub group_context: Vec<u8>,
+    #[sensitive]
+    pub targeted_message_psk: Vec<u8>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, derive_more::Constructor)]
+pub struct TntSecretPk {
```
coriolinus, line 28: This seems like a good opportunity to have `TntSecretPkRef<'a>` containing an `&'a ConversationRef`, so that you can impl `EntityGetBorrowed` and `EntityDeleteBorrowed`.

keystore/src/entities/mls/targeted_message_rx_counter.rs
```diff
@@ -0,0 +1,130 @@
...
+        load_all_helper(conn, Self::from_row)
+    }
+}
+
+impl EntityDatabaseMutation for TargetedMessageRxCounter {
+    type AutoGeneratedFields = ();
+
+    fn save(&self, tx: &rusqlite::Transaction) -> crate::CryptoKeystoreResult<()> {
+        let mut stmt = tx.prepare_cached(&format!(
+            "INSERT OR REPLACE INTO {} (conversation_id, sender, epoch, count) VALUES (?, ?, ?, ?)",
```
coriolinus: Should probably use named params here also.

crypto/src/mls/conversation/mutable/group_mutation.rs
```diff
@@ -36,6 +40,21 @@ impl ConversationMut {
         let Conversation { group, id, .. } = &*self.inner;
         let mut group = group.write().await;
         let epoch_before_operation = group.epoch();
+
+        // Save the tnt secret. We need to do this exactly once per epoch.
+        let tnt_secret_key = TntSecretPk::new(id.as_ref().to_vec(), epoch_before_operation.as_u64());
```
coriolinus: The needed `to_vec()` here is why I'm suggesting `EntityGetBorrowed` for `TntSecret`.

crypto/src/mls/conversation/mutable/group_mutation.rs
```diff
@@ -44,6 +63,24 @@ impl ConversationMut {
 
         if epoch_before_operation < group.epoch() {
             group.reset_targeted_message_tx_counters(tx).await;
+
+            let oldest_retained_epoch = group.epoch().as_u64().saturating_sub(MAX_PAST_EPOCHS as u64);
+            let conversation_id = DbConversationId::from(id.as_ref().to_vec());
```
coriolinus: I bet we could also invent a borrowed form of `ConversationId` that we could implement searching for.

crypto/src/mls/conversation/mutable/group_mutation.rs
```diff
@@ -44,6 +63,24 @@ impl ConversationMut {
...
+            // This is not a lot of data: just `MAX_PAST_EPOCHS` records. In practice, we could in most cases just
+            // delete the single one with the oldest epoch. However, to avoid any assumptions about the data, let's just
+            // check each of the `MAX_PAST_EPOCHS` records.
+            for secret in secrets {
+                if secret.epoch < oldest_retained_epoch {
+                    tx.remove::<TntSecret>(&TntSecretPk::new(secret.conversation_id.clone(), secret.epoch))
+                        .await
+                        .map_err(KeystoreError::wrap("removing old tnt secret"))?;
+                }
+            }
```
coriolinus: This is acceptable as a first draft but I'd rather see an `AnyEpochOlderThan(pub u64)`  newtype with a `DeletableBySearchKey` impl which ultimately delegates down to `DELETE FROM tnt_secrets WHERE epoch < ?`.

crypto/src/mls/conversation/mutable/tnt/targeted/decrypt.rs
```diff
@@ -0,0 +1,257 @@
...
+        let message_epoch = message.epoch.as_u64();
+        let group_epoch = mls_group.epoch().as_u64();
+        if message_epoch == group_epoch + 1 {
+            return Err(Error::BufferedFutureMessage { message_epoch });
+        } else if message_epoch > group_epoch + 1 {
+            return Err(Error::UnbufferedFarFutureMessage);
+        }
+
+        let database = self.database().await?;
+        let counter_pk = TargetedMessageRxCounterPk::new(self.id.to_bytes(), message.sender().u32(), group_epoch);
```
coriolinus: Borrowed form! 😄

crypto/src/mls/conversation/mutable/tnt/targeted/decrypt.rs
```diff
@@ -0,0 +1,257 @@
...
+            return Ok((context_data, decryption_key));
+        }
+
+        let epoch_distance = mls_group.epoch().as_u64() - message.epoch.as_u64();
+        if epoch_distance > MAX_PAST_EPOCHS as u64 {
+            return Err(Error::MessageEpochTooOld);
+        }
+
+        let database = self.database().await?;
+        let key = TntSecretPk::new(mls_group.group_id().to_vec(), message.epoch.as_u64());
```
coriolinus: Borrowed form.

crypto/src/mls/conversation/mutable/tnt/targeted/decrypt.rs
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    // Ideally, openmls would offer this API for us, but it doesn't. So we load the data we need from the database.
+    async fn load_decryption_key(&self, mls_group: &MlsGroup) -> Result<HpkePrivateKey> {
+        let epoch_keypair_id = [
+            mls_group.group_id().as_slice(),
+            &mls_group.own_leaf_index().u32().to_be_bytes(),
+            &mls_group.epoch().as_u64().to_be_bytes(),
+        ]
+        .concat();
```
coriolinus, line 215: This seems like a useful kind of thing to keep track of instead of just doing on an ad-hoc basis. It's out of scope for this PR unless you're into it, but we could add a `StoredEpochEncryptonKeypairPk(Vec<u8>)` newtype with a constructor which codifies this concatenation (as well as `From` + `Into` of course).

crypto/src/mls/conversation/mutable/decrypt/mod.rs
```diff
@@ -174,12 +180,21 @@ impl ConversationMut {
...
-
-        let decrypt_message_result = self
-            .decrypt_message_inner(mls_message_in, RecursionPolicy::AsNecessary)
-            .await;
+        let message = message.as_ref();
+        // Bytes 0 and 1 are the protocol version, bytes 2 and 3 are the wire format.
+        let wire_format =
+            u16::tls_deserialize_exact(&message[2..4]).map_err(TlsCodecError::deserialize("u16 (wire format)"))?;
+
+        let decrypt_message_result = if TntWireFormat::all().contains(&wire_format) {
```
coriolinus: I think a `match` might be easier to read, but then `TntWireFormat` needs to be an associated const. But then it looks like this:

```rust
match wire_format {
  TntWireFormat::ALL => { // tnt path }
  _ => { // mls path }
}
```

crypto-ffi/src/decrypted_message.rs
```diff
@@ -37,6 +37,24 @@ pub enum DecryptedMessage {
...
+    },
+    /// The decrypted message is a persisted targeted message.
+    PersistedTargeted {
+        /// Decrypted text message.
+        plaintext: Vec<u8>,
+        /// The sender's `ClientId`.
+        sender_client_id: Arc<ClientId>,
+        /// Identity claims present in the sender credential.
+        identity: WireIdentity,
+    },
```
coriolinus, line 57: The contents here are all identical to the contents of `ApplicationMessage`. What about a factoring like this?

```rust
struct PlaintextMessage {
  plaintext: Vec<u8>,
  sender_client_id: Arc<ClientId>,
  identity: WireIdentity,
}

enum DecryptedMessage {
  ApplicationMessage(PlaintextMessage),
  TransientTargeted(PlaintextMessage),
  PersistedTargeted(PlaintextMessage),
}
```

This design also lends itself to helpers for cases where the caller doesn't actually care about the nature of the message, such as

```rust
impl DecryptedMessage {
  fn into_plaintext_message(self) -> Result<PlaintextMessage, Self> { ... }
}
```

coriolinus (approved):
This is great work! And good job on the borrowed trait refactoring. I have more ideas from this readthrough, but nothing blocking.

crypto-ffi/src/decrypted_message.rs
SimonThormeyer: Unfortunately, tuple-struct-like enum variants lead to very unidiomatic access patterns in the target languages, I don't remember what exactly, but I think the generators create class members with names such as `0` and `1`.

That's why unfortunately I think we have to keep this layout.

crypto/src/mls/conversation/mutable/decrypt/mod.rs
SimonThormeyer: Unfortunately, rustc doesn't let me match a u16 directly against a `std::ops::RangeInclusive<u16>`. So the match would have to look like this:

```rust
match wire_format {
    wire_format if TntWireFormat::ALL.contains(&wire_format) => {
        // tnt path        
    }
    _ => {
        // mls path
    }
}
```
Even though it's not as pretty, I think it's still clearer than `if/else`.

crypto/src/mls/conversation/mutable/tnt/targeted/decrypt.rs
SimonThormeyer: I'm going to create a follow-up item.

crypto/src/mls/conversation/mutable/decrypt/mod.rs
coriolinus: Agree. But it's surprising that we can't match it directly, since we obviously can match a literal range. I wonder if this will change once they finish the range refactoring that's been going on recently.

PR #2427, opened 2026-08-03
refactor(keystore): remove `parent_id` from `PersistedMlsGroup`

fewerner:
This PR removes `parent_id` from `PersistedMlsGroup` and unused functions that made use of it.


keystore/tests/searchable_entities.rs
```diff
@@ -21,151 +21,6 @@ fn random_bytes(len: impl SampleRange<usize>) -> Vec<u8> {
...
-    fn get_search_key(entity: &PersistedMlsGroup) -> ParentGroupId<'_> {
-        entity
-            .parent_id
-            .as_deref()
-            .expect("all entities in these tests have a parent id")
-            .into()
-    }
-
-    #[apply(all_storage_types)]
-    async fn search_can_find_an_entity(context: KeystoreTestContext) {
```
fewerner, line 52: Not sure if the removed tests in ac9431d meant to generally test searchable entities or just `PersistedMlsGroup`. We have the same tests for `StoredCredential`. So I thought it's fine to remove these.

coriolinus (approved):
Looks good! You could probably merge the two sql commits IMO, but that's not important.

PR #2431, opened 2026-08-04
ci: install puppeteer from js bindings

fewerner:
This ensures we install the version compatible with our puppeteer dependency.


SimonThormeyer (commented):
Thanks for investigating this!

SimonThormeyer (approved):

PR #2438, opened 2026-08-06
test: add tests for proteus session regeneration bug [WPB-27807]

coriolinus:
This is an unusual "fix" PR, because the fix is already in the history; we just haven't released it. So instead we add some tests.

To demonstrate, apply this patch, which effectively undoes deb32514d4:

```patch
diff --git a/crypto-macros/src/entity_derive/derive_impl.rs b/crypto-macros/src/entity_derive/derive_impl.rs
index 2002211991..6028875f22 100644
--- a/crypto-macros/src/entity_derive/derive_impl.rs
+++ b/crypto-macros/src/entity_derive/derive_impl.rs
@@ -133,12 +133,8 @@ impl Entity {
                 where
                     for<'pk> &'pk Self::BorrowedPrimaryKey: crate::traits::KeyType,
                 {
-                    // Bind the borrowed key directly, exactly as the insert generated by
-                    // `store_expression` binds the id field: `&str` binds as `TEXT` and `&[u8]` as
-                    // `BLOB`, so the lookup and the insert cannot disagree about storage class.
-                    // Converting the key here is what once made a string key bind as a `BLOB`,
-                    // which silently matched zero rows, since SQLite never compares a `BLOB` equal
-                    // to a `TEXT` value and column affinity does not coerce a `BLOB` argument.
+                    let key = <&Self::BorrowedPrimaryKey as crate::traits::KeyType>::bytes(&key);
+                    let key = key.as_ref();
                     crate::entities::helpers::get_helper::<Self, _>(conn, #pk_column_name, key, |row| {
                         Ok(Self {
                             #( #field_assignments, )*
@@ -189,7 +185,8 @@ impl Entity {
                 where
                     for<'pk> &'pk <Self as crate::traits::BorrowPrimaryKey>::BorrowedPrimaryKey: crate::traits::KeyType,
                 {
-                    // Bind the borrowed key directly, for the same reason as in `get_borrowed`.
+                    let key = <&<Self as crate::traits::BorrowPrimaryKey>::BorrowedPrimaryKey as crate::traits::KeyType>::bytes(&id);
+                    let key = key.as_ref();
                     crate::entities::helpers::delete_helper::<Self>(tx, #id_column_name, id)
                 }
             }
```

The test will fail, in that case.

Note that 02b1bc09ea4fede5bb720a4054567171a09fb0cf got rid of `bind_key` entirely; it turned out to be pointless.


SimonThormeyer (commented):
Could you please mention `a0d08ad18f9e835b83791b8ea92246b9a5a8c6e3` in either the message of first commit on the branch or in the message of `598a625e5645aac046f61a69cf1dabd14cfd1ac1`? Or in the PR description?

crypto/src/proteus/session_cache.rs
```diff
@@ -88,3 +88,79 @@ impl std::fmt::Debug for ProteusSessionCache {
...
+
+    /// A session persisted by one [`ProteusCentral`] must be loaded by the next one over the same
+    /// database, never silently replaced by a freshly established session.
+    #[macro_rules_attribute::apply(smol_macros::test)]
+    async fn persisted_session_is_loaded_rather_than_regenerated() {
+        let (path, _db_file) = tmp_db_file();
+        let key = DatabaseKey::generate();
+        let session_id = uuid::Uuid::new_v4().hyphenated().to_string();
+
+        let mut bob = CryptoboxLike::init();
```
SimonThormeyer, line 109: Is there a particular reason why we instantiate `bob` as `CryptoboxLike` and not `ProteusCentral`?

SimonThormeyer (approved):
Great investigation and summary! Thank you!

crypto/src/proteus/session_cache.rs
coriolinus: Just that creating a `ProteusCentral` requires a transaction, whereas `CryptoboxLike` is intentionally in-memory and doesn't touch a real database at all.

PR #2441, opened 2026-08-07
feat: add multiple trust anchors [WPB-25632]

fewerner:
This PR allows to add multiple trust anchors to the PKI environment, getting existing trust anchors, and removing trust anchors from the PKI environment.


keystore/src/migrations.rs
```diff
@@ -71,6 +71,33 @@ pub(crate) struct LegacyStoredKeypackage {
     pub keypackage: Vec<u8>,
 }
 
+#[cfg(target_os = "unknown")]
+#[derive(core_crypto_macros::Debug, Clone, PartialEq, Eq, Zeroize, serde::Serialize, serde::Deserialize)]
+#[zeroize(drop)]
+pub struct LegacyE2eiAcmeCA {
```
coriolinus: Shouldn't all this live inside the `legacy/entities/mls` folder?

keystore/src/entities/mls/e2ei_acme_ca.rs
```diff
@@ -1,19 +1,19 @@
...
+    Clone,
+    PartialEq,
+    Eq,
+    Zeroize,
+    serde::Serialize,
+    serde::Deserialize,
+    core_crypto_macros::Entity,
+)]
 #[zeroize(drop)]
+#[entity(table_name = "e2e_acme_ca")]
```
coriolinus: This table name doesn't match the old one: `e2ei_acme_ca`. Intentional?

Cargo.toml
```diff
@@ -78,6 +78,7 @@ uniffi = "0.29.5"
 url = "2.5"
 uuid = "1.23"
 x509-cert = "0.2"
+spki = { version = "0.7", features = ["pem", "fingerprint"] }
```
coriolinus: This position is unexpected: it's neither alphabetical nor appended to the end of the list. Why put it precisely here?

keystore/src/connection/migrations/meta_migrations/v28.rs
```diff
@@ -0,0 +1,37 @@
+use x509_cert::der::Decode;
+
+use crate::{CryptoKeystoreError, CryptoKeystoreResult};
+
+pub(crate) const VERSION: i32 = 28;
+
+pub(crate) fn meta_migration(conn: &mut rusqlite::Connection) -> CryptoKeystoreResult<()> {
+    let tx = conn.transaction()?;
+
+    use rusqlite::OptionalExtension;
```
coriolinus: Why put this here inline instead of up with the imports?

e2e-identity/src/pki_env/mod.rs
```diff
@@ -204,40 +204,28 @@ impl PkiEnvironment {
...
     }
 
-    /// Remove the trust anchor with serial number `serial_number` from the PKI environment.
+    /// Remove the trust anchor from the PKI environment.
     ///
     /// Note that any certificates relying on the removed trust anchor may no longer
     /// validate.
-    pub async fn remove_trust_anchor(&self, tx: &Transaction, serial_number: &[u8]) -> Result<()> {
-        let mut guard = self.rjt_pki_env.lock().await;
+    pub async fn remove_trust_anchor(&self, tx: &Transaction, fingerprint: &Vec<u8>) -> Result<()> {
```
coriolinus: Why `&Vec<u8>`? `&[u8]` is more flexible.

cc-book/src/release_notes.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- `PkiEnvironment` now supports adding multiple trust anchors. Previously, `addTrustAnchor` replaced any existing trust
+  anchor. Get all added trust anchor certificates via `getTrustAnchor`. Remove a trustanchor via `removeTrustAnchor` by
```
istankovic: Typos:
 `getTrustAnchor` -> `getTrustAnchors`
`Remove a trustanchor` -> `Remove a trust anchor`

e2e-identity/src/pki_env/mod.rs
```diff
@@ -372,7 +372,15 @@ LOVS/gxNk618+PKA2bYq67MZQXCYGgk=
...
         pki_env
-            .remove_trust_anchor(&tx, certs[0].tbs_certificate.serial_number.as_bytes())
+            .remove_trust_anchor(
+                &tx,
+                &certs[0]
+                    .tbs_certificate
+                    .subject_public_key_info
+                    .fingerprint_bytes()
+                    .expect("Getting fingerprint of subject plublic key info")
+                    .to_vec(),
```
coriolinus: I wouldn't expect us to need to allocate here.

e2e-identity/tests/e2e.rs
```diff
@@ -315,7 +315,15 @@ async fn should_fail_without_trust_anchor(test_env: TestEnvironment) {
...
     pki_env
-        .remove_trust_anchor(&tx, certs[0].tbs_certificate.serial_number.as_bytes())
+        .remove_trust_anchor(
+            &tx,
+            &certs[0]
+                .tbs_certificate
+                .subject_public_key_info
+                .fingerprint_bytes()
+                .expect("Getting fingerprint of subject plublic key info")
+                .to_vec(),
```
coriolinus: ... or here.

keystore/src/connection/migrations/mod.rs
```diff
@@ -260,4 +262,75 @@ pub(crate) mod test {
             );
         });
     }
+    #[test]
```
istankovic, line 265: Nit: missing empty line above `#[test]`

coriolinus (commented):
Lots of nits I'd like to see addressed, but the big picture is nice.

keystore/src/entities/mls/x509_trust_anchor.rs
```diff
@@ -12,7 +12,7 @@ use zeroize::Zeroize;
 )]
 #[zeroize(drop)]
 #[entity(table_name = "x509_trust_anchor")]
-pub struct E2eiAcmeCA {
+pub struct X509TrustAnchor {
```
istankovic, line 15: Thanks for this!

I see we still have `E2eiCrl` and `E2eiIntermediateCert` entities which could use a rename too -- would you mind creating another item for that so we don't forget? (It's not the scope of this PR, so I wouldn't complicate this PR further, but it would be nice if we could do it sooner since we're already here...)

keystore/src/connection/migrations/mod.rs
fewerner: hm, very annoying that auto-format doesn't help me here.

keystore/src/connection/migrations/mod.rs
```diff
@@ -127,7 +127,9 @@ pub(crate) mod test {
     use std::io::Write;
 
     use openmls::prelude::Ciphersuite;
+    use spki::der::DecodePem;
     use tempfile::NamedTempFile;
+    use x509_cert::der::Encode;
```
istankovic: Trait imports should be done `as _`.

crypto-ffi/src/pki_env.rs
```diff
@@ -3,7 +3,7 @@ use std::{fmt, sync::Arc};
 #[cfg(feature = "cancellable-transactions")]
 use futures_util::{FutureExt as _, TryFutureExt as _};
 use wire_e2e_identity::pki_env;
-use x509_cert::der::DecodePem as _;
+use x509_cert::der::{DecodePem as _, EncodePem};
```
istankovic: Trait imports should be done `as _`.

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
```diff
@@ -772,16 +788,38 @@ final class WireCoreCryptoTests: XCTestCase {
...
+            stride(from: 0, to: pem2Fingerprint.count, by: 2).map { index in
+                UInt8(pem2Fingerprint.dropFirst(index).prefix(2), radix: 16)!
+            })
+        let database = try await newDatabase()
+        let pkiEnvironment = try await PkiEnvironment(
+            hooks: MockPkiEnvironmentHooks(), database: database)
+
+        try await pkiEnvironment.addTrustAnchor(certPem: Self.testCaPem)
+        try await pkiEnvironment.addTrustAnchor(certPem: Self.testCaPem2)
+        try await pkiEnvironment.removeTrustAnchor(fingerprint: fingerprintBytes)
```
istankovic, line 818: What happens if a trust anchor with a given fingerprint does not exist?

Cargo.toml
fewerner: The Certificate type from which we get the fingerprint comes from x509-cert. I wasn't aware that we'd do append or alphabetical order. It might be good to just enforce order through formatting? Personally, I don't care about this, but might be good to give everyone the same view through tooling.

keystore/src/connection/migrations/meta_migrations/v28.rs
fewerner: That's just wrong

crypto-ffi/bindings/swift/WireCoreCrypto/WireCoreCryptoTests/WireCoreCryptoTests.swift
fewerner: It just suceeds, the db doesn't error so we also don't. We then load all remaining TAs and create fresh source.
Do we want to throw an error here and return early?

keystore/src/migrations.rs
fewerner: I moved the trait impls but the type is also referenced in
`keystore/src/connection/idb_migration/mod.rs`

coriolinus: Even in that case IMO it should live in the entities module with a `pub(in crate::connection::idb_migration)` visibility.

e2e-identity/src/pki_env/mod.rs
```diff
@@ -177,6 +179,11 @@ impl PkiEnvironment {
             .map_err(Error::Spki)?
             .to_vec();
 
+        // A trust anchor can only be added once
+        if tx.get::<X509TrustAnchor>(&fingerprint).await?.is_some() {
```
coriolinus, line 183: Strictly speaking we should be using `tx.exists::<X509TrustAnchor>()` rather than counting; I don't know if we're taking advantage of the potential, but the potential does exist to use a more efficient implementation for `exists` calls.

e2e-identity/src/pki_env/mod.rs
coriolinus: It looks like this function's changed its input from `serial_number` to `fingerprint`, which are distinct fields with the same type. That feels dangerous to me; we have no help from the compiler to ensure the callers change their call patterns appropriately. 

Maybe incorporate a function rename (i.e. `remove_trust_anchor_by_fingerprint`) in this change, even if the rename gets reverted later in the sequence, so we know for a fact that we caught all callers?

fewerner: dd9b374 proves that. I will remove it once this is approved.

coriolinus (approved):

PR #2444, opened 2026-08-10
chore: use workspace for more dependencies

istankovic:
This makes it easier to maintain deps and minimises the chance of mismatch among crates.


fewerner (approved):
LGTM! thanks for doing this!

PR #2445, opened 2026-08-10
refactor: give `MlsPendingMessage` a primary key

coriolinus:
The absence of a primary for this type (only!) has been a continuing source of special-case behavior and overall weirdness. But at this point it's very easy to just give it a primary key; this does that.


coriolinus:
I've discovered some issues with this PR; don't review it yet.

keystore/src/connection/composite-schema.sql
```diff
@@ -87,3 +81,12 @@ CREATE TABLE "mls_pending_groups" (
   cfg BLOB,
   parent_id BLOB
 );
+
+CREATE TABLE "mls_pending_messages" (
+  hash_sha256 BLOB PRIMARY KEY NOT NULL,  -- hash of the concatenation of (conversation_id, message)
+  conversation_id BLOB NOT NULL,
```
fewerner, line 90: We implement SearchableEntity for conversation_id. We could think about indexing this column.

keystore/src/transaction/mod.rs
```diff
@@ -134,6 +134,10 @@ impl UniqueArc<Transaction> {
         let mut conn = database.conn().await;
         let tx = conn.transaction_with_behavior(TransactionBehavior::Immediate)?;
 
+        for conversation_id in cleared_conversations.iter() {
+            MlsPendingMessage::delete_all_matching(&tx, &conversation_id.as_slice().into())?;
+        }
+
         for entity in cache.values() {
```
fewerner, line 141: 6df8a70 should be folded into b5c3559

fewerner (commented):
Thanks for providing all the tests. Could you rebase your work on #2441 ? We have conflicts with the migrations.

keystore/src/connection/composite-schema.sql
coriolinus: We do! It's just after the table definition.

keystore/src/transaction/mod.rs
coriolinus: Done in the rebase.

keystore/src/connection/composite-schema.sql
```diff
@@ -90,3 +84,13 @@ CREATE TABLE x509_crls (
   distribution_point TEXT PRIMARY KEY NOT NULL,
   content BLOB NOT NULL
 );
+
+CREATE TABLE "mls_pending_messages" (
+  hash_sha256 BLOB PRIMARY KEY NOT NULL,  -- hash of the concatenation of (conversation_id, message)
```
fewerner, line 89: 8b3d0ff can be folded into 0d366ef2.

keystore/src/connection/composite-schema.sql
```diff
@@ -93,4 +93,3 @@ CREATE TABLE "mls_pending_messages" (
 );
 
 CREATE INDEX idx_mls_pending_messages_conversation_id ON mls_pending_messages(conversation_id);
-
```
fewerner: c62097c7 should be folded into 8b3d0ff or fold both into 0d366ef.

fewerner (approved):
This looks good to me.
Just 2 comments for cleaner history. Last should be done imo. Up to you if you want a separate commit for composite-schema.

PR #2450, opened 2026-08-11
refactor(keystore): rename remaining e2ei structs to x509

fewerner:
This PR renames remaining E2ei prefixes to X509.


keystore/src/connection/migrations/V30__rename_e2ei_to_x509.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TABLE x509_intermediate_certs (
+    ski_aki_pair TEXT PRIMARY KEY NOT NULL,
+    content BLOB NOT NULL
+);
+
+INSERT INTO x509_intermediate_certs (ski_aki_pair, content)
+SELECT ski_aki_pair, content
+FROM e2ei_intermediate_certs;
+
+DROP TABLE e2ei_intermediate_certs;
```
coriolinus, line 10: For both of these: why not just `ALTER TABLE RENAME TO`? We have to do the create, copy, delete dance if we modify the primary keys; that's not a permitted operation with `ALTER TABLE`. But simply renaming it is totally fine.

coriolinus (approved):
Low-priority nit: it would be nice to have the sql migration first in the commit sequence. But it makes no difference to the functionality of the patch.

keystore/src/connection/migrations/V30__rename_e2ei_to_x509.sql
fewerner: We introduce new primary keys here. Before, these fields were UNIQUE. That's the only reason for the dance.

PR #2454, opened 2026-08-11
fix: actually migrate other entities with `Sha256Blob` PKs

coriolinus:
Demonstrate that we'd misused the `sha256hash` SQL function in old migrations, and add a meta-migration which fixes the problem.

### DIFF
--- a/keystore/src/connection/migrations/meta_migrations/v31.rs
+++ b/keystore/src/connection/migrations/meta_migrations/v31.rs
@@ -1,0 +1,137 @@
+//! Rewrite sha256 keys which earlier migrations backfilled as hex text.
+//!
+//! `sha256_blob` is named for its input, not its output: it returns [`crate::sha256`], which is a
+//! 64-character hex string. Every migration which used it to backfill a key column therefore wrote
+//! TEXT, while [`Sha256Hash`] binds the 32 raw digest bytes. SQLite never compares a blob equal to
+//! text, so those rows survive `load_all` but are invisible to `get` and `delete`, and so to the
+//! transaction cache as well.
+//!
+//! Any database old enough to have run V12 or V20 therefore holds a mix: hex keys on rows those
+//! migrations wrote, and raw keys on everything saved since.
+
+use std::collections::HashSet;
+
+use rusqlite::types::ValueRef;
+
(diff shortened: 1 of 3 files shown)


keystore/src/connection/migrations/meta_migrations/v31.rs
```diff
@@ -0,0 +1,137 @@
...
+#[derive(Debug, Default)]
+struct Outcome {
+    /// Rows whose key was already stored correctly, and so were left alone.
+    already_bytes: usize,
+    /// Rows whose hex key was rewritten as bytes.
+    rewritten: usize,
+    /// Rows dropped because a correctly-keyed row for the same key already existed.
+    superseded: usize,
+    /// Rows whose key was neither a 32-byte blob nor 64 hex characters, and so were left alone.
+    unrecognized: usize,
```
typfel, line 55: In practice we expect this to be zero and this exists since we can't rule it out?

coriolinus: Exactly.

keystore/src/connection/migrations/mod.rs
```diff
@@ -402,4 +441,265 @@ r9IJmL6kDQ==
...
+
+            assert_eq!(
+                StoredCredential::count(&conn).unwrap(),
+                1,
+                "the migration must not lose the credential"
+            );
+
+            let encoding = stored_key_encoding(&conn, "mls_credentials", "public_key_sha256");
+            assert!(
+                StoredCredential::get(&conn, &Sha256Hash::hash_from(PUBLIC_KEY))
```
typfel, line 588: Off topic for this PR, but why is `StoredCredential::get` taking a hash while the other getters are hashing internally?

typfel (approved):
Changes look good to me, just added some clarifying questions.

keystore/src/connection/migrations/mod.rs
coriolinus: Older entities assert that their primary key is `Vec<u8>`, and keep a hash PK as an implementation detail not exposed to the user.

StoredCredential uses a `Sha256Hash` directly as a primary key, which means that its `::get` method takes a reference to that.

PR #2465, opened 2026-08-13
refactor: linear tx operations (attempt 2) [WPB-27876]

coriolinus:
- store a vector of operations in the transaction, not disjoint and order-independent maps
- rm most of `dynamic_dispatch` stuff which is no longer relevant
- use function pointers that capture type data to do dispatch instead
- ensure we have sufficient caching that we don't need linear scan 
- add several tests to demonstrate correct behavior

## Phases of Work

4961c43..ddb694c: intended scope of work\
f8033ec: huh, now the test suite is producing an error exactly like ios was complaining about, let's fix that\
9fbe9c3: huh, that doesn't look right, let's have a test\
9625c61: yeah let's fix that while we're at it


keystore/src/transaction/dynamic_dispatch/operation.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+impl Operation {
+    /// Apply this operation to the database.
+    pub(in crate::transaction) fn apply(self, tx: &rusqlite::Transaction<'_>) -> CryptoKeystoreResult<()> {
+        match self {
+            Operation::Nop => Ok(()),
+            Operation::Upsert { entity, apply } => (apply)(&*entity, tx),
+            Operation::Delete { entity_id, apply } => (apply)(&entity_id, tx),
+            Operation::BulkDelete { search_key, apply, .. } => (apply)(&*search_key, tx),
```
SimonThormeyer, line 78: I'm trying to understand here (and haven't succeeded so far):

- Why don't we need to use `BulkDelete`'s `matches` during apply?
- If we don't use the `matches` function here, what are we storing it for?

EDIT: Ah, I see, it goes into `BulkDeleteFilters`.

coriolinus: A `BulkDelete` operation comes down to SQL like `DELETE FROM foo WHERE search_key = ?;`. So actually applying it requires only the function which contains that statement, and the search key. 

We use `matches` all the time, just not when applying the operation to the database. In particular, we use it in `as_bulk_delete_filter`, which transforms this operation into a filter which can be applied to an in-memory entity instance. This in turn lets us ensure we're filtering out entities which previously existed at the database level, but should not appear in the transaction, because a bulk delete operation covers them.

keystore/src/connection/fetch_from_database.rs
```diff
@@ -3,27 +3,41 @@ use std::{borrow::Borrow, sync::Arc};
...
+
+        if let Some(record) = read_outcome.entity {
+            return Ok(record.map(Arc::unwrap_or_clone));
+        }
 
         // Otherwise get it from the database
         let conn = self.conn().await;
-        E::get(&conn, id)
+        let db_entity = E::get(&conn, id)?.filter(|entity| !read_outcome.should_omit(entity));
+
```
SimonThormeyer, line 39: AFAICS, we read eagerly from the DB here? If `should_omit()` was somehow bound to the entity id and not the enitty, we could in some cases avoid the DB read here, right?

keystore/src/transaction/entity_read.rs
```diff
@@ -1,47 +1,66 @@
...
     /// * `None` - there is no information about the record in the cache
-    async fn get_by_entity_id<E>(&self, entity_id: &EntityId) -> Option<Option<Arc<E>>>
+    ///
+    /// ## Caution
+    ///
+    /// This correctly filters out upserted entities which were later bulk-deleted.
+    /// However, it _cannot_ indicate bulk deletions for entities which were not
+    /// previously upserted in this transaction. Other composing methods which have
+    /// context from the database have to perform their own bulk-deletion filtering!
+    async fn get_by_entity_id<E>(&self, entity_id: &EntityId) -> ReadOutcome<E>
```
SimonThormeyer: Now that the return type has changed, the doc comment should be updated

keystore/src/transaction/entity_read.rs
```diff
@@ -1,47 +1,66 @@
...
-        self.find_in_cache::<E>(entity_id).await.map(Some)
+        outcome
     }
 
     /// The result of this function will have different contents for different scenarios:
     /// * `Some(Some(E))` - the transaction cache contains the record
     /// * `Some(None)` - the deletion of the record has been cached
     /// * `None` - there is no information about the record in the cache
-    pub(crate) async fn get<E>(&self, id: &E::PrimaryKey) -> Option<Option<Arc<E>>>
+    pub(crate) async fn get<E>(&self, id: &E::PrimaryKey) -> ReadOutcome<E>
```
SimonThormeyer: The doc comment seems stale, here, too

keystore/src/transaction/entity_read.rs
```diff
@@ -53,68 +72,87 @@ impl Transaction {
     /// * `Some(Some(E))` - the transaction cache contains the record
     /// * `Some(None)` - the deletion of the record has been cached
     /// * `None` - there is no information about the record in the cache
-    pub(crate) async fn get_borrowed<E>(&self, id: &E::BorrowedPrimaryKey) -> Option<Option<Arc<E>>>
+    pub(crate) async fn get_borrowed<E>(&self, id: &E::BorrowedPrimaryKey) -> ReadOutcome<E>
```
SimonThormeyer: Stale doc comment, here, too

keystore/src/transaction/entity_read.rs
```diff
@@ -53,68 +72,87 @@ impl Transaction {
...
 
-    pub(super) async fn find_all_in_cache<E>(&self) -> Vec<Arc<E>>
+    /// Apply the transaction's operations in order, producing all entities of type `E`.
+    ///
+    /// ## Caution
+    ///
+    /// This correctly filters out upserted entities which were later bulk-deleted.
+    /// However, it _cannot_ filter out bulk deletions for entities which were not
+    /// previously upserted in this transaction. Other composing methods which have
+    /// context from the database have to perform their own bulk-deletion filtering!
```
SimonThormeyer: I appreciate the comment!

keystore/src/connection/fetch_from_database.rs
coriolinus: That's a lazy DB read, not an eager one: if `read_outcome` knows about either an insertion or deletion at the transaction level, that's what gets returned before here.

`should_omit` is a composite of all the bulk delete filters. The thing is, bulk delete filters mostly don't operate on the primary key; that's the point of them. So they can't effectively be tied to the primary key.

keystore/src/transaction/entity_read.rs
coriolinus: Good point: f4d9ca88be

keystore/src/transaction/entity_read.rs
coriolinus: Annotating stuff like this was critical for me to be able to keep it all straight!

coriolinus (commented):
I replied to comments but GH decided that was a revivew not replies, so now I have to write this to get my real responses to show up.

keystore/src/transaction/entity_read.rs
coriolinus: f4d9ca88be

keystore/src/transaction/entity_read.rs
coriolinus: f4d9ca88be

SimonThormeyer (commented):
Amazing work!

SimonThormeyer (approved):

keystore/src/connection/migrations/V31__mls_pending_messages_primary_key.sql
```diff
@@ -1,11 +1,34 @@
 -- The absence of a primary key on the `mls_pending_messages` table is causing real problems.
 -- This migration creates a primary key by concatenating the conversation id with the message and hashing.
+--
+-- Rebuilding the table is also the only way to drop a constraint in SQLite, so this migration takes the
+-- opportunity to drop the foreign key which V7 declared on this table. That removal is deliberate and it
+-- fixes a bug:
+--
+-- V7 introduced `mls_pending_messages` for a single purpose: holding messages which arrived for a group we
```
typfel, line 8: Makes me question why we allow joining by external commit inside a transaction if it creates these kind of edge cases.

keystore/src/transaction/entity_write.rs
```diff
@@ -75,24 +60,47 @@ impl Transaction {
...
+            if let Some(idx) = last_delete_idx {
+                operations.make_nop(idx);
+            }
+        }
+
         Ok(())
     }
+
+    /// Bulk-remove entities matching a given search key.
+    pub async fn bulk_remove<E, S>(&self, search_key: S)
```
typfel, line 98: So much complexity comes from the bulk deletion. Did you ever consider getting rid of it?

SimonThormeyer: It's useful, e.g., to reset tnt message counters for a conversation. The more tables with foreign keys we add, the more usful this becomes.

PR #2466, opened 2026-08-14
chore: `FetchFromDatabase` returns arcs

coriolinus:
We still often have to `Arc::unwrap_or_clone`, but at least now it's not literally every time. Should be a trivial perf boost.


SimonThormeyer (approved):
Approved, pending green CI.

PR #2468, opened 2026-08-14
refactor(keystore): tnt message counters get their own table  

SimonThormeyer:
We now have seperate counters per epoch and receiver.


SimonThormeyer:
This was an accident. Meant to create #2469

PR #2469, opened 2026-08-14
refactor(keystore): targeted message tx counters get their own table  

SimonThormeyer:
We now have seperate counters per epoch and receiver.


crypto/src/mls/conversation/immutable/persistence.rs
```diff
@@ -34,28 +50,33 @@ impl Conversation {
...
             .await
             .map_err(KeystoreError::wrap("finding a persisted mls group"))?;
         let Some(group) = group else { return Ok(None) };
         let mut group = Arc::unwrap_or_clone(group);
-        let conversation = Self::from_serialized_state(
-            session,
-            std::mem::take(&mut group.state),
-            group.tnt_message_counter.into(),
-        )?;
+        let counters = database
```
typfel: So we are eagerly loading all sender counters when loading a conversation? But we only need them if we are encrypting a message and then only for a specific recipient. Couldn't we load the counters lazily?

SimonThormeyer: Good idea

SimonThormeyer: See 11dd5887d9e5e4234ef9e2ce413a8a1078f93804 and child

keystore/src/entities/mls/tnt_message_sender_counter.rs
```diff
@@ -0,0 +1,205 @@
...
+        Ok(Self {
+            conversation_id: row.get("conversation_id")?,
+            receiver: row.get("receiver")?,
+            count: row.get("count")?,
+        })
+    }
+}
+
+impl PrimaryKey for TntMessageCounter {
+    type PrimaryKey = (Vec<u8>, u32);
```
coriolinus: Nit: I'd prefer to see an explicit struct instead of a tuple for this, like

```rust
struct TntMessageCounterPk {
  converation_id: Vec<u8>,
  receiver_idx: u32,
}

impl KeyType for TntMessageCounterPk { ... }
```

IMO we're doing a bit too much with a primary key for this kind of tuple to really make sense. And writing it out as a struct encourages us to write a borrowed form so we can impl `EntityGetBorrowed`, etc.

keystore/src/traits/fetch_from_database.rs
```diff
@@ -26,6 +28,28 @@ pub trait FetchFromDatabase: Send + Sync {
...
         E: 'static + Entity + Clone + Send + Sync;
 
+    /// Load all `E`s from the database, grouped by their canonical `SearchKey`.
+    ///
+    /// This is useful when every entity needs to be loaded and associated with another set of entities by a search key.
+    /// It performs a single bulk load rather than one search per key.
+    ///
+    /// A requirement for this to be callable is that a `SearchKey` can be trivially extracted from an `&E`, and that
+    /// this extraction is implemented via `ExtractableSearchKey<SearchKey>`.
+    async fn load_all_by_search_key<E, SearchKey>(&self) -> CryptoKeystoreResult<HashMap<SearchKey, Vec<Arc<E>>>>
```
coriolinus: I'm not sure why this function exists. Any caller trivially has the ability to express this via `load_all`, and already knows what entity type it's dealing with, so we wouldn't need the trait `ExtractableSearchKey`. 

Is there a use case I'm not seeing? Because right now the PR doesn't seem to have a need for this method and that trait, as far as I can see.

crypto/src/mls/conversation/immutable/persistence.rs
```diff
@@ -73,21 +70,15 @@ impl Conversation {
...
                 // zeroizes on drop, which means we are forced to clone all the group's fields, because
                 // otherwise the drop impl couldn't run.
-                let conversation = Self::from_serialized_state(
-                    session.clone(),
-                    group.state.clone(),
-                    counters_by_group.remove(&group.id.clone().into()).unwrap_or_default(),
-                )?;
+                let conversation =
+                // The tnt message counters are empty when initializing, we're loading them lazily on usage.
+                    Self::from_serialized_state(session.clone(), group.state.clone(), Default::default())?;
```
coriolinus: If we're not using the counters parameter then we should remove it from the signature of `Self::from_serialized_state`.

crypto/src/mls/conversation/immutable/mod.rs
```diff
@@ -44,14 +45,35 @@ impl MlsGroupState {
...
 
     /// Get the tnt message counter bound to this conversation and the given recipient after incrementing it.
-    pub(in crate::mls::conversation) fn obtain_tnt_message_counter(
+    ///
+    /// If the counter hasn't been used yet for this conversation, it is loaded from the database or initialized
+    /// freshly.
+    pub(in crate::mls::conversation) async fn obtain_tnt_message_counter(
         &mut self,
         recipient: LeafNodeIndex,
+        database: &Database,
```
coriolinus: Maybe `&impl FetchFromDatabase`? There are a few contexts where we have a `Transaction` and not a `Database`.

coriolinus (commented):
Most of these are nits, but I would like to understand better the rationale for the existence of `load_all_by_search_key` + `ExtractableSearchKey` before approving.

keystore/src/traits/fetch_from_database.rs
SimonThormeyer: It's right, the need for this was removed when refactoring from eager counter to lazy counter loading.
The orginal intention of adding this in the keystore was just because I thought we might need this more often in the future, so every caller wouldn't need to re-implement this.

Happy to drop this function now, we can add it when we need it.

SimonThormeyer: 916ca49aa2bbb6c0c5c1178cb00611f528dbcf92

coriolinus (approved):
Nice! Obviously I'd prefer to have you address the other comments also but ultimately they're up to you; just get CI green and you're good to go.

crypto/src/mls/conversation/immutable/persistence.rs
typfel: Awesome!

keystore/src/connection/idb_migration/legacy/traits/key_type.rs
```diff
@@ -40,7 +40,6 @@ impl_keytype!(&str, |self| self.as_bytes());
 impl_keytype!(String, |self| self.as_bytes(), |bytes| str::from_utf8(bytes)
     .ok()
     .map(ToOwned::to_owned));
-
```
coriolinus, line 43: Nit: the whitespace here helped with readability IMO

keystore/src/mls.rs
```diff
@@ -72,12 +72,11 @@ impl Database {
         &self,
         group_id: impl AsRef<[u8]> + Send,
         state: &[u8],
-        tnt_message_counter: u32,
+        #[expect(unused)] tnt_message_counters: impl IntoIterator<Item = (u32, u32)>,
```
coriolinus: Is there any reason to keep this unused parameter for the future?

keystore/src/connection/migrations/V24__message_sender_nonce.sql
```diff
@@ -1 +1 @@
-ALTER TABLE mls_groups ADD COLUMN tnt_message_counter INTEGER NOT NULL DEFAULT 0;
+ALTER TABLE mls_groups ADD COLUMN sender_nonce INTEGER NOT NULL DEFAULT 0;
```
istankovic, line 1: Can we please avoid the term `nonce`?

keystore/src/mls.rs
SimonThormeyer: Oh, do we even need this function at all?

crypto/src/mls/conversation/immutable/mod.rs
```diff
@@ -6,22 +6,33 @@ mod e2ei;
...
+    CipherSuite, ConversationConfiguration, ConversationId, CredentialRef, ExternalSender, KeystoreError, OpenMlsError,
+    Session, mls::TntMessageCounter,
 };
 
 #[derive(derive_more::Constructor, derive_more::Deref, derive_more::DerefMut, derive_more::Debug)]
 pub(crate) struct MlsGroupState {
     #[deref]
     #[deref_mut]
     group: MlsGroup,
+    /// The count of tnt messages sent this epoch, keyed by recipients.
```
istankovic: It would be good to expand this comment, or elsewhere provide documentation on why we have these counters and how they are supposed to be used.

keystore/src/connection/migrations/V24__message_sender_nonce.sql
coriolinus: Not in this case: we released 10.3.0 with this as `sender_nonce`, then altered it later; this commit is restoring it so we don't explode the databases of everyone who ever ran a 10.3.0 version.

crypto/src/mls/conversation/immutable/mod.rs
```diff
@@ -6,22 +6,33 @@ mod e2ei;
...
     #[debug(skip)]
-    tnt_message_counter: TntMessageCounter,
+    tnt_message_counters: HashMap<LeafNodeIndex, TntMessageCounter>,
```
istankovic: What is recipient going to be in case of transient messages?

istankovic (changes requested):
It is not clear how we are going to handle transient message counters, I feel this needs more work.

crypto/src/mls/conversation/immutable/mod.rs
```diff
@@ -6,22 +6,36 @@ mod e2ei;
...
 
 #[derive(derive_more::Constructor, derive_more::Deref, derive_more::DerefMut, derive_more::Debug)]
 pub(crate) struct MlsGroupState {
     #[deref]
     #[deref_mut]
     group: MlsGroup,
+    /// The count of targeted messages sent this epoch, keyed by recipients.
+    /// Supposed to be used when encrypting targeted messages only, and to be reset whenever the mls epoch is
+    /// incremented. Do not access this field directly, use [MlsGroupState::obtain_tnt_message_counter] and
+    /// [MlsGroupState::reset_tnt_message_counters] only.
```
istankovic: Could you please also mention the purpose of the counters (to implement message replay protection)?

crypto/src/mls/conversation/immutable/mod.rs
```diff
@@ -6,22 +6,36 @@ mod e2ei;
...
     #[debug(skip)]
-    tnt_message_counter: TntMessageCounter,
+    targeted_message_counters: HashMap<LeafNodeIndex, TntMessageCounter>,
```
istankovic: Should this be `targeted_message_tx_counters`?

crypto/src/mls/conversation/immutable/mod.rs
```diff
@@ -6,22 +6,39 @@ mod e2ei;
...
     #[deref]
     #[deref_mut]
     group: MlsGroup,
+    /// The count of targeted messages sent (therefore `tx`) this epoch, keyed by recipients.
+    /// Supposed to be used when encrypting targeted messages only, and to be reset whenever the mls epoch is
+    /// incremented. Do not access this field directly, use [MlsGroupState::obtain_targeted_message_tx_counter] and
+    /// [MlsGroupState::reset_targeted_message_tx_counters] only.
+    ///
+    /// The purpose of these counters is replay protection on the recipient side: we provide the count when sending a
+    /// message to a receiver, and they check if the counter is greater than any they've seen before.
```
istankovic, line 39: Great, thanks!

istankovic (approved):

PR #2472, opened 2026-08-17
refactor: entity id without serialization

coriolinus:
Simon mentioned during review of one of the recent keystore PRs that there seemed to be no reason for entity IDs / keys to require serialization. It was a good idea, just out of scope back then. But this implements that. There's no more implicit serialization in the live transaction at all!


SimonThormeyer (approved):
Great to see this implemented already!
Approving, pending green CI.

keystore/src/transaction/dynamic_dispatch/operation.rs
```diff
@@ -109,7 +109,9 @@ impl Operation {
         Self::Delete {
             entity_id,
             apply: |entity_id, tx| {
-                let id = entity_id.primary_key::<E>()?;
+                let id = entity_id.primary_key::<E>().expect(
+                    "the entity id is set wehn we create a delete operation and must always have the correct type",
```
typfel: s/wehn/when

istankovic: Typo: `wehn`

PR #2475, opened 2026-08-18
fix: buffered future blocks group removal [WPB-28040]

coriolinus:
We should only restore buffered messages if we are still a member of the group at that point.

### DIFF
--- a/crypto/src/mls/conversation/mutable/decrypt/buffer_messages.rs
+++ b/crypto/src/mls/conversation/mutable/decrypt/buffer_messages.rs
@@ -334,6 +334,84 @@ mod tests {
         .await
     }
 
+    /// A buffered future message must not block us from processing our own removal.
+    ///
+    /// This is ordinary out-of-order delivery, the very thing the buffering machinery exists for:
+    /// the DS hands us an epoch `n + 1` application message before it hands us the epoch `n + 1`
+    /// commit which removed us. Once that commit arrives, the conversation must be wiped, and the
+    /// buffered message must go with it. If instead the buffered message is replayed against the
+    /// new epoch, decryption fails identically on every retry, in every subsequent transaction, so
+    /// the client can never process its own removal and the conversation is stuck forever.
+    #[apply(all_cred_cipher)]
+    async fn buffered_message_must_not_block_own_removal(case: TestContext) {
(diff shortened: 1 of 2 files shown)


crypto/src/mls/conversation/mutable/decrypt/buffer_messages.rs
```diff
@@ -334,6 +334,87 @@ mod tests {
...
+    ///
+    /// This is ordinary out-of-order delivery, the very thing the buffering machinery exists for:
+    /// the DS hands us an epoch `n + 1` application message before it hands us the epoch `n + 1`
+    /// commit which removed us. Once that commit arrives, the conversation must be wiped, and the
+    /// buffered message must go with it. If instead the buffered message is replayed against the
+    /// new epoch, decryption fails identically on every retry, in every subsequent transaction, so
+    /// the client can never process its own removal and the conversation is stuck forever.
+    #[apply(all_cred_cipher)]
+    async fn buffered_message_must_not_block_own_removal(case: TestContext) {
+        Box::pin(async move {
```
SimonThormeyer: I know this is done in many other places, but we probably don't need to `Box::pin` the entire test to avoid a stack overflow here?

coriolinus: ~~Doesn't overflow on my machine; or CI (so far); I'd only do it if we have evidence it's necessary.~~ Oops, misunderstood your comment.

coriolinus: Dropped the pinning en-passant while rebasing onto main.

crypto/src/mls/conversation/mutable/decrypt/mod.rs
```diff
@@ -358,7 +358,7 @@ impl ConversationMut {
 
                 // can't use `.then` because async
                 let mut buffered_messages = None;
-                if recursion_policy == RecursionPolicy::AsNecessary {
+                if is_active && recursion_policy == RecursionPolicy::AsNecessary {
```
SimonThormeyer, line 361: Wanna copy and paste the commit messasge body here, for context why this is needed?

SimonThormeyer (approved):
lgtm!

PR #2482, opened 2026-08-18
spike: evaluate and simplify errors [WPB-23465] 

coriolinus:
Turned out that we could improve them a bit!


crypto/src/mls/credential/x509.rs
```diff
@@ -149,7 +148,7 @@ impl Credential {
             let cert = self
                 .mls_credential()
                 .parse_leaf_cert()
-                .map_err(RecursiveError::mls_credential("parsing leaf certificate"))?
+                .map_err(RecursiveError::context("parsing leaf certificate"))?
                 // This can actually never happen and points to a type issue with credentials
                 .expect("parse_leaf_cert to return a Certificate");
```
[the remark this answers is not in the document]
coriolinus: /ar We explicitly check that the credential type is x509, and the option being unwrapped here only ever returns None if the credential is not x509. A failure of that expectation is not a runtime error; it's the credential's reported type disagreeing with its actual type. A panic is the appropriate response in that situation.

crypto/src/error/open_mls.rs
```diff
@@ -82,9 +69,6 @@ pub enum OpenMlsErrorKind {
     /// OpenMLS merge commit error
     #[error(transparent)]
     MlsMergeCommitError(#[from] openmls::prelude::MergeCommitError<core_crypto_keystore::CryptoKeystoreError>),
-    /// OpenMLS keypackage validation error
-    #[error(transparent)]
-    MlsKeyPackageValidationError(#[from] openmls::prelude::KeyPackageVerifyError),
```
typfel, line 87: In what way is this error not used anymore? Surely the error can happen when we digest an incoming key package like in `ConversationMut::propose_add_member`. What happens now is this error wrapped in some generic error?

coriolinus: The only proof I have is by demonstration: everything compiles and the tests pass, therefore we never constructed this error variant.

But I'll look into it.

crypto/src/mls/session/error.rs
```diff
@@ -13,6 +13,9 @@ pub enum Error {
     InvalidUserId,
     #[error("A MLS operation was requested but MLS hasn't been initialized on this instance")]
     MlsNotInitialized,
+    /// This error is emitted when the requested conversation couldn't be found in our store
+    #[error("Couldn't find conversation")]
+    ConversationNotFound(crate::ConversationId),
```
typfel, line 18: So we we got rid `LeafError` with the only drawback that we now define `ConversationNotFound` in two places?

I guess the commit message should say:
> **Fortunately** there are just two legitimate paths to this error

coriolinus: Haha, fair.

crypto/src/error/open_mls.rs
coriolinus: It got subsumed into `OpenMlsError`:

```rust
        let (proposal, _ref) = self
            .mutate_group(async |_, group, _| {
                group
                    .propose_add_member(&crypto_provider, signer, key_package)
                    .await
                    .map_err(OpenMlsError::wrap("propose add member"))
                    .map_err(Into::into)
            })
            .await?;
```

coriolinus: Wait, that's not a sufficient answer. The internal openlms `propose_add_member` returns a `ProposeAddMemberError`, not a `KeyPackageVerifyError`. `ProposeAddMemmberError` already has a variant which wraps `KeyPackageVerifyError`.

crypto/src/mls/session/error.rs
coriolinus: (Not adjusting before review is done though.)

typfel (commented):
Looks fine to me but I'd like some else to also review it which is more rust literate, ping @fewerner @istankovic

crypto/src/mls/session/error.rs
```diff
@@ -59,13 +28,3 @@ pub enum Error {
...
-
-impl Error {
-    pub fn tls_serialize(item: &'static str) -> impl FnOnce(tls_codec::Error) -> Self {
-        move |source| Self::TlsSerialize { item, source }
-    }
-
-    pub fn tls_deserialize(item: &'static str) -> impl FnOnce(tls_codec::Error) -> Self {
-        move |source| Self::TlsDeserialize { item, source }
-    }
-}
```
istankovic, line 71: ~~I'm struggling to imagine why these were ever needed. :-)~~
Ugh, nevermind, I misread it.

crypto/src/error/proteus.rs
```diff
@@ -1,27 +1,28 @@
...
     #[error(transparent)]
+    ProteusInternalError(#[from] proteus_wasm::error::ProteusError),
     /// Error when there's a critical error within a proteus Session
+    #[error(transparent)]
     ProteusSessionError(#[from] proteus_wasm::session::Error<core_crypto_keystore::CryptoKeystoreError>),
     /// Common errors we generate
-    #[cfg(feature = "proteus")]
     #[error("{0}")]
     Leaf(#[from] crate::LeafError),
 }
```
istankovic, line 28: Nice, this is much better. 👍

crypto/src/mls/session/error.rs
coriolinus: We have a bunch of TLS serialization/deserialization in this crate (mostly for dealing with the keystore, abstraction in the wrong place), and so we needed to wrap the relevant errors. This PR centralizes TLS error handling, but we still do need it.

crypto/src/transaction_context/e2e_identity/error.rs
```diff
@@ -1,38 +0,0 @@
...
-    X509CertDerError(#[from] x509_cert::der::Error),
-    #[error(transparent)]
-    OpenMls(#[from] crate::OpenMlsError),
-    #[error(transparent)]
-    Keystore(#[from] crate::KeystoreError),
-    #[error(transparent)]
-    PkiEnvironmentEror(#[from] wire_e2e_identity::pki_env::Error),
-    #[error(transparent)]
-    Recursive(#[from] crate::RecursiveError),
-}
```
istankovic, line 38: Niiiiiiice! 🍷

crypto/src/error/tls.rs
```diff
@@ -0,0 +1,45 @@
...
+    Deserializing,
+}
+
+impl std::fmt::Display for Direction {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(match self {
+            Self::Serializing => "serializing",
+            Self::Deserializing => "deserializing",
+        })
+    }
```
istankovic: Why not use `derive_more::Display`?

coriolinus: Hm, I guess I overlooked that.

coriolinus: d9a96ae992

crypto/src/error/recursive.rs
```diff
@@ -1,143 +1,82 @@
...
+                    source: Box<$source>,
+                },
+            )+
+        }
+
+        impl RecursiveError {
+            $(
+                /// Wrap an error convertible into
+                #[doc = concat!("[`", stringify!($source), "`],")]
+                /// given the current context.
```
istankovic, line 55: The `#[doc` line looks misplaced.

coriolinus: It's not: doc comments with `///` just concatenate with doc comments with `#[doc`, so this inserts a doc link pointing at the target of `$source`.

crypto/src/error/recursive.rs
```diff
@@ -149,12 +88,32 @@ macro_rules! impl_to_recursive_error_for {
...
+    /// The variant is selected from the error's type, so this is usually what you want:
+    ///
+    /// ```ignore
+    /// self.session().await.map_err(RecursiveError::context("getting mls client"))?
+    /// ```
+    ///
+    /// This only accepts the module error types themselves. When you need to convert first, for
+    /// instance because you are holding a [`crate::TlsCodecError`] which several module errors
+    /// can absorb, reach for the variant-specific constructor such as
+    /// [`Self::mls_conversation`] instead.
```
istankovic, line 116: Appreciate the comment!

istankovic (approved):
Looks good, this is a very nice cleanup. :+1:

PR #2488, opened 2026-08-19
chore: add prerelease script

coriolinus:
This should hopefully help us not release more than we intended to.

### DIFF
--- a/scripts/prerelease.sh
+++ b/scripts/prerelease.sh
@@ -10,7 +10,7 @@ if [ -z "${1:-}" ]; then
 fi
 major_version="$1"
 
-if ! rg '^\d+$' <<< "$major_version" > /dev/null; then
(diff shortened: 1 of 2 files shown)


scripts/prerelease.sh
```diff
@@ -0,0 +1,48 @@
...
+
+set -euo pipefail
+
+if [ -z "${1:-}" ]; then
+    echo "usage: $0 <MAJOR_VERSION>"
+    exit 1
+fi
+major_version="$1"
+
+if ! rg '^\d+$' <<< "$major_version" > /dev/null; then
```
typfel: This introduced `rg` as a new dependency we can't use sed or grep for parsing the number?

coriolinus: Oh, good point. For some reason I thought it was already part of the basic set of tools we use, but you're right.

typfel (approved):

PR #2490, opened 2026-08-20
refactor: move transaction file lock into rust WPB-28130

typfel:
Move the transaction file lock into rust so that it can also be enabled for the ios target of kmp builds.

### DIFF
--- a/keystore/Cargo.toml
+++ b/keystore/Cargo.toml
@@ -62,6 +62,9 @@ security-framework-sys = "2.16"
 
 [target.'cfg(not(target_os = "unknown"))'.dependencies]
 async-fs = "2.2"
+# runtime-agnostic thread pool, used to wait on the cross-process file lock without blocking the
+# executor; already in the tree as a dependency of `async-fs`
+blocking = { version = "1.6", optional = true }
 rusqlite = { version = "0.38", default-features = false, features = [
   "bundled-sqlcipher-vendored-openssl",
 ] }
--- a/keystore/Cargo.toml
+++ b/keystore/Cargo.toml
@@ -14,6 +14,10 @@ workspace = true
 default = ["proteus-keystore"]
 proteus-keystore = ["dep:proteus-traits"]
 log-queries = ["rusqlite/trace"]
+# Extend the guarantee that at most one keystore transaction runs at a time from a single process
+# to every process on the machine, via an advisory lock on a file next to the database.
+# Requires a real filesystem, so it must stay off for `target_os = "unknown"`.
+cross-process-lock = ["dep:blocking"]
 # kept here because legacy keystore module is using this feature.
 serde = []
 dummy-entity = []
--- a/keystore/src/connection/mod.rs
+++ b/keystore/src/connection/mod.rs
@@ -1,6 +1,8 @@
 mod encryption;
 mod entity_extension_methods;
 mod fetch_from_database;
+#[cfg(feature = "cross-process-lock")]
+mod file_lock;
 mod filesystem;
 #[cfg(target_os = "unknown")]
 mod idb_migration;
@@ -10,18 +12,20 @@ mod migrations;
 #[cfg(target_os = "unknown")]
 mod os_unknown;
 mod transaction;
+mod transaction_lock;
 
 use std::sync::Arc;
 
-use async_lock::{Mutex, MutexGuard, Semaphore};
+use async_lock::{Mutex, MutexGuard};
 use rusqlite::Connection;
 #[cfg(feature = "log-queries")]
 use rusqlite::trace::{TraceEvent, TraceEventCodes};
 
-pub(crate) use self::filesystem::Filesystem;
 #[cfg(target_os = "unknown")]
 pub use self::idb_migration::{delete_legacy_idb, legacy_idb_exists};
 pub use self::migrations::migrate_db_key_type_to_bytes;
+use self::transaction_lock::TransactionLock;
+pub(crate) use self::{filesystem::Filesystem, transaction_lock::TransactionGuard};
 use crate::{
     CryptoKeystoreResult, DatabaseKey, connection::migrations::MigrationTarget, transaction::Transaction,
     unique_arc::UniqueWeak,
@@ -46,9 +50,8 @@ pub struct Database {
     pub(crate) filesystem: Mutex<Box<dyn Filesystem>>,
     #[debug(skip)]
     pub(crate) transaction: Mutex<Option<UniqueWeak<Transaction>>>,
-    // we need this `Arc` so we can create an owned guard, so that
-    // `self.transaction` doesn't need a self-referential lifetime.
-    transaction_semaphore: Arc<Semaphore>,
+    // ensures at most one transaction is in flight against this database at a time
+    transaction_lock: TransactionLock,
 }
 
 impl Database {
@@ -98,8 +101,6 @@ impl Database {
         filesystem: Box<dyn Filesystem>,
         migration_target: MigrationTarget,
     ) -> CryptoKeystoreResult<Self> {
-        const ALLOWED_CONCURRENT_TRANSACTIONS_COUNT: usize = 1;
-
         #[cfg(feature = "log-queries")]
         conn.trace_v2(TraceEventCodes::SQLITE_TRACE_STMT, Some(log_query));
 
@@ -112,13 +113,15 @@ impl Database {
         }
 
         migrations::run_migrations(&mut conn, migration_target)?;
+
+        let transaction_lock = TransactionLock::new(conn.path().unwrap_or_default())?;
         let conn = conn.into();
 
         Ok(Self {
             conn,
             filesystem: filesystem.into(),
             transaction: Default::default(),
-            transaction_semaphore: Arc::new(Semaphore::new(ALLOWED_CONCURRENT_TRANSACTIONS_COUNT)),
+            transaction_lock,
         })
     }
 
@@ -168,14 +171,17 @@ impl Database {
 
     /// Wait for any running transaction to finish, then take the connection out of this database,
     /// preventing this from being used again.
-    async fn take(self) -> CryptoKeystoreResult<(Connection, Box<dyn Filesystem>)> {
-        let _semaphore = self.transaction_semaphore.acquire().await;
-        Ok((self.conn.into_inner(), self.filesystem.into_inner()))
+    ///
+    /// The returned guard keeps other processes out for as long as the caller holds it, so that
+    /// teardown is not interleaved with somebody else's transaction.
+    async fn take(self) -> CryptoKeystoreResult<(Connection, Box<dyn Filesystem>, TransactionGuard)> {
+        let guard = self.transaction_lock.acquire().await?;
+        Ok((self.conn.into_inner(), self.filesystem.into_inner(), guard))
     }
 
     // Close this database connection
     pub async fn close(self) -> CryptoKeystoreResult<()> {
-        let (conn, _fs) = self.take().await?;
+        let (conn, _fs, _guard) = self.take().await?;
         conn.close().map_err(|(_conn, err)| err)?;
         Ok(())
     }
@@ -186,7 +192,7 @@ impl Database {
     /// Future opens will always succeed with any arbitrary encryption key; they will
     /// simply open an empty database.
     pub async fn wipe(self) -> CryptoKeystoreResult<()> {
-        let (conn, fs) = self.take().await?;
+        let (conn, fs, _guard) = self.take().await?;
         conn.execute_batch(
             "
             PRAGMA writable_schema = 1;
@@ -200,6 +206,10 @@ impl Database {
         if let Some(path) = location {
             // not in-memory
             fs.delete(&path).await?;
+            // `_guard` is still alive, so we unlink the lock file while still holding it; that
+            // keeps a peer from acquiring the fresh lock file before the database is really gone.
+            #[cfg(feature = "cross-process-lock")]
+            file_lock::remove_lock_file(&path).await?;
         }
         Ok(())
     }
--- a/keystore/src/connection/transaction_lock.rs
+++ b/keystore/src/connection/transaction_lock.rs
@@ -98,3 +98,141 @@ impl TransactionLock {
         }))
     }
 }
+
+#[cfg(all(test, feature = "cross-process-lock", not(target_os = "unknown")))]
+mod tests {
+    use std::time::Duration;
+
+    use futures_lite::future;
+    use smol::Timer;
+
+    use crate::{CryptoKeystoreError, Database, DatabaseKey};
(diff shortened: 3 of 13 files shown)


keystore/src/connection/transaction_lock.rs
```diff
@@ -0,0 +1,263 @@
...
+const ALLOWED_CONCURRENT_TRANSACTIONS_COUNT: usize = 1;
+
+/// Gates the start of a keystore transaction.
+#[derive(Debug)]
+pub(crate) struct TransactionLock {
+    // we need this `Arc` so we can create an owned guard, so that the transaction doesn't need
+    // a self-referential lifetime into the database.
+    semaphore: Arc<Semaphore>,
+    /// `None` for in-memory databases, and whenever cross-process locking is compiled out.
+    file: Option<Arc<FileLock>>,
```
coriolinus, line 38: I think that it's more elegant, even if it's a bit more work, to conditionally compile this field only when `#[cfg(feature = "cross-process-lock")]`. It looks to me like there are only four references to the field; that's not an unreasonable number of places to special-case.

The benefits are that we don't need a stub module where the feature is not set, and we cannot possibly get into a state where the feature is set but the option is `None`.

keystore/src/connection/transaction_lock.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// Uninhabited: without the `cross-process-lock` feature there is never a file lock to hold.
+    pub(crate) enum FileLock {}
+
+    impl FileLock {
+        pub(crate) fn open(_database_path: &str) -> CryptoKeystoreResult<Option<Self>> {
+            Ok(None)
+        }
+
+        pub(crate) async fn lock(self: &Arc<Self>) -> CryptoKeystoreResult<FileLockGuard> {
+            match **self {}
```
coriolinus: I'm kind of shocked that this compiles; it took me a minute to figure it out. 

(For the future: `Self` is an uninhabited enum, so `Arc<Self>` can never be constructed, so the function can never be called; matching on the uninhabited enum just points that out to the compiler.)

Honestly I'd prefer the simpler-but-equivalent `unreachable!("Self is uninhabited")`.

keystore/src/connection/transaction_lock.rs
```diff
@@ -0,0 +1,263 @@
...
+                return Ok(None);
+            }
+
+            let path = lock_file_path(database_path);
+            let file = File::options()
+                .read(true)
+                .write(true)
+                .create(true)
+                .truncate(false)
+                .open(&path)
```
coriolinus: Wouldn't `let file = File::create(&path)` be equivalent?

keystore/src/connection/transaction_lock.rs
```diff
@@ -0,0 +1,263 @@
...
+        file: File,
+        path: String,
+    }
+
+    impl FileLock {
+        /// Open (creating if necessary) the lock file guarding the database at `database_path`.
+        ///
+        /// Produces `None` for in-memory databases, which have an empty path and are unreachable
+        /// from other processes anyway.
+        pub(crate) fn open(database_path: &str) -> CryptoKeystoreResult<Option<Self>> {
```
coriolinus: The critical operations, `lock` and `try_lock`, take `&Arc<Self>`. I'd have the constructor do its own arc-wrapping unless there's a strong reason otherwise.

keystore/src/connection/transaction_lock.rs
```diff
@@ -0,0 +1,263 @@
...
+            _file: file,
+        }))
+    }
+}
+
+/// The cross-process half of [`TransactionLock`].
+///
+/// This exists in two versions under one name, chosen by the `cross-process-lock` feature.
+#[cfg(feature = "cross-process-lock")]
+mod file_lock {
```
coriolinus: I'd externalize this module instead of writing it inline. This is simplest if we also remove the dummy module entirely, because then it's just

```rust
#[cfg(feature = "cross-process-lock")]
mod file_lock;
```

But we can also select between two distinct implementations of an external file if we have to:

**connection/transaction_lock/file_lock/mod.rs**

```rust
#[cfg(feature = "cross-process-lock")]
{
  mod impl;
  pub use impl::*;
}

#[cfg(not(feature = "cross-process-lock"))]
{
  mod dummy;
  pub use dummy::*;
}
```

keystore/src/connection/transaction_lock.rs
```diff
@@ -0,0 +1,263 @@
...
+    impl Drop for FileLockGuard {
+        fn drop(&mut self) {
+            // unlocking is a single non-blocking syscall, so it is fine to run inline here
+            if let Err(error) = self.lock.file.unlock() {
+                // dropping the file would release the lock too, but the `FileLock` outlives this
+                // guard, so a failure here really does leave the lock held; all we can do is say so
+                log::error!(
+                    "failed to release the cross-process transaction lock at {}: {error}",
+                    self.lock.path
+                );
```
coriolinus: Should we panic in this case? Honest question. The app can't create any more transactions, and we're not exposing a recovery method. For sure we should emit the log, but then a panic with a message like "irrecoverably blocked" might be a better way to get diagnostic information about this than just letting apps fail.

make/config.mk
```diff
@@ -18,7 +18,7 @@ else
   RELEASE_MODE := release
 endif
 
-SWIFT_CARGO_BUILD_ARGS := $(CARGO_BUILD_ARGS) --features cancellable-transactions
+SWIFT_CARGO_BUILD_ARGS := $(CARGO_BUILD_ARGS) --features cancellable-transactions,cancellable-transactions
```
coriolinus: I'm sure you meant to write `cross-process-lock` here, not duplicate `cancellable-transactions`.

coriolinus (commented):
This is a cool idea, and I like it, but there are some things (mostly structural) that I'd like to clean up before approving.

make/config.mk
typfel: 🤦‍♂️  yes

keystore/src/connection/transaction_lock.rs
typfel: I'll try this out

typfel: > The benefits are that we don't need a stub module where the feature is not set, and we cannot possibly get into a state where the feature is set but the option is None.

Even if the feature is set you could create an in-memory database it would still need to be `None`.

coriolinus: Hm, that's true.

typfel: Does it make sense to make it an enum?

```
pub(crate) enum TransactionGuard {
    SemaphoreGuard(SemaphoreGuardArc),
    FileGuard(SemaphoreGuardArc, FileLockGuard)
}
```

coriolinus: No, I think it has to be something like

```rust
pub(crate) struct TransactionGuard {
   _semaphore_guard: SemaphoreGuardArc,
  #[cfg(feature = "cross-process-lock")]
  _file_guard: Option<FileLockGuard>,
}
```

coriolinus: Conditionally compiling it means we can still get rid of the stub module, so it's still worth it I think.

keystore/src/connection/transaction_lock.rs
typfel: Only difference is that truncate is true which is irrelevant, I'll change it.

keystore/src/connection/transaction_lock.rs
typfel: Agree let's fail early.

keystore/Cargo.toml
```diff
@@ -14,6 +14,10 @@ workspace = true
 default = ["proteus-keystore"]
 proteus-keystore = ["dep:proteus-traits"]
 log-queries = ["rusqlite/trace"]
+# Extend the guarantee that at most one keystore transaction runs at a time from a single process
+# to every process on the machine, via an advisory lock on a file next to the database.
+# Requires a real filesystem, so it must stay off for `target_os = "unknown"`.
+cross-process-lock = ["dep:blocking"]
```
istankovic, line 20: Must this really be a feature? Can't we do the right thing automatically, based on the target platform?

coriolinus: Past experience shows us that it's better in the end to explicitly feature-gate things like this instead of making feature selection implicit according to target platform.

But also, no, we don't have a way to distinguish between android and ios builds at the rust cfg level.

istankovic: > But also, no, we don't have a way to distinguish between android and ios builds at the rust cfg level.

Why wouldn't https://doc.rust-lang.org/reference/conditional-compilation.html#r-cfg.target_os work?

coriolinus: I don't remember, but I do remember that we spent some time investigating the capability some time (a year?) ago, and decided that we couldn't reliably distinguish between the two. Something about the way uniffi builds, maybe?

keystore/src/connection/transaction_lock.rs
typfel: Fixed by [d33d1ac](https://github.com/wireapp/core-crypto/pull/2490/commits/d33d1ac513f86a2bae106de78e2cf7f27ff8847a)

make/config.mk
typfel: Fixed by https://github.com/wireapp/core-crypto/pull/2490/commits/be0249256324a2837075748f75aafd522074ffe3

keystore/src/connection/transaction_lock.rs
typfel: It's gone after https://github.com/wireapp/core-crypto/pull/2490/commits/336359226f33bb399bc5eb87724819b02e3e0233

keystore/src/connection/transaction_lock.rs
typfel: Fixed in https://github.com/wireapp/core-crypto/pull/2490/commits/336359226f33bb399bc5eb87724819b02e3e0233

keystore/src/connection/transaction_lock.rs
typfel: Fixed in https://github.com/wireapp/core-crypto/pull/2490/commits/e3cf031ae43759ecd84592b04a93016013f966db

keystore/Cargo.toml
istankovic: > I don't remember, but I do remember that we spent some time investigating the capability some time (a year?) ago, and decided that we couldn't reliably distinguish between the two. Something about the way uniffi builds, maybe?

Right, but IIRC that was mostly related to bindings where the uniffi build had to have access to "everything". This one seems to me it's an internal, implementation detail.

keystore/src/connection/transaction_lock.rs
typfel: Fixed in https://github.com/wireapp/core-crypto/commit/336359226f33bb399bc5eb87724819b02e3e0233

keystore/Cargo.toml
typfel: But can we somehow group the relevant cfg together? Otherwise it's a mess to understand which cfg:s belong together.

Maybe in future android also wants this or macos.

istankovic: Sure, I think we should anyway group things regardless of the specific `cfg` variable we're using. IOW, what Peter said in https://github.com/wireapp/core-crypto/pull/2490#discussion_r3821236850:
```rust
#[cfg(feature = "cross-process-lock")]
{
  mod impl;
  pub use impl::*;
}

#[cfg(not(feature = "cross-process-lock"))]
{
  mod dummy;
  pub use dummy::*;
}
```
That way we only have one place with the `cfg` and it's trivial to later update according to other platforms' needs.

keystore/src/connection/transaction_lock.rs
```diff
@@ -0,0 +1,240 @@
...
+     keystore (backed by IndexedDB) has none of; disable it for `target_os = \"unknown\"`"
+);
+
+use std::sync::Arc;
+
+use async_lock::{Semaphore, SemaphoreGuardArc};
+
+use crate::CryptoKeystoreResult;
+#[cfg(feature = "cross-process-lock")]
+pub(crate) use crate::connection::file_lock::remove_lock_file;
```
coriolinus: Could be `use super::file_lock::remove_lock_file;`

Kind of ambivalent about `super` vs an explicit full path. I do think we should probably not reexport the function here.

keystore/src/connection/mod.rs
```diff
@@ -112,13 +113,18 @@ impl Database {
         }
 
         migrations::run_migrations(&mut conn, migration_target)?;
+
+        // Prefer SQLite's own idea of the path over whatever the caller passed in: it is absolute,
+        // so two processes which opened the same database from different working directories still
+        // agree on which lock file guards it.
+        let transaction_lock = TransactionLock::new(conn.path().unwrap_or_default())?;
```
coriolinus: I don't think I agree with the reasoning in this comment: even if two processes reach the same database by two different relative paths, they'll still generate the same lockfile path. 

`conn.path()` is still the reasonable way to get the connection path given the context of this function; I'd just remove the comment.

keystore/src/connection/mod.rs
```diff
@@ -200,6 +209,10 @@ impl Database {
         if let Some(path) = location {
             // not in-memory
             fs.delete(&path).await?;
+            // `_guard` is still alive, so we unlink the lock file while still holding it; that
+            // keeps a peer from acquiring the fresh lock file before the database is really gone.
+            #[cfg(feature = "cross-process-lock")]
+            transaction_lock::remove_lock_file(&path).await?;
```
coriolinus: Why not call this directly from `file_lock`?


```suggestion
            file_lock::remove_lock_file(&path).await?;
```

coriolinus (approved):
Nice work! I do have more comments but none of them are critical. Go ahead once CI is green.

keystore/src/connection/mod.rs
typfel: Fixed by https://github.com/wireapp/core-crypto/pull/2490/commits/640c00efde7c49dcbcc9373d4b2e7f730b40c138

keystore/src/connection/mod.rs
typfel: True, removed.

keystore/src/connection/transaction_lock.rs
typfel: Removed in https://github.com/wireapp/core-crypto/pull/2490/changes/ea79df48f6e47c0d5ff3539b9bcebce0689c1982

keystore/Cargo.toml
typfel: We decided to remove the dummy implementation and make the file fields feature gated so there's more cfg:s now.

PR #2491, opened 2026-08-20
chore: review testing updates for each entity [WPB-9649]

coriolinus:
There was one entity where update testing made sense and we weren't previously doing it. Everything else which ignores it did so properly.


istankovic (approved):

coriolinus:
Stored credentials realistically should never be updated; the whole point is to tie together a bunch of data cryptographically. A better approach: actively reduce the number of entities which support upserts.

PR #2494, opened 2026-08-21
chore: reduce entities with upsert semantics [WPB-9649]

coriolinus:
Not every entity really needed to be able to upsert, but we hadn't previously done the exercise of figuring out which was which. Now we have, and we're testing for both the presence and the absence of upsertion.


keystore/tests/z_entities.rs
```diff
@@ -28,7 +28,7 @@ macro_rules! borrowed_key_round_trip {
 }
 
 macro_rules! test_for_entity {
-    ($test_name:ident, $entity:ident $(ignore_entity_count:$ignore_entity_count:literal)? $(ignore_update:$ignore_update:literal)? $(ignore_remove:$ignore_remove:literal)? $(ignore_find_many:$ignore_find_many:literal)? $(no_borrowed_key:$no_borrowed_key:literal)?) => {
+    ($test_name:ident, $entity:ident $(ignore_entity_count:$ignore_entity_count:literal)? $(ignore_update:$ignore_update:literal)? $(no_upsert:$no_upsert:literal)? $(ignore_remove:$ignore_remove:literal)? $(ignore_find_many:$ignore_find_many:literal)? $(no_borrowed_key:$no_borrowed_key:literal)?) => {
```
istankovic: Can we make this line more readable? For instance, by splitting the line somewhere sane, or even by macro arguments/options....

keystore/tests/z_entities.rs
```diff
@@ -256,6 +259,50 @@ mod tests_impl {
...
     }
 
+    /// The mirror image of [`can_update_entity`], for entities whose primary key determines their
+    /// contents: rewriting one under its existing key must fail rather than clobber it.
+    ///
+    /// The rejection happens at commit rather than at `save`, because `save` only buffers an
+    /// operation; nothing reaches SQL until the transaction is applied. The whole transaction is
+    /// therefore lost, which is the intended outcome — for these entities a duplicate key with
+    /// different contents is a bug, not a caller error to be recovered from.
+    pub(crate) async fn cannot_update_entity<E>(store: &Arc<CryptoKeystore>, entity: &E)
```
istankovic, line 273: Thank you so much for this test and the nice comment! This is truly great.

keystore/src/entities/mls/stored_psk_bundle.rs
istankovic, line 18: I noticed that we implement `Entity` for `StoredPskBundle` manually, rather than deriving it using `core_crypto_macros::Entity` -- is there a reason for that?

keystore/src/entities/mls/persisted_mls_pending_group.rs
```diff
@@ -4,6 +4,15 @@ use zeroize::Zeroize;
...
 /// Entity representing a temporarily persisted `MlsGroup`
+///
+/// This entity keeps its upsert semantics deliberately. Nothing updates one in place — there is a
+/// single writer, and the row is deleted once the external join it represents is merged or
+/// abandoned — so it reads like a candidate for a plain `INSERT`. It is not. `join_by_external_commit`
+/// invites the caller to retry, and whether the abandoned row is still there on the retry depends on
+/// what the caller did with the previous failure: an error crossing the FFI becomes an exception
+/// which by default cancels the transaction, discarding the row, but a caller may catch it and carry
+/// on with the same transaction, in which case the row survives. Both are legitimate, so the second
+/// save has to be allowed to replace the first.
```
istankovic, line 15: This kind of comment is incredibly helpful. 💜 🙏

istankovic (approved):
Left some minor comments/questions, but this looks really, really good. Thanks!

keystore/src/entities/mls/stored_psk_bundle.rs
coriolinus: The actual PK is called `id_sha256` and doesn't appear in the struct; the entity macro can't deal with that.

keystore/tests/z_entities.rs
coriolinus: ea523fb685

PR #2496, opened 2026-08-24
refactor: replace specialized pending conversation fns with generic [WPB-20731]

coriolinus:
The PR description just says to change the function signatures, but we don't need those functions at all.
We've put some effort into writing good general-purpose abstractions for bulk operations; this PR puts them to the test for the first time, and everything works out.

Also adds `ConversationIdRef` because we don't need a `'static` type for searching.


keystore/src/traits/searchable_entity.rs
```diff
@@ -30,7 +33,19 @@ pub trait SearchableEntity<SearchKey>: Entity {
...
 /// While the trait design does not require it, implementations should take advantage of
 /// database features such as indices to ensure that deletion by a search key is efficient.
-pub trait DeletableBySearchKey<SearchKey>: SearchableEntity<SearchKey> + EntityDatabaseMutation {
+///
+/// There is an asymmetry between this and [`SearchableEntity`]: in the former case,
+/// `SearchableEntity: ?Sized` because searching is an immediate operation, so there are no
+/// lifetime issues performing a search with a borrowed key. Here, however, we need to store
+/// the search key for later application (see [`Transaction::bulk_remove`][crate::Transaction::bulk_remove]),
+/// which requires the search key type to have a static lifetime because we can't know when
+/// the transaction will eventually be committed. That lifetime bound propagates here just
```
SimonThormeyer, line 42: Wouldn't it be enough if the search key lived as long as the transaction, not as long as `'static`? Not sure if that would bring any benefit or just introduce more unnecessary complexity, though.

SimonThormeyer (approved):
This is very straightforward, just a single question that I'm experimenting myself with now.

keystore/src/traits/searchable_entity.rs
coriolinus: Strictly speaking, yes, but a transaction itself has a `'static` lifetime (it borrows nothing), so it's a distinction without a difference. Nothing stops you from constructing a `Database`, opening a transaction on it, parking it in a `static` global, and then doing `loop {}` for the rest of time.

PR #2497, opened 2026-08-25
chore(crypto): use a better pattern for matching `TntWireFormat`

coriolinus:
Turns out that match statements can match ranges, but only literal ranges. Storing a range as a value means it doesn't match anymore.

### DIFF
--- a/crypto/src/mls/conversation/mutable/tnt/mod.rs
+++ b/crypto/src/mls/conversation/mutable/tnt/mod.rs
@@ -55,7 +55,10 @@ impl TntWireFormat {
     pub(crate) const TARGETED_MESSAGE: u16 = 0xF001;
     pub(crate) const TRANSIENT_TARGETED_MESSAGE: u16 = 0xF002;
 
-    pub(crate) const ALL: std::ops::RangeInclusive<u16> = Self::TRANSIENT_MESSAGE..=Self::TRANSIENT_TARGETED_MESSAGE;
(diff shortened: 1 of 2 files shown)


crypto/src/mls/conversation/mutable/tnt/mod.rs
```diff
@@ -55,6 +55,9 @@ impl TntWireFormat {
     pub(crate) const TARGETED_MESSAGE: u16 = 0xF001;
     pub(crate) const TRANSIENT_TARGETED_MESSAGE: u16 = 0xF002;
 
+    pub(crate) const MIN: u16 = Self::TRANSIENT_MESSAGE;
+    pub(crate) const MAX: u16 = Self::TRANSIENT_TARGETED_MESSAGE;
+
     pub(crate) const ALL: std::ops::RangeInclusive<u16> = Self::TRANSIENT_MESSAGE..=Self::TRANSIENT_TARGETED_MESSAGE;
```
SimonThormeyer: We should use `MIN` and `MAX` in `ALL`, so we don't have to update two constants when adding a new variant.

coriolinus: https://github.com/wireapp/core-crypto/pull/2497/commits/a69299bf4895ee3c3d45fc9a6d9f3f4ca4f6c9e5

SimonThormeyer (approved):

PR #2500, opened 2026-08-25
ci: push benchmarks results to prometheus

typfel:
Push benchmarks results to prometheus so that we can visualise them using grafana. This also includes some work to unify the benchmarks so that we display everything using the same grafana query.

The collected data is visualized on this dashbord: https://grafana.zinfra.io/d/ja65pzb/corecrypto-benchmarks?orgId=1&from=now-2d&to=now&timezone=browser&var-messageCount=1&var-messageSize=16&var-platform=jvm&var-messageBenchmark=$__all&var-userBenchmark=$__all&var-userCount=1

### DIFF
--- a/scripts/benchmarks-to-openmetrics.py
+++ b/scripts/benchmarks-to-openmetrics.py
@@ -1,0 +1,452 @@
+#!/usr/bin/env python3
+"""Convert JMH and tinybench benchmark results into Prometheus OpenMetrics 1.0 text.
+
+Usage:
+    scripts/benchmarks-to-openmetrics.py -f {jmh,tinybench} [RESULTS ...] [-o OUTPUT]
+
+RESULTS may be a JSON file holding an array of benchmark results, or a
+directory of such files (merged, read in sorted filename order). It defaults to
+the output location of the harness given by --format:
+
+    crypto-ffi/bindings/jvm/build/reports/jmh/results.json   (--format jmh)
+    crypto-ffi/bindings/js/benches_result                    (--format tinybench)
+
+One invocation handles one harness; results whose shape does not match
+--format are an error rather than something to guess at.
+
+Both harnesses map onto one schema, all gauges:
+
+    benchmark_score                    the primary metric
+    benchmark_score_error              absolute uncertainty of the score
+    benchmark_samples                  number of measurements collected
+
+The harnesses differ in one more way worth spelling out: JMH reports either
+throughput or latency depending on the benchmark's mode, while tinybench
+reports both at once. So a tinybench task becomes two series that differ only
+in `mode`/`unit` -- `mode="thrpt"` in ops/s and `mode="avgt"` in ns/op --
+which is exactly how JMH would model the same pair of measurements.
+"""
+
+import argparse
+import json
+import math
+import re
+import sys
+from collections.abc import Callable, Iterable
+from pathlib import Path
+from typing import Any, NamedTuple, TypeAlias
+
+SCORE = "benchmark_score"
+SCORE_ERROR = "benchmark_score_error"
+SAMPLES = "benchmark_samples"
+
+SCORE_ERROR_HELP = (
+    "Absolute margin of error of the score, in the score unit. The confidence "
+    "label gives the interval: JMH reports a 99.9% mean error, tinybench a 95% "
+    "relative margin of error, so the two are not directly comparable."
+)
+
+# JMH's scoreError is the half-width of a 99.9% confidence interval; tinybench
+# derives its rme from Benchmark.js' 95% t-table.
+JMH_CONFIDENCE = "0.999"
+TINYBENCH_CONFIDENCE = "0.95"
+
+# tinybench measures in milliseconds; toCustomBenchmarkEntries() reports
+# latency in nanoseconds, which is JMH's `ns/op` under another name.
+TINYBENCH_LATENCY_UNIT = "ns/op"
+
+# Keys of the `Label: value ± error` lines that toCustomBenchmarkEntries()
+# packs into a result's `extra` field.
+EXTRA_AVERAGE_LATENCY = "Average Latency (ns)"
+EXTRA_MEDIAN_LATENCY = "Median Latency (ns)"
+EXTRA_MEDIAN_THROUGHPUT = "Median Throughput (ops/s)"
+EXTRA_SAMPLES = "Samples"
+
+NAN = float("nan")
+
+# A leading decimal number, optionally in exponent notation.
+NUMBER = re.compile(r"[-+]?(?:\d+\.?\d*|\.\d+)(?:[eE][-+]?\d+)?")
+
+Sample: TypeAlias = tuple[dict[str, str], float]
+
+class Measurement(NamedTuple):
+    """A value plus its uncertainty, both in the value's own unit."""
+
+    value: float
+    error: float
+
+
+MISSING = Measurement(NAN, NAN)
+
+
+class Series(NamedTuple):
+    """One benchmark measurement under one mode, ready to emit."""
+
+    labels: dict[str, str]
+    confidence: str
+    score: Measurement
+    samples: float
+
+
+def escape_label_value(value: str) -> str:
+    """Escape a label value per the OpenMetrics text format."""
+    return value.replace("\\", "\\\\").replace('"', '\\"').replace("\n", "\\n")
+
+
+def format_labels(labels: dict[str, str]) -> str:
+    """Render a label set as `{k="v",...}`, sorted for stable output."""
+    if not labels:
+        return ""
+    inner = ",".join(
+        f'{key}="{escape_label_value(str(value))}"'
+        for key, value in sorted(labels.items())
+    )
+    return "{" + inner + "}"
+
+
+def format_value(value: float) -> str:
+    """Render a float the way OpenMetrics expects (finite -> decimal)."""
+    if math.isnan(value):
+        return "NaN"
+    if value == float("inf"):
+        return "+Inf"
+    if value == float("-inf"):
+        return "-Inf"
+    return repr(value)
+
+
+def sample(name: str, labels: dict[str, str], value: float) -> str:
+    return f"{name}{format_labels(labels)} {format_value(value)}"
+
+
+def to_float(value: Any) -> float:
+    """Coerce a JSON scalar to a float, mapping anything else to NaN.
+
+    Both harnesses can put "NaN" in as a JSON *string* -- JMH for a scoreError
+    with a single measurement, tinybench wherever a table cell was empty.
+    """
+    if isinstance(value, bool) or value is None:
+        return NAN
+    try:
+        return float(value)
+    except (TypeError, ValueError):
+        return NAN
+
+
+def parse_measurement(text: Any) -> Measurement:
+    """Parse a `12390123 ± 0.83%` / `12400000 ± 200000` / `81` reading.
+
+    A percentage error is relative to the value, so it gets scaled into the
+    value's unit; that keeps benchmark_score_error in the score unit for every
+    harness.
+    """
+    if not isinstance(text, str):
+        return Measurement(to_float(text), NAN)
+
+    raw_value, _, raw_error = text.partition("±")
+    value = to_float(raw_value.strip())
+    error = to_float(raw_error.replace('%', '').strip())
+    if "%" in raw_error:
+        error = abs(value) * error / 100
+    return Measurement(value, error)
+
+
+def parse_extra(extra: str) -> dict[str, Measurement]:
+    """Parse tinybench's newline-separated `Label: value ± error` block."""
+    readings: dict[str, Measurement] = {}
+    for line in extra.splitlines():
+        label, separator, reading = line.partition(":")
+        if separator:
+            readings[label.strip()] = parse_measurement(reading)
+    return readings
+
+
+def parse_name_labels(name: str) -> dict[str, str]:
+    """Split a tinybench result name into `benchmark` plus parameter labels.
+
+    Names look like `Add a User - cipherSuite=Mls128... userCount=1`: the task
+    name after the separator is the parameter set the benchmark ran with, which
+    is where JMH would have a `params` object. Anything in there that isn't a
+    `key=value` pair is kept verbatim under a `task` label rather than silently
+    dropped.
+    """
+    benchmark, separator, task = name.partition(" - ")
+    labels: dict[str, str] = {}
+    leftovers: list[str] = []
+
+    for token in task.split() if separator else []:
+        key, is_pair, value = token.partition("=")
+        if is_pair and key:
+            labels[key] = value
+        else:
+            leftovers.append(token)
+
+    if leftovers:
+        labels["task"] = " ".join(leftovers)
+
+    labels["benchmark"] = benchmark if separator else name
+    return labels
+
+
+def identity(labels: dict[str, str], harness: str, mode: str, unit: str) -> dict[str, str]:
+    """Complete a parameter label set with the labels this script controls.
+
+    Applied last, so a benchmark parameter named `mode` or `harness` cannot
+    shadow the identity of the series it belongs to.
+    """
+    return {**labels, "harness": harness, "mode": mode, "unit": unit}
+
+
+def jmh_sample_count(entry: dict[str, Any], primary: dict[str, Any]) -> float:
+    """Best available measurement count for a JMH result.
+
+    Sampling modes report `sampleCount` directly; for the others the number of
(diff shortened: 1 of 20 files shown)


scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+
+def format_labels(labels: dict[str, str]) -> str:
+    """Render a label set as `{k="v",...}`, sorted for stable output."""
+    if not labels:
+        return ""
+    inner = ",".join(
+        f'{key}="{escape_label_value(str(value))}"'
+        for key, value in sorted(labels.items())
+    )
+    return "{" + inner + "}"
```
SimonThormeyer, line 104: Nit: Why not use an f-string here?

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+
+
+def format_value(value: float) -> str:
+    """Render a float the way OpenMetrics expects (finite -> decimal)."""
+    if value != value:  # NaN
```
SimonThormeyer: Nit: using `float('nan')` would be more consistent with the other conditions

SimonThormeyer (approved):

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+        return "NaN"
+    if value == float("inf"):
+        return "+Inf"
+    if value == float("-inf"):
+        return "-Inf"
+    return repr(float(value))
```
istankovic: Why call `float` on a type that is already `float`?

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+    try:
+        return float(value)
+    except (TypeError, ValueError):
+        return NAN
+
+def parse_number(text: Any) -> float:
+    """Read the leading number out of `text`, or NaN if there isn't one."""
+    if isinstance(text, str):
+        match = NUMBER.match(text.strip())
+        return float(match.group()) if match else NAN
```
istankovic: When checking against `None`, it is preferable to do an explicit check:
```python
... if match is not None else NAN
```

This is to distinguish from cases where objects might evaluate to `True` in boolean contexts.

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+def to_float(value: Any) -> float:
+    """Coerce a JSON scalar to a float, mapping anything else to NaN.
+
+    Both harnesses can put "NaN" in as a JSON *string* -- JMH for a scoreError
+    with a single measurement, tinybench wherever a table cell was empty.
+    """
+    if isinstance(value, bool) or value is None:
+        return NAN
+    if isinstance(value, (int, float)):
+        return float(value)
```
istankovic: These two lines are redundant given the lines below.

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+    if opts.output is None:
+        sys.stdout.write(exposition)
+    else:
+        opts.output.write_text(exposition)
+
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
```
istankovic: This is not idiomatic, why not just
```python
main()
```
?

scripts/benchmarks-to-openmetrics.py
istankovic: Note that `float('nan') == float('nan')` evaluates to `False`.
A better way to check for NaN is to just use `math.isnan()` function.

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+
+def read_tinybench(entry: dict[str, Any]) -> list[Series]:
+    name_labels = parse_name_labels(str(entry.get("name", "")))
+    extra = parse_extra(entry.get("extra"))
+    range_ = entry.get("range")
+
+    # `value` is the mean throughput and `range` its margin of error, as a
+    # percentage of the value.
+    throughput = parse_measurement(
+        f"{entry.get('value')} ± {range_}" if range_ else entry.get("value")
```
istankovic: We should check `range_` explicitly for `None` here.
By the way, why `range_` and not `range`?

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+        series("thrpt", str(entry.get("unit", "")), throughput, median_throughput),
+        series("avgt", TINYBENCH_LATENCY_UNIT, latency, median_latency),
+    ]
+
+
+class Harness(NamedTuple):
+    """Everything that differs between the two supported result formats."""
+
+    name: str
+    default_input: Path
```
istankovic, line 281: Why name this `default_input`? Is there another input?

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+    read: Callable[[dict[str, Any]], list[Series]]
+    # Keys a result of this format must have, used to reject a mismatched
+    # --format with an actionable message instead of a pile of NaNs.
+    required_keys: tuple[str, ...]
+
+
+HARNESSES = {
+    harness.name: harness
+    for harness in (
+        Harness(
```
istankovic: This is unnecessary, we can just make dict literal instead of iterating over the items, they are already there.

scripts/benchmarks-to-openmetrics.py
SimonThormeyer: True, `math.isnan()` would be best here

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+which is exactly how JMH would model the same pair of measurements.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import re
+import sys
+from decimal import Decimal, InvalidOperation
```
istankovic: I could not find usage of these anywhere.

scripts/benchmarks-to-openmetrics.py
SimonThormeyer: IIRC, `range` is python standard name

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+    name: str, help_text: str, samples: Iterable[tuple[dict[str, str], float]]
+) -> list[str]:
+    """Render one metric family, omitting it entirely if it has no readings."""
+    rendered = [
+        sample(name, labels, value)
+        for labels, value in samples
+        # A family that nothing reported is better left out than filled with
+        # NaN: it lets one exposition carry both harnesses without inventing
+        # measurements neither of them made.
+        if value == value
```
istankovic: This should use `math.isnan()` to check for NaNs.

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,468 @@
...
+                file=sys.stderr,
+            )
+            continue
+        seen.add(key)
+        unique.append(one)
+    return unique
+
+
+def family(
+    name: str, help_text: str, samples: Iterable[tuple[dict[str, str], float]]
```
istankovic: The content inside `Iterable` is a bit much, I think a type alias would be helpful here.

scripts/benchmarks-to-openmetrics.py
istankovic: I'm also wondering whether we need the regular expression at all here. What kind of strings are we expecting to get?

Do we need to handle strings like `mx1mnjki-9.812e-13kma`?

scripts/benchmarks-to-openmetrics.py
istankovic: Ah, right, good point, it would shadow the builtin function.  :+1:

scripts/benchmarks-to-openmetrics.py
typfel: It's it parses value plus error strings that looks like: `12390123 ± 0.83%` and it's split before and so it parses strings like ` 0.83%` or `12390123 `.

scripts/benchmarks-to-openmetrics.py
typfel: It's the default input path which is used if none is provided

scripts/benchmarks-to-openmetrics.py
istankovic: Hm, then something is off -- this function is called `parse_number` and it looks for the pattern defined by `NUMBER` which is documented as
```
 A leading decimal number, optionally in exponent notation.
```

If we are expecting to parse two values here, then the function is incorrect.
For example, what would you expect the function to return on input `xx12390123 ± 0.83%yy`?

typfel: Ok, i was unclear the whole string is split up in `parse_measurement()` and then it extract numbers from each string component using  `parse_number()`.

scripts/benchmarks-to-openmetrics.py
typfel: It's to prevent this code from executing if you try to import the script as a module. Nothing we intend do so I removed this guard.

scripts/benchmarks-to-openmetrics.py
typfel: Because the string includes the `{` which must be escaped like this `f"{{{inner}}}"`

scripts/benchmarks-to-openmetrics.py
istankovic: Note my comment was only about the `raise` statement, sorry for not making it clearer. I.e.
```
if __name__ == "__main__":
    main()
```
is totally fine and idiomatic.

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+def to_float(value: Any) -> float:
+    """Coerce a JSON scalar to a float, mapping anything else to NaN.
+
+    Both harnesses can put "NaN" in as a JSON *string* -- JMH for a scoreError
+    with a single measurement, tinybench wherever a table cell was empty.
+    """
+    if isinstance(value, bool) or value is None:
+        return NAN
```
istankovic, line 129: These two lines can be dropped given that such cases (and more) are covered by the `try...except` below.

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,462 @@
...
+    A percentage error is relative to the value, so it gets scaled into the
+    value's unit; that keeps benchmark_score_error in the score unit for every
+    harness.
+    """
+    if not isinstance(text, str):
+        return Measurement(to_float(text), NAN)
+
+    raw_value, _, raw_error = text.partition("±")
+    value = parse_number(raw_value)
+    error = parse_number(raw_error) if raw_error.strip() else NAN
```
istankovic: I suggest the following:
```python
value = to_float(raw_value.strip())
error = to_float(raw_error.strip())
```
and then we can remove `parse_number` and the regex pattern completely.

scripts/benchmarks-to-openmetrics.py
istankovic: Ah... made another comment: https://github.com/wireapp/core-crypto/pull/2500/changes#r3880013885 and am  going to close this one.

scripts/benchmarks-to-openmetrics.py
typfel: Ah ok, but we should return the exit code and from what I've seen have to do either `raise SystemExit()` or `sys.exit()`

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,462 @@
...
+    if "%" in raw_error:
+        error = abs(value) * error / 100
+    return Measurement(value, error)
+
+
+def parse_extra(extra: Any) -> dict[str, Measurement]:
+    """Parse tinybench's newline-separated `Label: value ± error` block."""
+    readings: dict[str, Measurement] = {}
+    if not isinstance(extra, str):
+        return readings
```
istankovic: I would move this check to the caller and require that `parse_extra` receives a `str`. Note that many functions receive `Any` argument type, which is bad practice and forces every function to have a number of checks. (`Any` does not really say anything about the type and one might as well just omit it. But it's better to ensure correct types are passed to functions and make sure the callers do the right thing.)

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,462 @@
...
+    benchmark_samples                  number of measurements collected
+
+The harnesses differ in one more way worth spelling out: JMH reports either
+throughput or latency depending on the benchmark's mode, while tinybench
+reports both at once. So a tinybench task becomes two series that differ only
+in `mode`/`unit` -- `mode="thrpt"` in ops/s and `mode="avgt"` in ns/op --
+which is exactly how JMH would model the same pair of measurements.
+"""
+
+from __future__ import annotations
```
istankovic: Why do you need this? Which version of Python you're using?

scripts/benchmarks-to-openmetrics.py
typfel: Doesn't work when value contains a `%` character which can happen. So it would be something like 

```
value = to_float(raw_value.strip())
error = to_float(raw_error.replace('%', '').strip())
```

scripts/benchmarks-to-openmetrics.py
istankovic: > Ah ok, but we should return the exit code and from what I've seen have to do either `raise SystemExit()` or `sys.exit()`

Some people might think that's actually required, but that is not the case.  ;-)

In fact, likely the only case where you need to explicitly use `sys.exit()` or raise the `SystemExit` exception (which is really the same thing) is when you want to communicate special exit code to the parent process, or when you want to terminate the interpreter immediately.

However, those are special cases and the majority of programs do not need to think or care about it.
If there was no exception raised during a Python program execution, the Python interpreter will just exit successfully and you will implicitly get a 0 exit code from the process. If there was an exception, you will get a failure exit code:

```
$ python -c 'pass'; echo $?
0

$ python -c '1/0'; echo $?
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    1/0
    ~^~
ZeroDivisionError: division by zero
1
```

typfel: It's not quite the same though because:
```
$ python3 -c '1'; echo $?
0
```

Which means if someone attempts to change the exit code by changing the value which is returned from main it will not work.

I would also stop returning a value from main to avoid misleading anyone that it will be the exit code.

istankovic: > It's not quite the same though because:
> 
> ```
> $ python3 -c '1'; echo $?
> 0
> ```
> 
> Which means if someone attempts to change the exit code by changing the value which is returned from main it will not work.

Exactly. `main` in Python is just another function, unlike in, say, C.

> I would also stop returning a value from main to avoid misleading anyone that it will be the exit code.

Right, that should definitely be removed. :+1:

scripts/benchmarks-to-openmetrics.py
typfel: Boolean values would be converted to "1.0" instead of NAN then.

istankovic: Hm, right.

scripts/benchmarks-to-openmetrics.py
```diff
@@ -0,0 +1,462 @@
...
+def identity(labels: dict[str, str], harness: str, mode: str, unit: str) -> dict[str, str]:
+    """Complete a parameter label set with the labels this script controls.
+
+    Applied last, so a benchmark parameter named `mode` or `harness` cannot
+    shadow the identity of the series it belongs to.
+    """
+    return {**labels, "harness": harness, "mode": mode, "unit": unit}
+
+
+def jmh_sample_count(entry: dict[str, Any], primary: dict[str, Any]) -> float:
```
istankovic, line 200: What are the effects of this function returning `NaN`?

typfel: It gets skipped when the sample is rendered https://github.com/wireapp/core-crypto/blob/684f7c3dd9820e945a61498ea08871ba68481e61/scripts/benchmarks-to-openmetrics.py#L376

scripts/benchmarks-to-openmetrics.py
typfel: We don't need it

istankovic (approved):
Looks good, thanks.

PR #2502, opened 2026-08-25
refactor: add custom PK to epoch encryption keypair pk [WPB-28254]

coriolinus:
Adding a custom primary key means we don't need to combine fields into a new vector when searching for this entity; we can just search by the borrowed reference.


keystore/src/connection/migrations/meta_migrations/v34.rs
```diff
@@ -0,0 +1,56 @@
...
+    Ok(())
+}
+
+/// [`StoredEpochEncryptionKeypair`][crate::entities::StoredEpochEncryptionKeypair] as it appeared in v34
+#[derive(core_crypto_macros::Entity)]
+#[entity(table_name = "mls_epoch_encryption_keypairs")]
+struct StoredEpochEncryptionKeypair {
+    id: Vec<u8>,
+    keypairs: Vec<u8>,
+}
```
SimonThormeyer: This is exactly the same as the `V33StoredEpochEncryptionKeypair` struct we introduced for the legacy migrations. Isn't the use case similar enough to share it?

coriolinus: It gets removed later when I notice that.

SimonThormeyer: Ah, thanks, I missed that. Could you please fold the removal into the commit that added it?

SimonThormeyer (approved):
Good work! A single history-related nit that I'd appreciate to be resolved.

keystore/src/connection/migrations/V34__split_epoch_encryption_keypair_pk.sql
```diff
@@ -0,0 +1,20 @@
...
+-- We don't do that today because ultimately this is an OpenMLS table and that
+-- pipeline is built into the OpenMLS assumptions. It's been like that since
+-- 51a7e1393ac277f0e621b5d2866c666edf74c687 in 2023, and changing it would require
+-- adjusting the `MlsEntity` and `OpenMlsKeyStore` traits.
+CREATE TABLE epoch_encryption_keypairs (
+    conversation_id BLOB NOT NULL,
+    own_leaf_index INTEGER NOT NULL,
+    epoch INTEGER NOT NULL,
+    keypairs BLOB NOT NULL,
+    PRIMARY KEY (conversation_id, own_leaf_index, epoch)
```
SimonThormeyer, line 19: Is there a reason we're not making the `conversation_id` a `FOREIGN KEY` with `ON DELETE CASCADE`?

coriolinus: Good point! https://github.com/wireapp/core-crypto/pull/2502/commits/69ec33f6db386241646e2dea61bea699f59422da

keystore/src/connection/migrations/meta_migrations/v34.rs
coriolinus: Done.

keystore/src/connection/migrations/V34__split_epoch_encryption_keypair_pk.sql
coriolinus: Ended up reverting that because it turns out this is not a trivial win. It looks like there are two separate problems with this:

- We do not eagerly store new conversations (behavior from before transactions), which causes the keypairs to fail because the conversation has not been saved yet
- Pending conversations also appear to be able to have keypairs

PR #2504, opened 2026-08-26
fix(tnt/targeted): when decrypting, take protocol version from message

SimonThormeyer:
... and error out if it doesn't match `V1`.


coriolinus (approved):
I'm not sure that we need to spell out `new_with_current_protocol_version`; that's kind of what people expect from a bare `new` function anyway. But it's harmless to spell it out.

PR #2505, opened 2026-08-26
chore: rm e2ei legacy module [WPB-26519]

coriolinus:
This PR is not sufficient to close that ticket, but it's as much work as can be done without leading to merge conflicts.


e2e-identity/src/x509_check/mod.rs
```diff
@@ -50,7 +50,7 @@ impl From<certval::Error> for RustyX509CheckError {
 
 pub type RustyX509CheckResult<T> = Result<T, RustyX509CheckError>;
 
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
```
istankovic, line 53: Why do we need `Ord` and `PartialOrd` here?

coriolinus: Because we now use this instead of the legacy wrapper in `crypto`, and we stick an instance in structs that derive `Ord`.

istankovic: Can you give concrete examples of such structs?
I'm not sure that `Ord`/`PartialOrd` make sense for this, and even less for something that contains these.

coriolinus: https://github.com/wireapp/core-crypto/blob/3a8391478a2386ed157d8044eedbb3ee2eb188e9/crypto/src/identity.rs#L9-L21

istankovic: Why do we need `*Ord` on `WireIdentity`?

coriolinus: IDK, that's outside the scope of changes I was looking at. It's preexisting.

coriolinus: Looks like it comes from [here](https://github.com/wireapp/core-crypto/commit/55b75feee671d044272348e77fcf6eee8b93529d).

istankovic: Going to make an item to remove ordering from both status and identity types, for now let's keep things as they were.

istankovic (approved):

PR #2506, opened 2026-08-26
chore: add THIRD PARTY NOTICE - WPB-25975

marcoconti83:
# What's new in this PR?

Adding a THIRD PARTY NOTICES file. This was generated with an internal tool, that requires a bit of per-project customization. I'm also committing the per-project configuration file for that tool.


typfel:
How is the THIRD PARTY NOTICE updated when we add/remove a dependency?

coriolinus (approved):
The files are fine, but how do we run the internal tool in CI to keep this up to date?

istankovic (dismissed):
I don't think we should do it this way -- third party notices content should not be commited to the repo.

PR #2507, opened 2026-08-27
docs: add a section about object lifetimes WPB-26866

typfel:


cc-book/src/concepts/client/overview.md
```diff
@@ -26,6 +26,30 @@ operations available on `CoreCrypto` is intentionally kept narrow.
 Any operation that might change state — creating or deleting a conversation, adding members, encrypting or decrypting a
 message — requires a transaction.
 
+## Lifetimes
+
+The lifetime for objects created by `CoreCrypto` vary slightly depending on the platform. Objects have a rust
+counterpart, and it might hold on to resources like files or database connections. Exactly when the rust counterpart and
```
SimonThormeyer: `Exactly when` -> `When exactly` ? I don't know for a fact which one is more idiomatic, the latter sounds more natural to me

SimonThormeyer: It's slightly unclear to me, what exactly `it` is referring to in "it might hold on to...". Might help a bit to just repeat the concrete thing instead of "it".

SimonThormeyer (approved):
Two remarks, nothing blocking

PR #2508, opened 2026-08-27
feat(tnt): transient messages [WPB-27215]

SimonThormeyer:
Encryption and decryption of transient messages.


keystore/src/entities/mls/transient_message_tx_counter.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use crate::{
+    CryptoKeystoreResult,
+    entities::helpers::{count_helper, delete_helper, get_helper, get_helper_composite_key, load_all_helper},
+    traits::{BorrowPrimaryKey, EntityDatabaseMutation, EntityDeleteBorrowed, EntityGetBorrowed, PrimaryKey},
+};
+
+#[derive(core_crypto_macros::Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct TransientMessageTxCounter {
+    pub conversation_id: Vec<u8>,
```
coriolinus: It's not wrong, but I wonder why you chose not to use the `ConversationId` newtype?

keystore/src/entities/mls/transient_message_tx_counter.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn from_row(row: &rusqlite::Row<'_>) -> rusqlite::Result<Self> {
+        Ok(Self {
+            conversation_id: row.get("conversation_id")?,
+            count: row.get("count")?,
+        })
+    }
+}
+
+impl PrimaryKey for TransientMessageTxCounter {
+    type PrimaryKey = TransientMessageTxCounterPk;
```
coriolinus: Why define a custom type here? This is what the `ConversationId` newtype was made for, and gives you `ConversationIdRef` for free.

keystore/src/entities/mls/transient_message_tx_counter.rs
```diff
@@ -0,0 +1,119 @@
...
+impl crate::traits::Entity for TransientMessageTxCounter {
+    const TABLE_NAME: &str = Self::TABLE_NAME;
+
+    fn get(
+        conn: &rusqlite::Connection,
+        key: &TransientMessageTxCounterPk,
+    ) -> crate::CryptoKeystoreResult<Option<Self>> {
+        get_helper_composite_key(
+            conn,
+            &[Self::PRIMARY_KEY_COLUMN_NAME],
```
coriolinus, line 43: Using the composite key version works, but where there's a single key, it feels cleaner to use the basic version of the helper.

keystore/tests/z_entities.rs
```diff
@@ -200,11 +205,16 @@ mod tests_impl {
...
+        let mut entity = E::random();
 
         let tx = store.new_transaction().await.unwrap();
+        if let Some(counter) = (&mut entity as &mut dyn Any).downcast_mut::<TransientMessageTxCounter>() {
+            let group = PersistedMlsGroup::random();
+            counter.conversation_id = group.id.clone();
+            tx.save(group).await.unwrap();
+        }
+        let primary_key = entity.primary_key();
+        let borrowed_primary_key = entity.borrow_primary_key();
```
coriolinus: You could keep the lead-in (let entity, let primary_key, let borrowed_primary_key)  by mutating `group` in this clause instead of mutating `entity`.

keystore/tests/z_entities.rs
```diff
@@ -573,6 +588,21 @@ pub mod utils {
         }
     }
 
+    impl EntityRandomExt for TransientMessageTxCounter {
```
coriolinus: Mutating the relevant group on presave instead of mutating the entity would also mean we could use the macro to impl these traits instead of implementing them manually.

crypto/src/mls/conversation/immutable/mod.rs
```diff
@@ -33,12 +36,18 @@ pub(crate) struct MlsGroupState {
...
     /// incremented. Do not access this field directly, use [MlsGroupState::obtain_targeted_message_tx_counter] and
-    /// [MlsGroupState::reset_targeted_message_tx_counters] only.
+    /// [MlsGroupState::reset_tnt_message_tx_counters] only.
     ///
     /// The purpose of these counters is replay protection on the recipient side: we provide the count when sending a
     /// message to a receiver, and they check if the counter is greater than any they've seen before.
-    #[debug(skip)]
     targeted_message_tx_counters: HashMap<LeafNodeIndex, TntMessageCounter>,
+
+    /// The count of transient messages sent (therefore `tx`) this epoch. Because all currently online members receive
```
coriolinus: Nit: "therefore" -> "hereafter".

keystore/tests/z_entities.rs
```diff
@@ -134,6 +134,16 @@ mod tests_impl {
                 let group = PersistedMlsGroup::random();
                 counter.conversation_id = group.id.clone();
 
+                tx.save(group).await.unwrap();
+            } else if let Some(counter) = any_e.downcast_mut::<TargetedMessageRxCounter>() {
+                let group = PersistedMlsGroup::random();
+                counter.conversation_id = group.id.clone();
+
+                tx.save(group).await.unwrap();
+            } else if let Some(counter) = any_e.downcast_mut::<TransientMessageRxCounter>() {
```
coriolinus: At some point it becomes worthwhile to copy the `match_heterogenous` macro over here so we can have a `match` structured block doing this kind of thing.

crypto-ffi/src/decrypted_message.rs
```diff
@@ -5,11 +5,16 @@ use core_crypto::{BufferedDecryptedMessage as CcBufferedDecryptedMessage, Decryp
 use crate::{ClientId, WireIdentity};
 
 /// Represents the items a consumer might require after decrypting a message.
+//
+// Implementation note: from a pure rust perspective, it would be desirable to use tuple-struct-like variants where
+// structure is equal. However, uniffi's generated code isn't particularly nice with those kind of variants, because it
+// generates classes with members named after their index in the tuple, which isn't idiomatic in any of our target
+// languages.
```
coriolinus, line 12: I still think the actual ideal might be something like 

```rust
pub struct PlaintextMessage {
  pub plaintext: Vec<u8>,
  pub sender_client_id: Arc<ClientId>,
  pub Identity: WireIdentity,
}

pub enum DecryptedMessage {
  MlsApplicationMessage {
    plaintext: PlaintextMessage,
  }
  Transient {
    plaintext: PlaintextMessage,
  }
  // etc
}
```

Being able to easily extract/read any plaintext message regardless of type has value.

coriolinus (commented):
Several small things, no major changes. The main thing I really want to see changed is using the existing `ConversationId` PK helper where appropriate instead of defining yet another new PK type.

keystore/src/entities/mls/transient_message_tx_counter.rs
SimonThormeyer: What do you think about removing the single key one? The usage of the more generic one looks fine imo even in this case, where the pk is just a single column.

crypto-ffi/src/decrypted_message.rs
SimonThormeyer: What would usage look like then?
e.g., in swift

```swift
switch decryptedMessage {
case let .applicationMessage(plaintext):
    // Get the plaintext
    let plaintext = plaintext.plaintext;
// ... other cases
}
```

isn't that worse than the status quo

```swift
switch decryptedMessage {
case let .applicationMessage(plaintext, senderClientId, identity):
    // We already have the plaintext
// ... other cases
}
```

keystore/tests/z_entities.rs
SimonThormeyer: b448feb80abddde6d334fa62277fba26d211dd6f

keystore/tests/z_entities.rs
SimonThormeyer: 95cb67d2c0cc08db676b4e27bf23354cb4f06488

keystore/tests/z_entities.rs
SimonThormeyer: 95cb67d2c0cc08db676b4e27bf23354cb4f06488

keystore/tests/z_entities.rs
SimonThormeyer: 95cb67d2c0cc08db676b4e27bf23354cb4f06488

keystore/src/entities/mls/transient_message_tx_counter.rs
coriolinus: Removing the single-key one is fine technically, but just adds churn + noise to those entities which only require a single key. I'm a bit against it.

crypto-ffi/src/decrypted_message.rs
coriolinus: You may be right; I don't have a rebuttal to that point. But this is itching at me. I'll keep thinking about it. Certainly not a blocker for this PR.

keystore/src/entities/mls/transient_message_tx_counter.rs
SimonThormeyer: Okay, if we're going to keep both, I'll use the single one here and refactor the single one to call the more generic one.

crypto/src/mls/conversation/id.rs
```diff
@@ -5,19 +5,9 @@ use std::{
...
-    PartialEq,
-    Eq,
-    PartialOrd,
-    Ord,
-    Hash,
-    Clone,
+    core_crypto_macros::Debug, derive_more::From, derive_more::Into, PartialEq, Eq, PartialOrd, Ord, Hash, Clone,
 )]
 #[sensitive]
-#[as_ref([u8])]
```
coriolinus, line 20: It's a cool idea to implement `AsRef<ConversationIdRef>` for `ConversationId`, but we don't actually need to give up the ability to reference it as a byte slice to do so. We can have both `AsRef` impls simultaneously.

crypto/src/mls/conversation/immutable/mod.rs
```diff
@@ -122,11 +122,11 @@ impl MlsGroupState {
     pub(in crate::mls::conversation) async fn reset_tnt_message_tx_counters(&mut self, tx: &Transaction) -> Result<()> {
         self.targeted_message_tx_counters = Default::default();
         self.transient_message_tx_counter = Default::default();
-        let id = self.group.group_id();
-        tx.bulk_remove::<TargetedMessageTxCounter, _>(id.to_vec().into()).await;
-        tx.remove_borrowed::<TransientMessageTxCounter>(KeystoreConversationIdRef::new(id.as_slice()))
+        let id: KeystoreConversationId = self.group.group_id().to_vec().into();
```
coriolinus, line 125: Could we construct a `KeystoreConversationIdRef` here directly from the slice?

keystore/tests/z_entities.rs
```diff
@@ -14,6 +14,69 @@ macro_rules! pat_to_bool {
...
+            };
+            #[cfg(not($meta))]
+            let result = None;
+            result
+        }
+    };
+}
+
+/// This is moderately horrific and we hopefully will not require it anywhere else, but
+/// it solves a real problem here: how do we match against the innermost error variants,
```
coriolinus: "error variants" is stale docs.

coriolinus (approved):
I like what you've done here! A few more nits, but overall this is great.

crypto/src/mls/conversation/id.rs
SimonThormeyer: I saw that, but I personally found it more convenient to insert another `as_ref()` in the rare case where I need a `&[u8]` instead of a `&ConversationIdRef`, instead of, with two `AsRef` impls for `ConversationId`, having to specify the type explicitly where the compiler thinks is ambiguous (which happens more frequently than I'd expect).

crypto/src/mls/conversation/immutable/mod.rs
SimonThormeyer: Would that get us anything, since we need the owned variant for `bulk_remove()`?

PR #2517, opened 2026-08-31
refactor: custom transaction guard [WPB-28247]

coriolinus:
This PR invents a custom transaction guard type which, unlike `rusqlite::Transaction`, is perfectly suitable for long-term holding in an asynchronous context. It can't `impl Deref<Connection>`, unfortunately, but we can and do provide a helper which does do so.

This is followed by some work abstracting over the differences between our transaction guard and rusqlite's, as we don't really care which we hold as long as we have type-level assurance that a transaction exists.

Then we do a bunch of refactoring to use the Entity API everywhere; nearly all of the Transaction API has disappeared.

Some notable side effects:

- errors from keystore operations now propagate immediately, not waiting for a commit
- it is now possible to recover the error from a bulk delete operation
- we had to replace `INSERT OR REPLACE` statements with proper `INSERT ... ON CONFLICT` conflict resolution; it turns out that foreign keys don't play nicely with `INSERT OR REPLACE` and we have some now.


keystore/src/transaction/conn.rs
```diff
@@ -0,0 +1,70 @@
...
+use rusqlite::Connection;
+
+use crate::{CryptoKeystoreError, CryptoKeystoreResult, unique_arc::ArcWithReadGuard};
+
+use super::{GUARD_EXPECTATION, Transaction};
+
+/// The type of the guard we end up holding over the connection.
+type InternalGuard = ArcMutexGuard<RawMutex, Option<MutexGuardArc<Connection>>>;
+/// Keeps the wrapper alive for as long as this guard exists. `None` when the caller
+/// already holds a `&TransactionWrapper`, where borrowck provides the same guarantee.
```
SimonThormeyer, line 13: typo: `borrowck`?

coriolinus: No, that's an abbreviation for the borrow checker.

keystore/src/connection/mls.rs
```diff
@@ -0,0 +1,60 @@
...
+            .await
+            .flatten()
+    }
+
+    async fn read<V: MlsEntity>(&self, id: &[u8]) -> Option<V>
+    where
+        Self: Sized,
+    {
+        self.with_transaction(async |tx| tx.read(id).await).await.ok().flatten()
+    }
```
SimonThormeyer, line 54: Why do we need `with_transaction()` in a `read()` function?

coriolinus: 1. To ensure that reads on the DB see the (currently uncommitted) state
2. Also the DB connection is locked for the duration of the transaction, so the only path to the DB is through the transaction.

coriolinus: 3. So OpenMLS sees the same thing that CC does: `Database::conn` also goes through the transaction, if one is live.

coriolinus: Ah, you're asking why there needs to be a transaction at all, more than why we take a view into it. That's mostly for simplicity: this is a one-liner, and we know that we only invoke openmls to do things within a transaction's context. 

If we routed through `self.conn()` we could get most of the same benefits, at the drawback that we'd have to duplicate most of the entity loading code from the `Transaction` impl of `OpenMlsKeyStore`; that's 60 lines we could avoid. But now that I understand the question better, I'm less opposed to changing it if you feel strongly.

SimonThormeyer: > we know that we only invoke openmls to do things within a transaction's context

I would rephrase a little, maybe to _openmls invokes our keystore via this trait only while we're within a transaction's context_, and then wonder, can we really guarantee that? If so, I'd be fine to just add a comment here, but if we can't, I'd prefer to refactor this so that this call is independent of a transaction. 
I feel that it could just be our test setup for which this happens to work, but am unsure about real usage.

typfel: @SimonThormeyer wasn't the previous behaviour if a client does a read operation outside a transaction it will not see the changes of any concurrent ongoing transaction.

SimonThormeyer: @typfel yes, that would explain why we did it.

coriolinus: > openmls invokes our keystore via this trait only while we're within a transaction's context

Your phrasing is more accurate than mine, so we can use that. But the proof is trivial: 

- [`CoreCrypto`](https://wireapp.github.io/core-crypto/main/rust/core_crypto/struct.CoreCrypto.html) exposes two mls-related methods: `history_client` and `mls_session`
- `history_client` instantiates its own transaction internally
- `mls_session` exposes `Session` which doesn't do anything which calls into openmls

So for now I'll just add that comment.

coriolinus: I said I'd add a comment but it turns out to be simple-ish (https://github.com/wireapp/core-crypto/pull/2517/commits/4126df4e3a982f5c78911e041488bf02c468468a) to factor out the common behavior instead, which is strictly better.

keystore/src/transaction/conn.rs
```diff
@@ -54,7 +54,7 @@ impl Transaction {
     /// Guard over this transaction's connection.
     ///
     /// Errors if SQLite has already rolled this transaction back under us.
-    pub(crate) fn conn(&self) -> CryptoKeystoreResult<TransactionConnection> {
+    pub fn conn(&self) -> CryptoKeystoreResult<TransactionConnection> {
```
SimonThormeyer: Do we really want this return type to be the concrete `TransactionConnection`? Do callers need to know about this type? Or would an abstract type like `impl FetchFromDatabase` (or another type shared with the keystore connection) maybe be a better fit? 

Just thinking out loud, not a strong feeling.

coriolinus: It needs to be a concrete type, see for example `keystore/src/transactionlike.rs`.

And even if we wanted type-erasure, it would have to be `impl Deref<Target = Connection>`, not `FetchFromDatabase`. This is particularly important for transactions because they actually do need to write.

crypto/src/transaction_context/conversation/welcome.rs
```diff
@@ -31,17 +32,52 @@ impl TransactionContext {
...
+                let mut stmt = conn
+                    .prepare_cached(formatcp!("RELEASE SAVEPOINT {SAVEPOINT_NAME}"))
+                    .map_err(KeystoreError::wrap("creating savepoint release stmt"))?;
+                stmt.execute([]).map_err(KeystoreError::wrap("releasing savepoint"))?;
+            } else {
+                let mut stmt = conn
+                    .prepare_cached(formatcp!("ROLLBACK TO SAVEPOINT {SAVEPOINT_NAME}"))
+                    .map_err(KeystoreError::wrap("creating savepoint rollback stmt"))?;
+                stmt.execute([])
+                    .map_err(KeystoreError::wrap("rolling back savepoint"))?;
```
SimonThormeyer: I love the approach! But it doesn't feel entirely right that we're executing sql statements in this crate. Can we add a savepoint API on the `Transaction` type that we call here?

That would also mean that we can make `tx
                .conn()` `pub(crate)` again, I suppose?

coriolinus: Eh. A proper savepoint API on the `Transaction` type would be complicated; it would involve either a guard approximately as complicated as `Transaction` itself, or an easy-to-misuse triad of `create_savepoint()`, `release_savepoint()`, `rollback_savepoint()`. And that triad as methods on `Transaction` is strictly worse than just writing the sql in the relevant function, because it doesn't enforce the guarantees that are written out in the notes here.

coriolinus: You were right; your idea works better than I'd expected. Implemented in https://github.com/wireapp/core-crypto/pull/2517/commits/2deebea22aa47f2f62dccb2c2f3a44a7571567e4 and https://github.com/wireapp/core-crypto/pull/2517/commits/e4670640beb701aa0cc21fc4273becddcaf7a95e.

keystore/src/transaction/savepoint.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// output of this method. If that result is ok, the savepoint is released (i.e. committed).
+    /// If that result is an error, the savepoint is rolled back.
+    ///
+    /// Error-mapping is the most complicated part of this. When an error is produced in the course
+    /// of this method, we run the `map_err` function with a static context string describing what was
+    /// happening. That function must return a second function which transforms a [`CryptoKeystoreError`]
+    /// into an instance of `E`.
+    pub async fn with_savepoint<T, E>(
+        &self,
+        savepoint_name: &str,
```
SimonThormeyer, line 22: Why do we require callers to provide an explicit `savepoint_name` instead of, e.g., generating a random string for the name? No strong feelings though

keystore/src/transaction/conn.rs
SimonThormeyer: I'd slightly prefer not to leak this special type outside the keystore crate as it seems like it's an implementation detail callers shouldn't need to care about. But this is entirely optional

SimonThormeyer (approved):
Great job! I'm so happy that this works.
Just two remaining comments that are non-blocking.

keystore/src/transaction/savepoint.rs
coriolinus: Mainly because I don't know what the use cases are, and providing a name seems to support a broader set of use cases than generating one.

keystore/src/transaction/conn.rs
coriolinus: Mostly they don't have to care; they just treat it as a `Connection`. But naming the type here at least lets them store an instance in a type if they have to for some reason.

PR #2518, opened 2026-08-31
refactor(tnt): use a single `TntMessageTxCounter`

SimonThormeyer:
1. Reduces storage complexity on the sender side by sharing a tx counter between transient and targeted messages.
2. Splits rx counter tables by sender and delivery channel (transient and transient targeted share a table, persisted targeted gets its own)


crypto/src/mls/conversation/mutable/tnt/mod.rs
```diff
@@ -69,6 +74,38 @@ impl TntWireFormat {
     pub(crate) const MAX: u16 = Self::TRANSIENT_TARGETED_MESSAGE.0;
 
     pub(crate) const ALL: std::ops::RangeInclusive<u16> = Self::MIN..=Self::MAX;
+
+    /// The total received count of the given message type from the given sender, conversation, and epoch.
+    ///
+    /// Implementation note: because transient targeted message and targeted message share a type but don't share a
+    /// counter table, we're implementing this here instead of the types themselves.
+    async fn total_received_message_count(
```
coriolinus: This isn't the total count of received messages, it's the current counter for the given message type. 

Maybe name it `current_rx_counter` or something?

coriolinus (approved):
Had one nit, but this is nice.

PR #2523, opened 2026-09-02
chore: update certval, rand and RustCrypto crates (WPB-27957)

istankovic:
This updates a number of our dependencies, notably:
- x509-cert
- certval
- rand and getrandom
- RustCrypto crates

that have accumulated quite a few changes over the last several years.

### DIFF
--- a/crypto-ffi/bindings/js/Cargo.rust-modules.toml
+++ b/crypto-ffi/bindings/js/Cargo.rust-modules.toml
@@ -1,2 +1,7 @@
 [profile.release]
 opt-level = "s"
+
+# We require a pre-release version of x509-ocsp because certval depends on it.
+# Once x509-ocsp 0.3 is released and certval moves to it, drop the patch.
+[patch.crates-io]
+x509-ocsp = { git = "https://github.com/RustCrypto/formats.git", package = "x509-ocsp", version = "0.3.0-pre" }
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -105,6 +105,7 @@ macro_rules_attribute.workspace = true
 wasm-bindgen-test.workspace = true
 openmls.workspace = true
 hex-literal = "1.1"
+der.workspace = true
 
 [target.'cfg(not(target_os = "unknown"))'.dev-dependencies]
 smol-macros.workspace = true
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -60,9 +60,9 @@ obfuscate = { workspace = true, features = [
   "openmls",
   "openmls_basic_credential",
 ] }
-aes-gcm = "0.10"
+aes-gcm = "0.11"
 sha2 = { workspace = true, features = ["oid"] }
-chacha20poly1305 = "0.10"
+chacha20poly1305 = "0.11"
 hmac = "0.13"
 ed25519-dalek = { workspace = true, features = ["pkcs8"] }
 signature.workspace = true
@@ -70,7 +70,7 @@ ecdsa = { workspace = true, features = ["der", "pkcs8"] }
 p256 = { workspace = true, features = ["pkcs8"] }
 p384 = { workspace = true, features = ["pkcs8"] }
 p521 = { workspace = true, features = ["pkcs8"] }
-hkdf = "0.12"
+hkdf = "0.13"
 x509-cert = { workspace = true, features = ["builder", "hazmat"] }
 web-time.workspace = true
 rand = { workspace = true, features = ["getrandom"] }
@@ -79,8 +79,8 @@ rand_core = "0.6"
 rand_chacha = "0.3"
 
 [dependencies.hpke]
-version = "0.12"
-features = ["x25519", "p256", "p384", "p521"]
+version = "0.14"
+features = ["x25519", "nistp"]
 
 [target.'cfg(target_os = "unknown")'.dependencies]
 serde-wasm-bindgen.workspace = true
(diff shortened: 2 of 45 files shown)


crypto/Cargo.toml
```diff
@@ -66,6 +66,7 @@ chacha20poly1305 = "0.11"
 hmac = "0.13"
 ed25519-dalek = { workspace = true, features = ["pkcs8"] }
 signature.workspace = true
+elliptic-curve = "0.14"
```
coriolinus, line 69: Why not use the workspace dependency here?

e2e-identity/src/pki.rs
```diff
@@ -174,124 +89,295 @@ fn get_extended_keyusage(is_ca: bool) -> x509_cert::ext::pkix::ExtendedKeyUsage
...
+    }
+
+    add_crl_distribution_points(&mut builder, args)?;
+
+    builder
+        .build::<_, Signature>(&keypair)
+        .map_err(|_| E2eIdentityError::CertificateGenerationError)
+}
+
+fn generate_cert_intermediate<Signature: spki::SignatureBitStringEncoding, S>(
```
coriolinus, line 213: You've got two styles of generic parameter here, `Signature` and `S`. You've got the bounds for each of them in a different place. This works, but I don't like the style. It's much easier to understand if you write it like this:

```rust
fn generate_cert_intermediate<Signature, Signer>(arg1, arg2) -> Output
where
    Signature: spki::SignatureBitStringEncoding,
    Signer: signature::Signer<Signature> + etc,
    Signer::VerifyingKey: spki::EncodePublicKey,
{ ... }
```

e2e-identity/src/pki.rs
```diff
@@ -174,124 +89,295 @@ fn get_extended_keyusage(is_ca: bool) -> x509_cert::ext::pkix::ExtendedKeyUsage
...
+
+fn generate_cert_end_entity<Signature: spki::SignatureBitStringEncoding, S>(
+    args: &CertificateGenerationArgs,
+    issuer: Name,
+    issuer_spki: spki::SubjectPublicKeyInfoOwned,
+    keypair: S,
+) -> E2eIdentityResult<x509_cert::Certificate>
+where
+    S: signature::Signer<Signature> + spki::SignatureAlgorithmIdentifier + signature::KeypairRef,
+    S::VerifyingKey: spki::EncodePublicKey,
```
coriolinus, line 346: Same generic parameter notes here.

coriolinus (approved):
Great work! A few nits, but it's awesome to get the ball rolling on this upgrade.

crypto/Cargo.toml
istankovic: Ooops, good catch! Fixed, thanks.

e2e-identity/src/pki.rs
istankovic: Aye, that looks much better. Fixed, thanks.

crypto-ffi/bindings/js/Cargo.rust-modules.toml
```diff
@@ -1,2 +1,5 @@
 [profile.release]
 opt-level = "s"
+
+[patch.crates-io]
+x509-ocsp = { git = "https://github.com/RustCrypto/formats.git", package = "x509-ocsp", version = "0.3.0-pre" }
```
SimonThormeyer, line 7: I'd appreciate a comment here that states why this patch is needed and at what point we can drop it, in addition to the hint in the commit message

istankovic: Done.

e2e-identity/src/x509_check/revocation.rs
```diff
@@ -297,6 +299,7 @@ impl PkiEnvironment {
 
             if perform_revocation_check {
                 cps.set_check_crls(true);
+                cps.set_revocation_max_age(Duration::from_hours(24));
```
typfel, line 302: What happens if a CRL refresh does not happen within 24 hours? The actual CRL refresh interval may be longer due to many reason like being offline, no internet connection, ...

istankovic: There are two cases:
1. the CRL has `nextUpdate` time set
2. the CRL has no `nextUpdate` time set

This is relevant only to the second case. If the CRL has aged beyond `max_age`, it will not be considered for revocation checking as it's going to be considered invalid.

typfel: So `nextUpdate` is something provided by the CRL issuer?  If an issuer doesn't include `nextUpdate` and we fetch max every 24 hours I think `revocation_max_age` should be bigger than our fixed refresh interval, no?

Like at least 2 * 24 hours.

istankovic: Well, I'm not quite sure if that's more preferable, for example, we don't have any extra buffering in case there is some `nextUpdate` time. In that sense, the behaviour could be seen as "similar", in that the same client could be offline for such a time that the CRL with `nextUpdate` ages beyond it.

typfel: Could this scenario happen:

1. We refresh a CRL which includes a revoked certificate
2. All groups which this client is a member of degrade
3. 24h pass without a refresh the CRL and now suddenly all the groups are no longer degraded?

That IMO would not be a desirable behaviour.

istankovic: No, that cannot happen, once a CRL (without `nextUpdate`) is considered invalid, it cannot be turned to valid anymore. Unless someone messes with the system clock, of course. (And perhaps not even then, due to caching....)

PR #2530, opened 2026-09-04
fix(keystore): allow divergent V16 migration [WPB-28480]

SimonThormeyer:
In some releases we accidentally included a whitespace-only modification
of the V16 migration file. This breaks refinery's check of executed vs.
filesystem checksums and consequently migrations fail for some clients
that had already executed the migration with the extra whitespace.

Since it's just whitespace, it's safe to ignore the divergency in this
case, assuming we're not going to accidentally change the actual SQL in
there. Luckily, due to our stepwise migration scheme, we can scope
refinery's divergency tolerance to a single migration step.


typfel (approved):

typfel:
Also note this issue only affects beta users since we haven't released this production in any client.

PR #2531, opened 2026-09-04
chore: replay keystore migrations from `main` onto `10.x`

coriolinus:
Only focuses on the migrations themselves in their end state, not replaying the full history.


coriolinus:
This doesn't work because it doesn't have all the entities defined. But we're about to do another attempt, this time including more context for the refactoring.

PR #2532, opened 2026-09-04
backport keystore changes to 10.x

coriolinus:
This branch ports the keystore refactoring that landed on `main` after the 10.4.0 release, without the transient/targeted message (TNT) feature work that landed interleaved with it. The two were entangled: several TNT PRs carried general keystore infrastructure alongside their feature code, and several of the keystore PRs touched TNT entity files. 

## Pull requests included

- #2472
- ~#2490~
- #2494
- #2496
- #2502
- #2503
- #2517
- #2527

Two commits from these PRs did not come across: 

- 022bcbb5c6 (#2496) taught `TargetedMessageTxCounter` to search by `ConversationIdRef`, and that entity does not exist here. 
- da47b0194d (#2502) became empty once its only file, `crypto/src/mls/conversation/mutable/tnt/targeted/decrypt.rs`, was dropped. 

Three further commits — 13dc7ccb3f (#2503), 46d2949d4a (#2517) and 15f96528ca (#2517) — landed with their TNT-entity hunks removed.

## Migrations

The full migration set from `main` is here: V32 through V38, plus the v34 and v37 meta-migrations. The migration directory and the generated composite-schema.sql are identical to main, which is the point.

V34/V35 and V37/V38 ride along with #2502 and #2527. V32, V33 and V36 originate in the excluded TNT PRs (#2469, #2425, #2508), so they are ported as bare SQL in two dedicated commits. They leave three tables that nothing on this branch queries, which is harmless. 

## Additional commits from excluded PRs

Six commits were cherry-picked out of the TNT PRs because the included PRs depend on them and they contain no TNT code:

Commit	| From	| Reason
---  | --- | ---
145bb510e1	| #2469	| moves ConversationId into its own module; #2496 builds ConversationIdRef on it
dd37de632e, 5ac031e93c	| #2469	| inline and then remove the mls_* convenience methods on Database
42a2588039, b723993bc8	| #2425	| BorrowPrimaryKey accepts key types with lifetimes
d50391a8fd	| #2508	| get_helper delegates to the generic composite-key helper
c15caac9ae	| #2508	| derives ToSql/FromSql and friends for ConversationId

The `mls_*` removal is worth calling out. #2517 deletes `keystore/src/mls.rs` outright, so leaving those six convenience methods in place would have meant hand-porting them into the new `connection/mls.rs` and carrying a permanent divergence from `main`. Removing them instead — which is what `main` did — made the #2517 block apply with substantially less friction.

One further commit, `get_helper_composite_key` in `keystore/src/entities/helpers.rs`, was extracted by hand from 3b29fc10f6 (#2469). That commit adds the targeted-message tx counter entity, which we don't want, but the composite-key helper it introduces is general infrastructure that `StoredEpochEncryptionKeypair` needs for its new composite key. The extraction is its own commit and records where it came from.

## Adaptations

Two upstream commits assume APIs that do not exist on this branch, and are adapted under fix(crypto): adapt the transaction-guard port to this branch:

- RecursiveError::context comes from #2482, an error-handling refactor that is out of scope. The two call sites use RecursiveError::transaction, which is what they would have used before #2482.
- #2517's new welcome tests match on `transaction_context::Error::ConversationAlreadyExists` and read the decrypted payload with `as_application_message()`. Here those are `LeafError::ConversationAlreadyExists` and `as_text()`; the Text → ApplicationMessage rename is part of #2425.

`persist_conversation_from_mls_group` also now routes through `MlsGroupState::persist`, matching ma`in, since the `Transaction::save` API it previously used is gone.


typfel:
Let's remove https://github.com/wireapp/core-crypto/pull/2490 I think that should go into CC 11

SimonThormeyer (approved):
Approving under the assumption that we're cutting out the commit range corresponding to #2490

typfel (approved):

cc-book/src/release_notes.md
```diff
@@ -1,5 +1,27 @@
 # Release Notes
 
+## Unreleased
+
+- Operations in a CoreCrypto transaction now write to the database as they happen, within a real database transaction,
```
typfel, line 5: Should this now move into the CoreCrypto 10 section?

coriolinus: Yes, but that'll happen when we drop the release which actually includes it, not in this PR.

PR #2533, opened 2026-09-04
backport: allow divergent v16 migration on 10.x

SimonThormeyer:
see above

### DIFF
--- a/keystore/src/connection/migrations/V16__prepare_credential_merge.sql
+++ b/keystore/src/connection/migrations/V16__prepare_credential_merge.sql
@@ -6,3 +6,4 @@ CREATE TABLE mls_credentials_new (
     public_key BLOB NOT NULL,
     private_key BLOB NOT NULL
 );
+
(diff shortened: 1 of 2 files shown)


keystore/src/connection/migrations/mod.rs
```diff
@@ -37,9 +37,19 @@ pub(super) fn run_migrations(conn: &mut rusqlite::Connection, target: MigrationT
...
+            runner = runner.set_abort_divergent(false);
+        }
+
         runner = runner.set_target(Target::Version(version));
         let report = runner.run(conn).map_err(Box::new)?;
 
+        if version == 16 {
+            runner = runner.set_abort_divergent(true);
+        }
+
```
coriolinus, line 52: It occurred to me last night in bed that we can just do a single unconditional

```rust
runner = runner.set_target(version != 16);
```

near the top of this loop.

... but it's probably better to retain the cherry-pick behavior instead of changing it to the prettier version.

coriolinus (approved):

keystore/src/connection/migrations/V16__prepare_credential_merge.sql
```diff
@@ -6,3 +6,4 @@ CREATE TABLE mls_credentials_new (
     public_key BLOB NOT NULL,
     private_key BLOB NOT NULL
 );
+
```
typfel, line 9: Is the intention to keep or drop the unnecessary blank line? The commit message reads like we don't want to keep it.

PR #2540, opened 2026-09-07
refactor: store both pending and reified mls groups in one table [WPB-28049]

coriolinus:
This lets us use the groups table as a foreign key in tables where that might make sense: whether it's pending or reified, a group is a group.


keystore/src/entities/mls/persisted_mls_group.rs
```diff
@@ -1,24 +1,153 @@
...
-    serde::Deserialize,
-)]
+#[derive(core_crypto_macros::Debug, Clone, PartialEq, Eq, ZeroizeOnDrop, serde::Serialize, serde::Deserialize)]
 #[zeroize(drop)]
-#[entity(table_name = "mls_groups")]
-#[sensitive]
 pub struct PersistedMlsGroup {
-    pub id: Vec<u8>,
+    #[sensitive]
+    pub id: ConversationId,
```
SimonThormeyer, line 14: Why is the id `#[sensitive]`?

coriolinus: Ugh, it shouldn't be. I'll fix it.

coriolinus: Fixed in c7a7099279db521ddc4ae7239cd8936e42811334.

keystore/src/entities/mls/persisted_mls_group.rs
```diff
@@ -33,6 +33,20 @@ pub struct PersistedMlsGroup {
...
+    /// This accomplishes that while avoiding TOCTOU issues.
+    pub fn delete_if_pending<'a, Tx>(tx: &'a Tx, id: &ConversationIdRef) -> CryptoKeystoreResult<bool>
+    where
+        &'a Tx: Into<Transactionlike<'a>>,
+    {
+        let conn = tx.into().conn()?;
+        let changed = conn.execute("DELETE FROM mls_groups WHERE id = ? AND is_pending = 1", [id])?;
+        Ok(changed > 0)
+    }
+
```
SimonThormeyer, line 48: Why did you decide against a search key type, e.g.,

```rust
enum GroupStage {
    Proper,
    Pending,
}
```

That we impl `SearchableEntity` for?

crypto/src/mls/conversation/immutable/persistence.rs
```diff
@@ -32,19 +32,27 @@ impl Conversation {
...
         let group = session
             .database()
-            .get_borrowed::<PersistedMlsGroup>(id.as_ref())
+            .get_borrowed::<PersistedMlsGroup>(ConversationIdRef::new(id.as_ref()))
             .await
             .map_err(KeystoreError::wrap("finding a persisted mls group"))?;
-        let Some(group) = group else { return Ok(None) };
+        let Some(group) = group.filter(|group| !group.is_pending) else {
+            return Ok(None);
+        };
```
SimonThormeyer, line 47: Why not filter on db level with a search key?

crypto/src/mls/conversation_cache.rs
```diff
@@ -81,9 +81,9 @@ impl ConversationCache {
             return Ok(true);
         }
         database
-            .get_borrowed::<PersistedMlsGroup>(id.as_ref())
+            .get_borrowed::<PersistedMlsGroup>(id.keystore())
             .await
-            .map(|maybe_entry| maybe_entry.is_some())
```
SimonThormeyer, line 86: Same here – why not filter on DB level?

keystore/src/entities/mls/persisted_mls_group.rs
coriolinus: We didn't have a use case for deleting only if not pending, so this was simpler.

crypto/src/mls/conversation/immutable/persistence.rs
coriolinus: I could have. But what I wanted to express in the commit message was that doing this the right way is hard and arguably out of scope of this PR.

I got excited at the beginning: "We're using the Entity API everywhere now. So I'll just add some methods!" I wrote `get_not_pending`, `get_pending`, `load_all_not_pending`, `load_all_pending`, and mapped all that such that the trait methods defer to the `not_pending` versions. 

The problem was that al those methods take `conn: &Connection`. We don't hand those out anywhere because once you've released a `&Connection` into the wild, there's no guarantee of immutability. It would cut the knees out from under our whole "all mutations happen in transactions" structure. 

The best approach I've thought of so far runs like this:

1. Invent `pub struct ImmutableConnection<'a>(&'a Connection);`
2. Adjust all Entity API read methods to take `ImmutableConnection`; there's a keystore-private method which turns an immutable connection into a real connection.
3. Have everything which currently implements `FetchFromDatabase` implement a new trait also: `trait HasConnection { fn conn(&self) -> ImmutableConnection<'_>; }`
4. Delete `FetchFromDatabase` and all implementations
5. Adjust everything we just broke.

And this would work as far as using specialized methods on an entity to filter at the DB level without inventing a search key. But it does nothing to stop the keystore itself from messing around mutably once it has that purportedly-immutable connection. (It just can't use any of the "official" helpers to do so, because those all require a transaction of one kind or another, not just a connection.)

Using a search key in this particular case isn't really worth it; it would only save us the deserialization cost of one entity, at the cost of some complexity. It's much easier to justify where the surrounding code is doing a `load_all`. But even then, I want to figure out how to do it right, and in the meantime, something obviously wrong is an easy bandaid to rip off when the time comes.

crypto/src/mls/conversation_cache.rs
coriolinus: Same answer as [here](https://github.com/wireapp/core-crypto/pull/2540#discussion_r3949459415).

keystore/src/connection/migrations/V40__mls_group_not_null_and_credential_fk.sql
```diff
@@ -0,0 +1,129 @@
...
+  -- but we can store several small fields for our own information and sort/search/filter capabilities
+  epoch INTEGER NOT NULL,
+  ciphersuite INTEGER NOT NULL,
+  credential_id BLOB NULL,
+  credential_type INTEGER NULL,
+  own_leaf_index INTEGER NOT NULL,
+  -- this field distinguishes between proper groups and pending groups
+  is_pending BOOLEAN NOT NULL DEFAULT 0,
+  FOREIGN KEY (credential_id, credential_type) REFERENCES mls_credentials (public_key_sha256, credential_type)
+    ON DELETE SET NULL
```
SimonThormeyer: Wondering here, instead of `ON DELETE SET NULL`, what if we just rejected deleting a credential that is still referenced by a group on DB level? E.g, `ON DELETE RESTRICT` (https://sqlite.org/foreignkeys.html#fk_actions)?

coriolinus: I did consider `ON DELETE RESTRICT`. I choose `SET NULL` instead because I decided that as there's a real use case for a group without a credential, it's not a bug to get into that state. If I'd gone with `RESTRICT`, then really it's just introducing an ordering dependency between groups and credentials. 

I don't feel strongly about this; it would not be hard to convince me that `RESTRICT` is the correct setting here.

crypto/src/transaction_context/conversation/mod.rs
SimonThormeyer, line 90: I know this wasn't touched in this PR, but do we even need to care about `StoredBufferedCommit`s, since they have a foreign key to conversations that has `ON DELETE CASCADE`?

coriolinus: You're right, this line should be a noop at the DB level at this point.

crypto/src/mls/conversation/group_metadata.rs
```diff
@@ -64,15 +64,21 @@ fn extract_own_updated_node_from_proposals<'a>(
...
+/// This derives the credential from our leaf in the ratchet tree, which only identifies us while we
+/// are a member. Once we have been evicted our leaf is gone, and if our former slot was recycled by
+/// a member added in the same commit, it identifies *them* — so this would report their credential,
+/// not ours. Callers must establish [`MlsGroup::is_active`] first; an evicted conversation has to
+/// reuse the credential already recorded for it instead of deriving a new one.
 pub(super) async fn current_credential(
     group: &MlsGroup,
     database: &impl FetchFromDatabase,
-) -> Result<Option<StoredCredentialPk>> {
+) -> Result<StoredCredentialPk> {
```
SimonThormeyer: We're using this function to find out about the current credential type and credential public key, i.e., the credential DB primary key, right? The function name doesn't really support this fact, maybe there could be a better name?

coriolinus: Sure, makes sense. 9f0ffdaea8

crypto/src/mls/conversation/group_metadata.rs
```diff
@@ -106,3 +106,199 @@ pub(super) async fn current_credential(
...
+
+        let intermediate_ca = alice.x509_chain_unchecked().find_local_intermediate_ca();
+        let certificate = CertificateBundle::new_with_default_values(intermediate_ca, None);
+        let new_credential = Credential::x509(case.cipher_suite(), certificate).unwrap();
+        alice
+            .transaction
+            .add_credential_without_clientid_check(new_credential.clone())
+            .await
+            .unwrap();
+        let new_credential = CredentialRef::from_credential(&new_credential);
```
SimonThormeyer, line 294: What's the reason why we're not using `transaction.add_credential()`, which returns a `CredentialRef`?

coriolinus: If I use a version which includes a client id check, it returns a `WrongCredential` error.

keystore/tests/z_entities.rs
```diff
@@ -195,9 +217,15 @@ mod tests_impl {
...
             group.id = id;
+            attach_valid_credential(&*tx, &mut group);
             group.save(&*tx).unwrap();
         }
 
+        if let Some(group) = any_e.downcast_mut::<PersistedMlsGroup>() {
+            // normally we'd avoid mutating the entity under test, but reversing sha256 is out of scope for this test
+            attach_valid_credential(&*tx, group);
+        }
+
```
SimonThormeyer, line 228: Why not embed this into the `match_heterogeneous!()` call?
Might be worth adding a match arm that matches on `mut $val`.

keystore/tests/z_entities.rs
```diff
@@ -271,11 +299,17 @@ mod tests_impl {
...
     {
-        let entity = E::random();
+        let mut entity = E::random();
+
+        let tx = store.new_transaction().await.unwrap();
+
+        if let Some(group) = (&mut entity as &mut dyn Any).downcast_mut::<PersistedMlsGroup>() {
+            attach_valid_credential(&*tx, group);
+        }
+
```
SimonThormeyer, line 309: Same here

keystore/tests/z_entities.rs
```diff
@@ -454,9 +489,14 @@ mod tests_impl {
...
                 let mut group = PersistedMlsGroup::random();
                 group.id = id;
+                attach_valid_credential(&*tx, &mut group);
                 group.save(&*tx).unwrap();
             }
 
+            if let Some(group) = any_e.downcast_mut::<PersistedMlsGroup>() {
+                attach_valid_credential(&*tx, group);
+            }
+
```
SimonThormeyer, line 499: same here

SimonThormeyer (approved):
One last nit. Great work! I like that we're leaning further into SQL features by encoding more and more in our schema.

keystore/tests/z_entities.rs
coriolinus: It needs to mutate the entity (i.e. `downcast_mut`), which `match_heterogeneous` doesn't do.

keystore/tests/z_entities.rs
coriolinus: Same mutation issue.

keystore/tests/z_entities.rs
coriolinus: Same mutation issue.

keystore/tests/z_entities.rs
SimonThormeyer: Anyway, in the past we've circumvented the need for this by doing things in a different order: in this case, this would mean creating the credential first, and only then the group. Is that possible in this case?

coriolinus: Only if we special-cased the group's creation; we couldn't use `E::random()` the way we do for everything else.

SimonThormeyer (approved):

PR #2549, opened 2026-09-07
fix(keystore): adjust refinery schema history if necessary [WPB-28480]

SimonThormeyer:
Second attempt to fix that item. Instead of using
`set_abort_divergent(false)` on the migrations runner, we just pretend that we ran the migration with the correct checksum.

### DIFF
--- a/keystore/src/connection/migrations/mod.rs
+++ b/keystore/src/connection/migrations/mod.rs
@@ -196,6 +196,40 @@ pub(crate) mod test {
         (NamedTempFile::new().unwrap(), DatabaseKey::generate())
     }
 
+    #[test]
+    fn repairs_divergent_v16_checksum_before_running_migrations() {
+        let (db_file, key) = temp_db();
+        let path = db_file
+            .path()
+            .to_str()
+            .expect("tmpfile path is representable in unicode");
+
+        smol::block_on(async {
+            let db = Database::open_at_schema_version(path, &key, MigrationTarget::Version(16))
+                .await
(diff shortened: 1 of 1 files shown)


keystore/src/connection/migrations/mod.rs
```diff
@@ -36,20 +36,34 @@ pub(super) fn run_migrations(conn: &mut rusqlite::Connection, target: MigrationT
...
 
+    // Adjust the schema history only if it already exists.
+    let has_refinery_schema_history = conn.query_row(
+        "SELECT EXISTS(SELECT 1 FROM sqlite_master WHERE type = 'table' AND name = 'refinery_schema_history')",
+        [],
+        |row| row.get::<_, bool>(0),
+    )?;
+    if has_refinery_schema_history {
+        conn.execute(
+            "UPDATE refinery_schema_history SET checksum = ?1 WHERE version = ?2",
```
typfel, line 58:
```suggestion
            "UPDATE refinery_schema_history SET checksum = ?1 WHERE version = ?2 AND checksum =?3",
```

Why not add the broken checksum as a conditional and this can be done in one query?

SimonThormeyer: Note that in the previous query we query for the table to exist, not the checksum. The table doesn't exist on fresh clients, that's why we're checking that first.

typfel: I see we would anyway need to check if table exists?

typfel (approved):

PR #2556, opened 2026-09-09
fix(kotlin)!: migrate timestamps to kotlin.time.Instant

Garzas:
Targets `main` for Core Crypto 11, since this changes the public Kotlin API.

Core Crypto's Kotlin bindings still use kotlinx-datetime 0.6.1 and expose kotlinx.datetime.Instant. This keeps consumers such as Kalium dependent on the 0.6.x compatibility artifact when moving to kotlinx-datetime 0.7.1.

Update kotlinx-datetime to 0.7.1 and migrate the UniFFI timestamp mappings to kotlin.time.Instant and the standard library's Java conversion functions.

**Breaking API change:** the generated JVM and Android Timestamp alias now resolves to kotlin.time.Instant. This affects certificate validity fields such as notBefore and notAfter. Consumers must rebuild and migrate any uses of kotlinx.datetime.Instant at this boundary.

Validation before rebasing from `release/10.x` onto `main` (the migration diff is unchanged; CI will validate the new base):
- All 61 JVM tests passed against the locally built native library.
- All 61 Android instrumented tests passed on an ARM64 Android 16 emulator, with no skipped tests and with the native library included.
- Built native Android libraries for arm64-v8a, armeabi-v7a, and x86_64, then built the debug AAR with all three libraries included. No build steps were skipped.
- JVM test static analysis passed after removing the converter-only test.

Local Android validation used NDK 29, Rust 1.96, and JDK 25. Runtime tests covered ARM64; ARMv7 and x86_64 were built and packaged.

### DIFF
--- a/cc-book/src/release_notes.md
+++ b/cc-book/src/release_notes.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Kotlin: Timestamps (Instant) are now using `kotlin.time` instead of `kotlinx.datetime`.
+
 - Kotlin: transactions are now cancellable. This applies to all Kotlin bindings: jvm, android and kmp.
 
 - Kotlin(iOS): The macos/ios targets now have a file lock guarding against concurrent access from separate CoreCrypto
--- a/crypto-ffi/bindings/gradle/libs.versions.toml
+++ b/crypto-ffi/bindings/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 kotlin = "2.3.10"
 coroutines = "1.7.3"
-kotlinx-datetime = "0.6.1"
+kotlinx-datetime = "0.7.1"
 jna = "5.17.0"
 assertj = "3.24.2"
 espresso = "3.6.1"
--- a/crypto-ffi/uniffi-android.toml
+++ b/crypto-ffi/uniffi-android.toml
@@ -4,9 +4,9 @@ android_cleaner = true
 
 [bindings.kotlin.custom_types.KotlinInstant]
 imports = [
-  "kotlinx.datetime.Instant",
-  "kotlinx.datetime.toKotlinInstant",
-  "kotlinx.datetime.toJavaInstant",
+  "kotlin.time.Instant",
+  "kotlin.time.toKotlinInstant",
+  "kotlin.time.toJavaInstant",
 ]
 type_name = "Instant"
 lift = "{}.toKotlinInstant()"
--- a/crypto-ffi/uniffi.toml
+++ b/crypto-ffi/uniffi.toml
@@ -4,9 +4,9 @@ cdylib_name = "core_crypto_ffi"
 
 [bindings.kotlin.custom_types.Timestamp]
 imports = [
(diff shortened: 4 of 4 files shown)


crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/TimestampTest.kt
```diff
@@ -0,0 +1,21 @@
+package com.wire.crypto
+
+import java.nio.ByteBuffer
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.time.Instant
+
+class TimestampTest {
```
typfel: Can this live in shared/src/commonTest instead?

istankovic:
@Garzas what do you mean by this?
> the native-library preparation step was skipped for the Kotlin compilation check.

crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/TimestampTest.kt
```diff
@@ -0,0 +1,21 @@
...
+    @Test
+    fun timestampPreservesNanosecondsAcrossFfiConversion() {
```
istankovic: What is the purpose of this test?

crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/TimestampTest.kt
Garzas: I think we can just drop this test — it mostly checks generated code, so moving it to shared probably isn't worth it

crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/TimestampTest.kt
Garzas: It was meant to check the new Instant type and nanosecond precision, but it mostly tests the generated converter. I'll remove it

Garzas:
> @Garzas what do you mean by this?
> 
> > the native-library preparation step was skipped for the Kotlin compilation check.

I was missing the Android native libraries locally, so I initially only checked Kotlin compilation. I've now built all three ABIs and the debug AAR without skipping any steps. All 61 tests also passed on an Android 16 ARM64 emulator. I've updated the PR description.

typfel:
@Garzas looks good but since this is a breaking change let's re-target the PR to towards `main`. We are about to release CC 11. It will only have small breaking changes so it should easy to to migrate to https://wireapp.github.io/core-crypto/main/cc-book/release_notes.html

crypto-ffi/bindings/jvm/src/test/kotlin/com/wire/crypto/TimestampTest.kt
istankovic: Yeah, I think we don't really need it, could you please just fold the removal commit into the first one that adds the test?

typfel (approved):

PR #2559, opened 2026-09-09
ci: improve debuggability

coriolinus:
Several misc improvements to CI intended to make slow failures into fast failures.
Also, turns out there's a linter for GHA, so fix what it reports, then include it.


SimonThormeyer (approved):
This is a nice improvement, thanks!

PR #2563, opened 2026-09-09
refactor(crypto/crypto-ffi): identities don't have a natural order

SimonThormeyer:
Note: this is not a breaking change, because uniffi doesn't by default export derived traits.


istankovic (approved):
Thanks!

PR #2569, opened 2026-09-10
ci: improve `bindings-swift` workflow

SimonThormeyer:
Three seperate issues, see commit sequence.

### DIFF
--- a/make/ffi.mk
+++ b/make/ffi.mk
@@ -6,7 +6,10 @@
 #
 # We need to build this binary if it does not exist, or if the uniffi version
 # has changed (see https://github.com/mozilla/uniffi-rs/issues/2622).
-GET_UNIFFI_VERSION = perl -ne 'print "$1\n" if /^uniffi\s=\s"([^"]+)"/' Cargo.toml
+#
+# Avoid `cargo metadata` to get uniffi version. If we used that, the `cargo metadata` call would trigger a toolchain
+# installation in any workflow/step that depends on the uniffi lib.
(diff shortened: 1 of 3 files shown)


make/ffi.mk
```diff
@@ -6,7 +6,7 @@
 #
 # We need to build this binary if it does not exist, or if the uniffi version
 # has changed (see https://github.com/mozilla/uniffi-rs/issues/2622).
-GET_UNIFFI_VERSION = perl -ne 'print "$1\n" if /^uniffi\s=\s"([^"]+)"/' Cargo.toml
+GET_UNIFFI_VERSION = perl -ne 'print "$$1\n" if /^uniffi\s=\s"([^"]+)"/' Cargo.toml
```
coriolinus, line 12: I get that perl gives us the shorter expression, but I'd honestly prefer something based on `cargo metadata` and `jq` to extract the uniffi version. It's both more readable and less fragile: this version could be broken by someone adding an additional space character to align the equals or something.

coriolinus (approved):
Good catch on all these fixes, though I think we could take it farther.

make/ffi.mk
SimonThormeyer: You're right that this would be nicer. However, we want to be able to call this without a rust toolchain. IIRC this is why we moved from `jq`/`cargo metadata` to `perl`.

SimonThormeyer: See 3473812366ddde4fde3c0a68a4317f07ca8e7a41.

I added a comment in the makefile.

PR #2574, opened 2026-09-11
test(keystore): refinery's checksums remain constant over time [WPB-2…

SimonThormeyer:
…8637]

### DIFF
--- a/keystore/src/connection/migrations/mod.rs
+++ b/keystore/src/connection/migrations/mod.rs
@@ -224,6 +224,84 @@ pub(crate) mod test {
     }
 
     #[test]
+    fn released_migration_checksums_are_unchanged() {
+        // Refinery hashes the migration name, version, and SQL, including whitespace and comments.
+        // Don't update existing, already released entries to accommodate an edit: add a new migration instead. Append a
+        // checksum whenever a migration is added. Running this test after addiing a new migration file will
+        // output the expected tuple.
+        const RELEASED_CHECKSUMS: &[(i32, u64)] = &[
+            (1, 13981446244764045850),
+            (2, 10014296231349852013),
+            (3, 10809117100344207944),
+            (4, 14920516941281032742),
+            (5, 12966312308626130346),
+            (6, 8044509004432369903),
+            (7, 17729748238669772100),
+            (8, 3048540027269991615),
+            (9, 7039584946400084759),
+            (10, 16156164571533126355),
+            (11, 13404488064808444530),
+            (12, 6597288508644970480),
+            (13, 5731883127003238395),
+            (14, 10748374302670893447),
+            (15, 2169875193630311763),
+            (16, 2746928252018864774),
+            (17, 12233330215929448823),
+            (18, 7775189113825083442),
+            (19, 7443032372885862401),
+            (20, 9998984195998664656),
+            (21, 7013639425315607404),
+            (22, 3158904360480291309),
+            (23, 17323643733849745028),
+            (24, 17262053458201743073),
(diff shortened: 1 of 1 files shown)


keystore/src/connection/migrations/mod.rs
```diff
@@ -223,6 +223,84 @@ pub(crate) mod test {
...
+                RELEASED_CHECKSUMS
+                    .iter()
+                    .any(|&(version, _)| version == migration.version()),
+                "migration V{}__{} has no pinned checksum; add this checksum tuple: ({}, {}),",
+                migration.version(),
+                migration.name(),
+                migration.version(),
+                migration.checksum(),
+            );
+        }
```
coriolinus, line 287: Could we use [`zip_eq`](https://docs.rs/itertools/latest/itertools/fn.zip_eq.html) to avoid a search operation? I'm imagining an implementation like

```rust
for ((expect_version, expect_checksum), migration) in RELEASED_CHECKSUMS.into_iter().zip_eq(runner.get_migrations()) {
  assert_eq!(expect_version, migration.version());
  assert_eq!(expect_checksum, migration.checksum());
  // plus some prettiness for the error messages as desired
}
```

Which seems both simpler and faster.

coriolinus (approved):
It's good as-is, but could be even better, I think.

keystore/src/connection/migrations/mod.rs
SimonThormeyer: Isn't it useful to retain two seperate checks, where the first is matching version number sets (so we can output the missing checksum after a new migration was added), and the second checks the checksum?

SimonThormeyer: If so, that would lead to this diff, which I'm not sure is simpler?

```
diff --git a/keystore/src/connection/migrations/mod.rs b/keystore/src/connection/migrations/mod.rs
index ac8375dd86..5e7fef0cef 100644
--- a/keystore/src/connection/migrations/mod.rs
+++ b/keystore/src/connection/migrations/mod.rs
@@ -166,6 +166,7 @@ pub async fn migrate_db_key_type_to_bytes(
 pub(crate) mod test {
     use std::io::Write;
 
+    use itertools::Itertools as _;
     use openmls::prelude::{Ciphersuite, Credential as MlsCredential, TlsSerializeTrait as _};
     use tempfile::NamedTempFile;
     use x509_cert::der::{DecodePem as _, Encode as _};
@@ -273,7 +274,13 @@ pub(crate) mod test {
         ];
 
         let runner = super::migrations::runner();
-        for migration in runner.get_migrations() {
+        let migrations = runner
+            .get_migrations()
+            .iter()
+            .sorted_by_key(|m| m.version())
+            .collect::<Vec<_>>();
+
+        for migration in migrations.iter() {
             assert!(
                 RELEASED_CHECKSUMS
                     .iter()
@@ -286,12 +293,14 @@ pub(crate) mod test {
             );
         }
 
-        for &(version, expected_checksum) in RELEASED_CHECKSUMS {
-            let migration = runner
-                .get_migrations()
-                .iter()
-                .find(|migration| migration.version() == version)
-                .unwrap();
+        for (&(version, expected_checksum), migration) in RELEASED_CHECKSUMS.iter().zip_eq(migrations) {
+            assert_eq!(
+                version,
+                migration.version(),
+                "expected migration V{version}, found V{}__{}",
+                migration.version(),
+                migration.name(),
+            );
             assert_eq!(
                 migration.checksum(),
                 expected_checksum,

```

typfel (approved):

PR #2576, opened 2026-09-11
fix: legacy idb import cannot succeed [WPB-28635]

coriolinus:
I asked AI to do an audit yesterday of the keystore for potential bugs, and it came back with a startling conclusion: legacy IDB import could not succeed.

This PR is structured as five problem-fix pairs, where the problem side adds one or more tests to demonstrate something with the migration going wrong, and the fix fixes it.

Don't panic about the diff size: of the 13k additions, 10883 are in two opaque test fixtures with IDB blobs corresponding to two different CC versions. The "real" diffstat is +2233/-169, which is still big but in the realm of human review.


keystore/src/connection/idb_migration/fixtures/generate-legacy-idb-v9.3.4.rs
```diff
@@ -0,0 +1,271 @@
...
+    latest_commit_bundle: Mutex<Option<MlsCommitBundle>>,
+}
+
+#[async_trait::async_trait(?Send)]
+impl MlsTransport for Transport {
+    async fn send_commit_bundle(&self, commit_bundle: MlsCommitBundle) -> core_crypto::Result<MlsTransportResponse> {
+        if self.retry_commits.load(Ordering::SeqCst) {
+            return Ok(MlsTransportResponse::Retry);
+        }
+        *self.latest_commit_bundle.lock().unwrap() = Some(commit_bundle);
```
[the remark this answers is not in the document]
coriolinus: /other This code will probably never be run again; it exists to document how the fixture file which lives adjacent to it in the repo was produced. It was one-time use code.

typfel (commented):
What's the value in the 10.1.0 fixture, no web client as ever been released with that version.

Aren't we covered by 9.3.4 fixture?

coriolinus:
> What's the value in the 10.1.0 fixture, no web client as ever been released with that version.
> 
> Aren't we covered by 9.3.4 fixture?

No web client requires the 10.1.0 fixture, but IMO it's worthwhile to prove that we've fixed all weird migration situations, not just all released weird migration situations.

keystore/src/connection/idb_migration/fixtures/generate-legacy-idb-v9.3.4.rs
```diff
@@ -0,0 +1,271 @@
...
+const ESTABLISHED_CONVERSATION_ID: &[u8] = b"legacy-fixture-established-conversation";
+const PENDING_CONVERSATION_ID: &[u8] = b"legacy-fixture-pending-conversation";
+const BUFFERED_COMMIT: &[u8] = b"a commit buffered for the established conversation";
+const PROTEUS_PREKEY_ID: u16 = 7;
+const PROTEUS_SESSION_ID: &str = "legacy-fixture-proteus-session-with-bob";
+const CONSUMER_DATA: &[u8] = b"consumer data kept across the import";
+const CRL_DISTRIBUTION_POINT: &str = "https://example.com/legacy-fixture.crl";
+
+const ROOT_CA_PEM: &str = "-----BEGIN CERTIFICATE-----\nMIIBjzCCAUGgAwIBAgIUTTSOvZWaT1V5dD+BpsKO66MNQe4wBQYDK2VwMDUxHzAd\nBgNVBAMMFkxlZ2FjeSBGaXh0dXJlIFJvb3QgQ0ExEjAQBgNVBAoMCVdpcmUgVGVz\ndDAeFw0yNjA5MTEwODAzMjdaFw0zNjA5MDgwODAzMjdaMDUxHzAdBgNVBAMMFkxl\nZ2FjeSBGaXh0dXJlIFJvb3QgQ0ExEjAQBgNVBAoMCVdpcmUgVGVzdDAqMAUGAytl\ncAMhADxKbOwXVgOmzS0d+iLZnSGsww7r3mc21jNE2S10XZSmo2MwYTAfBgNVHSME\nGDAWgBQxcZm+PcYZBeGnbcBf+4jS1NjeAzAPBgNVHRMBAf8EBTADAQH/MA4GA1Ud\nDwEB/wQEAwIBBjAdBgNVHQ4EFgQUMXGZvj3GGQXhp23AX/uI0tTY3gMwBQYDK2Vw\nA0EAWzblUPUv6Urg4yjVsnZ7ri0ZnbnbXs+/J8OaddwHhlPfZa+GrJegA+ReRd2O\nRfvc9tpgpXpoRDt37oBkHgJCBw==\n-----END CERTIFICATE-----\n";
```
istankovic, line 35: Why are `\n` necessary here?

istankovic:
Could you please move Markdown formatting fixes from the last commit to appropriate commits that originally broke the formatting?

keystore/src/connection/idb_migration/fixtures/generate-legacy-idb-v10.1.0.rs
```diff
@@ -0,0 +1,169 @@
+// How `legacy-idb-v10.1.0.json` was produced. Not compiled: this is a wasm-bindgen integration test written against
+// the v10.1.0 keystore API, and only builds in a checkout of that tag. See README.md alongside for the procedure.
+
+use core_crypto_keystore::{Database, DatabaseKey, connection::ConnectionType, entities::*};
+use wasm_bindgen::JsValue;
+use wasm_bindgen_futures::JsFuture;
+use wasm_bindgen_test::*;
```
istankovic, line 7: The import order be such that std comes first, 3rd-part crates second, and our crates last, with an empty line between groups.

keystore/src/connection/idb_migration/fixtures/generate-legacy-idb-v10.1.0.rs
```diff
@@ -0,0 +1,169 @@
...
+    .await
+    .unwrap();
+
+    db.commit_transaction().await.unwrap();
+    db.close().await.unwrap();
+
+    let dump: js_sys::Function = js_sys::eval(DUMP_JS).unwrap().into();
+    let promise: js_sys::Promise = dump.call1(&JsValue::NULL, &JsValue::from_str(name)).unwrap().into();
+    let json = JsFuture::from(promise).await.unwrap().as_string().unwrap();
+    panic!("FIXTURE_BEGIN\n{{\"generated_by\": \"v10.1.0\", \"database_key\": \"{DATABASE_KEY_HEX}\",{}\nFIXTURE_END", &json[1..]);
```
istankovic, line 168: I would appreciate a comment on why we panic here.

coriolinus: We're abusing the test machinery in this normally-not-compiled test to generate data to copy-paste into the JSON fixture file. Panicing makes that data visible.

keystore/src/connection/idb_migration/fixtures/generate-legacy-idb-v9.3.4.rs
coriolinus: PEM format includes newlines.

keystore/src/connection/idb_migration/fixtures/generate-legacy-idb-v10.1.0.rs
coriolinus: Yes, but technically in these files CC and the keystore are 3rd party crates. Adjusted it manually for you.

keystore/src/connection/idb_migration/fixtures/generate-legacy-idb-v9.3.4.rs
istankovic: Eh, sorry, I'm terrible at comms today... could you please use `r#""` string literals instead of plain `""` literals so that we do not need to insert newlines like that? It would be nice to format such things in a way that's more readable and easy to copy and paste into other tools for processing.

Please see `e2e-identity/src/acquisition/identity.rs` for an example.

istankovic:
I looked at `generate-legacy-idb-v10.1.0.rs` and `generate-legacy-idb-v9.3.4.rs` and while they share some parts of the code, they also differ sufficiently that I'm not sure it's worth spending additional effort to factor out the common parts. What do you think?

coriolinus:
> I looked at `generate-legacy-idb-v10.1.0.rs` and `generate-legacy-idb-v9.3.4.rs` and while they share some parts of the code, they also differ sufficiently that I'm not sure it's worth spending additional effort to factor out the common parts. What do you think?

Yes, that's why they're different files. They're meant to be run against different checkouts of the keystore in any case. Attempting a common factoring would just make that more difficult.

SimonThormeyer (approved):

PR #2579, opened 2026-09-11
chore(keystore): do not clobber `user_version` in ios [WPB-28638]

coriolinus:
We use it to track the current DB migration value, and check it in a few places, so having ios setting it unconditionally to `2` is a bit of a footgun. So far the only place it changes behavior is by invalidating the check against duplicate calls of `migrate_db_key_type_to_bytes`, but it's just nicer to have it do a sensible thing.


SimonThormeyer (approved):

typfel:
This comes from sqlcipher documentation https://www.zetetic.net/sqlcipher/sqlcipher-api/#cipher_plaintext_header_size in order to force the header to be updated.

typfel (approved):

PR #2580, opened 2026-09-11
fix: failed meta migrations only retry when DB is in the correct state [WPB-28634]

coriolinus:
Adding meta-migrations was necessary, but did add a non-trivial amount of complexity to the problem of running them.

The old implementation worked well on the happy path. This implementation works better when we need to i.e. recover from an error which happened during a meta-migration.


keystore/src/error.rs
```diff
@@ -103,6 +103,8 @@ pub enum CryptoKeystoreError {
         #[source]
         error: sqlite_wasm_vfs::relaxed_idb::RelaxedIdbError,
     },
+    #[error("the database includes migrations newer than this version of CC knows about")]
```
SimonThormeyer: Nit: most if not all other display strings start with an uppercase letter

coriolinus: right: fb8b3e05b6

SimonThormeyer (approved):
Thanks!

PR #2582, opened 2026-09-11
feat(keystore): use sqlite3mc on native targets other than Android and iOS

mtrossbach:
Hello,

while working with CoreCrypto on the JVM, we noticed a few things we'd like to contribute back. **This is part 1 of 3.** Together, these PRs make the JVM bindings, including the encrypted keystore, available on more desktop platforms:

- Linux arm64

1. #2582 sqlite3mc on desktop, in SQLCipher's format **(this PR)**
2. #2583 Linux JVM libraries with glibc 2.28, plus Linux arm64
3. #2584 Windows x86_64 JVM library

> **Merge this one first.** #2583 and #2584 build on it. It doesn't depend on anything else.

### What this PR does

If merged, the keystore uses sqlite3mc instead of SQLCipher with a vendored OpenSSL on Linux, macOS and Windows. Android and iOS keep SQLCipher.

- **Same files as before.** The build makes SQLCipher's scheme, version 4, the default of sqlite3mc. Existing keystores open without a migration, earlier CoreCrypto versions open keystores written by this one, and the raw `x'<hex>'` keys work as before.
- **No keystore code changes.** The switch is a Cargo dependency per target; `encryption.rs` and the migrations are untouched.
- **Affected builds:** the JVM builds, the Kotlin/Native targets for macOS and Linux, and the native TypeScript builds (`@wireapp/core-crypto/native`). Android and iOS build as before. Their `libsqlite3-sys` now comes from the vendored copy, which behaves as published unless the new feature is on.

### Why sqlite3mc on desktop

- **No crypto library to build.** sqlite3mc compiles from a single C file. SQLCipher needs OpenSSL, and the vendored OpenSSL needs Perl and Perl modules (in `manylinux_2_28`: `perl-IPC-Cmd`, `perl-Time-Piece`), plus NASM for its assembly on Windows. `libsqlite3-sys` describes its bundled SQLCipher build as "available for the moment only on Unix". Without OpenSSL, there's also one library less to follow for security updates.
- **All desktop platforms.** sqlite3mc supports Windows (MSVC and MinGW), Linux and macOS. A Windows build links only system DLLs.
- **The library the wasm keystore already uses**, there in a different scheme.
- **A recent SQLite.** sqlite3mc 2.4.0 is based on SQLite 3.53.4, the bundled SQLCipher 4.10.0 on SQLite 3.50.4.
- **Smaller libraries.** With the `jvm-darwin` flags, `libcore_crypto_ffi.dylib` shrinks from 11.4 MB to 8.3 MB, the static library from 29.6 MB to 19.4 MB.

### The vendored crate

rusqlite's `libsqlite3-sys` can't bundle sqlite3mc, so this PR adds a copy of `libsqlite3-sys` 0.36.0 under `vendor/`, wired in through `[patch.crates-io]`.

- The first commit adds the crate exactly as published. The second adds the sqlite3mc amalgamation with its MIT `LICENSE`, and a `bundled-sqlite3mc` feature (+42/−3 lines in `build.rs`, `Cargo.toml` and `Cargo.toml.orig`).
- Besides SQLCipher's scheme, only ChaCha20-Poly1305 stays in, unused: sqlite3mc 2.4.0 doesn't build its value-level encryption without ChaCha20 or Ascon. AES uses AES-NI or the Armv8 crypto extension where available.
- `vendor/libsqlite3-sys/sqlite3mc/README.md` records the source, the checksums, the signature check and how to update.
- `.fdignore`, the pre-commit config and `.gitattributes` keep the copy out of the formatting checks and fold it in diffs. `make/sources.mk` still hashes it for the artifact keys.
- As a path dependency, the crate drops out of the advisory check of `cargo deny`, so SQLite updates have to be followed by hand.

If you'd rather have the crate in a `wireapp` fork referenced by git, or propose the feature to rusqlite first, we're happy to do that instead.

### Testing

- `cargo test -p core-crypto-keystore` passes (81 tests, macOS arm64, Rust 1.98.0). New tests open a keystore that a SQLCipher build of `main` wrote in WAL mode (`keystore/tests/fixtures`), and cover `update_key` in WAL mode and `export_copy`.
- A `main` build (SQLCipher 4.10.0) and this branch (sqlite3mc 2.4.0) opened and wrote each other's keystores, in both directions: closed normally, with data only in the WAL file, after `update_key`, and exported. We can share the harness.
- The JVM binding tests pass against a `jvm-darwin` build (64 tests). `cargo tree` confirms the target split, and the keystore still builds for Android and iOS.
- All checks of `rust.yml`, `markdown.yml`, `toml.yml` and `deny.yml` pass locally, on this PR and on the whole series. We couldn't run the workflows, the wasm-pack keystore tests, or the Swift and TypeScript binding tests.

Thanks for having a look.


istankovic:

typfel:
@mtrossbach let's have a discussion first about  requirements, solutions, maintainability before we dive into making large changes like this.

PR #2583, opened 2026-09-11
build: build the Linux JVM libraries against glibc 2.28, and add Linux arm64

mtrossbach:
Hello,

while working with CoreCrypto on the JVM, we noticed a few things we'd like to contribute back. **This is part 2 of 3.** Together, these PRs make the JVM bindings, including the encrypted keystore, available on more desktop platforms:

- Linux arm64

1. #2582 sqlite3mc on desktop, in SQLCipher's format
2. #2583 Linux JVM libraries with glibc 2.28, plus Linux arm64 **(this PR)**
3. #2584 Windows x86_64 JVM library

> **Depends on #2582** (sqlite3mc on desktop). With no OpenSSL to build, the container below needs no extra packages.
> Please merge #2582 first. Until then, this PR also shows its three commits. The changes of this PR are the last two commits.

If you'd like this PR before #2582: we built `main`, with OpenSSL, the same way for the measurements below. The container then needs two more packages, `perl-IPC-Cmd` and `perl-Time-Piece`, which we'd add.

### What this PR does

If merged, the JVM package ships Linux libraries that load on older distributions too, and gains a Linux arm64 library.

**glibc 2.28 for Linux.** The library links against the glibc it is built with. The published 10.5.1 library, built on `ubuntu-latest`, requires glibc 2.38. It doesn't load on RHEL 8/9, SLES 15 up to SP5, Debian 11/12 or Ubuntu 20.04/22.04, where it fails with `version 'GLIBC_2.38' not found`.

- In CI, the `jvm-linux` rule now runs cargo in a `manylinux_2_28` container (AlmaLinux 8), the usual baseline for portable Linux binaries. The library then needs glibc 2.28 at most, so it loads on RHEL 8, SLES 15 SP3, Debian 11, Ubuntu 20.04 and later.
- `scripts/in-manylinux.sh` mounts the repository and the runner's Rust toolchain into the container, whose image is pinned by digest. The Linux JVM jobs therefore need Docker on the runner, which GitHub's Ubuntu images have. It also passes `SOURCE_DATE_EPOCH` into the container, so builds that set it are reproducible.
- The container build has its own target directory, `target/manylinux_2_28`, because objects from host builds, for example from a restored cache, link against the host's glibc. The libraries are then copied to where the host build puts them.
- `scripts/check-linux-library.sh` then checks in the container that the library needs glibc 2.28 at most, and loads it with Python's `ctypes`, which resolves all of its symbols. If either fails, the rule fails and removes the libraries.
- Without `CI` set, the host builds as before. `JVM_LINUX_MANYLINUX=1` uses the container on a Linux host with Docker, and `JVM_LINUX_MANYLINUX=0` turns it off in CI. The script stops with a clear message on a host of another OS or architecture, or without Docker, and it also works from a git worktree.
- `make/jvm.mk` and both scripts are prerequisites of the rules, so artifacts built before this change aren't reused. It also means that later edits to `make/jvm.mk` rebuild the Linux libraries.

**Linux arm64.**

- The new `jvm-linux-arm64` rule builds `aarch64-unknown-linux-gnu` the same way, on GitHub's `ubuntu-24.04-arm` runner.
- The jvm action picks the rule on arm64 Linux runners, and so does `make jvm` on arm64 Linux hosts.
- The JVM package carries the library as `linux-aarch64`.
- The workflows that build the package download it the same way as the darwin library.

The KMP package isn't touched, apart from getting its x86_64 static library from the same container build.

### Costs

- The image is about 520 MB compressed. A job pulls it when the Linux JVM library isn't in the artifact store yet.
- The arm64 library adds a job on `ubuntu-24.04-arm` to the pipeline. These runners are free for public repositories.
- quay.io becomes a new external dependency. The image digests are updated by hand; the script says how.

We also looked at `cargo-zigbuild` and `cross`, and at running the whole job in a `container:`. The first two bring in a new tool, and a job container would need `gh`, `jq`, `fd`, make and rustup in the image. Wrapping only the cargo command kept the workflows as they are.

### Testing

- We checked the published 10.5.1 library, and built `main` with the `jvm-linux` command line on Ubuntu 22.04 and in `manylinux_2_28`. The highest glibc symbol version each requires:

  | library | requires |
  |---|---|
  | published 10.5.1, built on `ubuntu-latest` | `GLIBC_2.38` |
  | `main`, built on Ubuntu 22.04 | `GLIBC_2.34` |
  | `main`, built in `manylinux_2_28` | `GLIBC_2.28` |
- We ran `RUSTFLAGS='-D warnings' make jvm-linux-arm64 RELEASE=1 CI=1` on a Linux arm64 host with Docker. The rule built the library in the `manylinux_2_28_aarch64` container into `target/manylinux_2_28`, copied it into place, and the check reported that it loads and needs glibc 2.28 at most. A second `make` had nothing to do.
- `make -n` shows the host command line locally, with `CI=false` and with `JVM_LINUX_MANYLINUX=0`, and the container one with `CI=1` or `CI=true`, for both rules. On an arm64 Linux host, `make jvm` picks `jvm-linux-arm64`.
- We didn't run the x86_64 rule end to end. It takes the same path with the x86_64 image, and we compiled the sqlite3mc sources with that image's GCC. In CI, the same check loads the x86_64 library.
- `actionlint` passes for the workflows, and `shellcheck` passes for both scripts. We couldn't run the workflows themselves.

If you'd prefer a different approach, let us know and we'll adapt the change.

Thanks for having a look.


istankovic:

PR #2584, opened 2026-09-11
build: add a Windows x86_64 JVM library

mtrossbach:
Hello,

while working with CoreCrypto on the JVM, we noticed a few things we'd like to contribute back. **This is part 3 of 3.** Together, these PRs make the JVM bindings, including the encrypted keystore, available on more desktop platforms:

- Linux arm64

1. #2582 sqlite3mc on desktop, in SQLCipher's format
2. #2583 Linux JVM libraries with glibc 2.28, plus Linux arm64
3. #2584 Windows x86_64 JVM library **(this PR)**

> **Depends on #2583** (Linux glibc and arm64), and through it on #2582. It extends the make rules and workflows of #2583. It builds on #2582 because with sqlite3mc there's no OpenSSL to cross-compile; we haven't tried the OpenSSL build with MinGW.
> Please merge #2582 and #2583 first. Until then, this PR also shows their five commits. The change of this PR is the last commit.

### What this PR does

If merged, the JVM package gains a Windows x86_64 library, for Windows 10 and later, the minimum of Rust's Windows targets.

- **Build.** The new `jvm-windows` rule cross-compiles `core-crypto-ffi` for `x86_64-pc-windows-gnu` with MinGW. The `build-jvm-windows` job runs it on `ubuntu-latest`, after installing `gcc-mingw-w64-x86-64`, so it needs no Windows runner and no other tools. The README describes the build.
- **Check.** The DLL imports only Windows system DLLs: `KERNEL32.dll`, `advapi32.dll`, `bcrypt.dll`, `bcryptprimitives.dll`, `ntdll.dll`, `msvcrt.dll` and `api-ms-win-core-synch-l1-2-0.dll`, so nothing needs to be installed next to it. `scripts/check-dll-imports.sh` fails the build, and removes the DLL, if it ever imports a MinGW runtime library such as `libgcc_s_seh-1.dll` or `libwinpthread-1.dll`.
- **Package.** The JVM package carries the DLL as `win32-x86-64/core_crypto_ffi.dll`, where JNA looks for it. Windows libraries have no `lib` prefix, so `copyFfiLibrary` now lists file names. The DLL adds 3.8 MB to the jar, 8.7 MB uncompressed.
- **Workflows.** The workflows that build the package download the DLL like the other libraries.

### Testing

- We ran `RUSTFLAGS='-D warnings' make jvm-windows RELEASE=1`, as in CI, in an Ubuntu 24.04 container with `gcc-mingw-w64-x86-64` and Rust 1.98.0. It built `core_crypto_ffi.dll` with sqlite3mc 2.4.0 and no OpenSSL, and `scripts/check-dll-imports.sh` passed.
- The DLL exports the same 448 `uniffi_core_crypto_ffi_*` and `ffi_core_crypto_ffi_*` functions as the Linux arm64 library built from the same commit. We compared the lists from `x86_64-w64-mingw32-objdump -p` and `nm -D`.
- Gradle's `copyFfiLibrary` puts the DLL at `win32-x86-64/core_crypto_ffi.dll` in the package resources.
- `actionlint` passes for the changed workflows, and `shellcheck` for the script. We couldn't run the workflows themselves.
- We haven't loaded this DLL on Windows yet; the checks above are static.

If you'd rather build natively with MSVC on a Windows runner, we can switch. The make-based build supports only Linux and macOS hosts today, though, so it would have to be ported first.

Thanks for having a look.


istankovic:

PR #2585, opened 2026-09-14
chore: clean up e2e-identity, part 2

istankovic:
This is the second part in the series of cleanups of e2e-identity. Most notably, this finally drops the inner PkiEnvironment wrapper and we now use `certval::environment::PkiEnvironment` directly.


e2e-identity/src/x509_check/revocation.rs
```diff
@@ -238,7 +238,7 @@ impl PkiEnvironment {
         }
     }
 
-    fn validate_cert_internal(
+    pub(crate) fn validate_cert(
         &self,
         end_identity_cert: &x509_cert::Certificate,
         perform_revocation_check: bool,
```
SimonThormeyer, line 244: optional: since we're touching the function anyway, we could move away from using a `bool` parameter to a two-member enum parameter whose variants correspond to the check policy. This would make more obvious what is happening in the call-sites

typfel (approved):

SimonThormeyer (approved):
I appreciate the commit sequence! Just one remark.

e2e-identity/src/x509_check/revocation.rs
istankovic: Heh, we thought similarly. :-) I had the same idea on my mind, but in the end decided it's not worth having another type, with likely verbose naming, just because of this. If you feel more strongly about it, I can work on it, though.

SimonThormeyer: No, I don't feel super strongly. Just in other places in the workspace we have done something like this, so I thought it'd be worth mentioning.

PR #2586, opened 2026-09-14
chore: clean up e2e-identity, part 3

istankovic:
This is the third part in the series of cleanups of e2e-identity. The highlight here is the removal of the `crl_info` module as the necessary infra is already in `certval` and we can just use it.


coriolinus (approved):
Changes look very reasonable!

PR #2598, opened 2026-09-15
chore: clean up e2e-identity, part 4

istankovic:
This is the fourth part in the series of cleanups of e2e-identity. We remove the module with old types that aren't used anymore and rename the `x509_check` module to `validation` to better reflect what it does.


e2e-identity/src/lib.rs
```diff
@@ -132,7 +132,7 @@ pub mod acme;
 pub mod pki;
 pub mod pki_env;
 pub mod utils;
-pub mod x509_check;
+pub mod validation;
```
coriolinus, line 135: This change should really be part of the previous commit, renaming the module itself.

coriolinus (approved):
Also sensible changes!

PR #2606, opened 2026-09-16
build(ts): seperate browser and native packages [WPB-26775]

SimonThormeyer:
See above


crypto-ffi/bindings/js/packages/browser/package.json
```diff
@@ -1,6 +1,50 @@
 {
-    "name": "@wireapp/core-crypto-browser",
+    "name": "@wireapp/core-crypto",
```
fewerner, line 2: Should the browser package not get a suffix as well? It's an additional breaking change but to me it would make sense. Right now we don't follow a naming convention.

.github/workflows/prepare-publish.yml
fewerner, line 1: The changes in the release workflow seem reasonable. I guess next release will test the correctness.

fewerner:
There is a typo in subject and message of a11e907 and subject of 974ec40.
The correct spelling is "separate".

fewerner (commented):
In general looks good to me. I just don't like the asymmetry with the package names.

crypto-ffi/bindings/js/packages/browser/package.json
SimonThormeyer: The convention is: no suffix for the default platform, suffix for all non-default platforms.

.github/workflows/prepare-publish.yml
SimonThormeyer: Yes, we should make a pre-release first. It is execpted that the initial publish of the new package will fail: first publish always has to be created manually (see npm/documentation#1926)

crypto-ffi/bindings/js/packages/browser/package.json
fewerner: Fair, if we consider browser the default I can see why we don't want to add another suffix.

fewerner (approved):

PR #2609, opened 2026-09-16
ci: generate and attach third party notices [WPB-25975]

coriolinus:
Generate `THIRD_PARTY_NOTICES.txt` and `sbom.cdx.json` with wireapp/notices-generator, attach them to a release, and upload them as workflow artifacts on a pull request so a change to the corrections is reviewable before it is merged.

The corrections in `.third-party-license-corrections.yaml` are read from `main` rather than from the tag: a tag carries whatever misdetections were current when it was cut, and for older releases the file does not exist at all. A pull request is the exception and uses its own copy.

Releases cut by the pipeline are created with `GITHUB_TOKEN`, and GitHub fires no workflow events for those, so the tag push is what covers the normal release path; `workflow_dispatch` regenerates the notices of a release that already exists. The workflow only ever edits a release, it never creates one.

The generator lives in an internal repository that this repository's `GITHUB_TOKEN` cannot read, so a GitHub App mints a short-lived token scoped to that one repository.

A failed Gradle scan is tolerated on a pull request, where the remaining output is still worth reading, and fatal on a release, where a notice silently missing every JVM dependency is worse than no notice at all.


.github/workflows/third-party-notices.yml
```diff
@@ -0,0 +1,330 @@
...
+          # from pinned inputs.
+          tool: cargo-sbom@0.10.0
+
+      - name: set up node
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: "26"
+
+      - name: install cdxgen
+        run: npm install -g "@cyclonedx/cdxgen@${CDXGEN_VERSION}"
```
SimonThormeyer: Why not use bun, as in almost all other places in our repo?

coriolinus: Wasn't really thinking. Honestly we should use the pre-compiled binary, which it turns out is recommended by [the docs](https://cdxgen.github.io/cdxgen/#/?id=standalone-executables-recommended).

SimonThormeyer (approved):
One question I'd like to clarify, but it's not a blocker for us to start trying it out

.github/workflows/third-party-notices.yml
coriolinus: https://github.com/wireapp/core-crypto/pull/2609/commits/bc549df232836a908a42e57dd4ec88144eafaaa0

PR #2613, opened 2026-09-17
test(keystore): improve released migrations checksums test

coriolinus:
Instead of doing quadratic search operations, check and report these cases specifically (and efficiently):

- too many items in RELEASED_CHECKSUMS
- too few items in RELEASED_CHECKSUMS
- wrong version number in RELEASED_CHECKSUMS
- changed migration

Also ensure that all relevant errors are reported instead of erroring on the first.


keystore/src/connection/migrations/mod.rs
```diff
@@ -346,31 +347,70 @@ pub(crate) mod test {
...
+                    "V{expect_version}: checksum {expect_checksum} (expected) != {} (found)",
+                    migration.checksum()
+                );
+                eprintln!("  instead of changing the old migration, add a new one");
+            }
+        }
+
+        if pending_failure {
+            panic!("pinned RELEASED_CHECKSUMS did not match the discovered migrations");
         }
```
SimonThormeyer: I'd prefer 

```rust
assert!(!pending_failure, "pinned RELEASED_CHECKSUMS did not match the discovered migrations");
```

SimonThormeyer (approved):
This is better! Just one subjective nit, that may be disregarded

PR #2615, opened 2026-09-17
chore: clean up e2e-identity, part 5

istankovic:
This is the fifth part in the series of cleanups of e2e-identity. This moves the `pki.rs` module with test infra to the `crypto` crate, which is only user of it. That in turn allows us to remove the `builder` feature of `e2e-identity`.


coriolinus (approved):
Nice!

PR #2616, opened 2026-09-17
ci: ensure that the license corrections file is up to date with ubrn …

coriolinus:
…in package.json


SimonThormeyer (approved):
Approving given the scheduled follow up work in WPB-28788

PR #2618, opened 2026-09-17
refactor!(ffi): deprecate `proteus_new_prekey()` [WPB-28571]

SimonThormeyer:
In favor of `proteus_new_prekey_auto()`. Turns out that using
language-specific deprecation warnings was very involved: we'd have to
re-introduce wrappers of `CoreCryptoContextFfi` for each language, just
to be able to deprecate it. A runtime warning is almost as good and will
reach our consumers eventually, so we go with that.


coriolinus (commented):
Looks good, pending green CI.

coriolinus (approved):

PR #2623, opened 2026-09-18
chore: increase expiration time to ease flakiness on CI

coriolinus:
This test was flaky with an expiration time of 5 seconds on CI. Increasing it by 60% should hopefully reduce that flakiness.


crypto/src/mls/credential/mod.rs
```diff
@@ -332,7 +332,7 @@ mod tests {
         }
         Box::pin(async move {
```
SimonThormeyer: Good opportunity to do a drive-by improvement and remove this `Box::pin`?

SimonThormeyer (approved):

PR #2625, opened 2026-09-18
refactor(keystore): acquire transaction lock before running migrations [WPB-28641]

SimonThormeyer:
See above


coriolinus (approved):
I was surprised to see the lock acquired twice, but expanding the context to see that one covers init and the other covers migrating the DB key type to bytes made it make sense. Nicely done!

PR #2629, opened 2026-09-21
chore: clean up e2e-identity, part 6

istankovic:
This is the sixth part in the series of cleanups of e2e-identity, where we focus on the `acme` submodule, removing dead code that is no longer necessary with the new flow. We also clean error types.


coriolinus (approved):
Very nice. Love to see the error enums slim down!

PR #2630, opened 2026-09-21
fix: unprocessable buffered messages do not lock out the group [WPB-28820]

coriolinus:
If we had too many out-of-order buffered messages pending a commit, it used to be the case that processing the commit would fail, putting the group into an invalid and irrecoverable state: processing the commit would fail, and would continue to fail forever.

Now, processing the buffered messages is done on a best-effort basis, and cannot on its own put the group into an invalid state.


SimonThormeyer (approved):
Good catch!

PR #2632, opened 2026-09-21
chore: clean up e2e-identity, part 7

istankovic:
This is the seventh part in the series of cleanups of e2e-identity. This is a set of mostly minor changes, with the most notable thing being removal of some unnecessary tests.


coriolinus (approved):
Everything here seems to be in order.

PR #2633, opened 2026-09-21
test(crypto): restore `mls_provider` tests

coriolinus:
It looks like this whole thing used to be some integration tests in a separate crate some four years ago, and when it became a module nobody noticed that they were no longer running.

It doesn't take a ton of work to get them up and passing again, mostly some updates to the names of things which have drifted over time. Lots and lots of whitespace changes though.

Note that I did remove one test entirely, which tests signatures. This is because in the actual implementaiton (`crypto_provider.rs:446`), we just left it `unimplemented` on the basis that real OpenMLS never calls this particular function.


SimonThormeyer (approved):

istankovic (approved):
Thanks for fixing this!

